PR #110, opened 2020-12-11
New prop labelFormatCallback in slider

marcialfps:
Related issue: https://github.com/dxc-technology/halstack-style-guide/issues/8

### DIFF
--- a/lib/src/slider/Slider.jsx
+++ b/lib/src/slider/Slider.jsx
@@ -28,6 +28,11 @@ const DxcSlider = ({
 
   const customTheme = useContext(ThemeContext);
   const colorsTheme = useMemo(() => getCustomTheme(theme, getCustomTheme(defaultTheme, customTheme)), [customTheme]);
+  let minLabel, maxLabel;
+  if (labelFormatCallback) {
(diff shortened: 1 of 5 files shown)


lib/src/slider/Slider.jsx
```diff
@@ -51,7 +52,11 @@ const DxcSlider = ({
   return (
     <ThemeProvider theme={colorsTheme.slider}>
       <SliderContainer margin={margin} size={size}>
-        {showLimitsValues && <MinLabelContainer disabled={disabled}>{minValue}</MinLabelContainer>}
+        {showLimitsValues && (
+          <MinLabelContainer disabled={disabled}>
+            {labelFormatCallback ? labelFormatCallback(minValue) : minValue}
```
jsardid, line 57: @marcialfps, can we use memoized values when the callback and the minValue/maxValue remain unchanged? (with useMemo)

jsardid (changes requested):

lib/src/slider/Slider.jsx
marcialfps: Added memoization in the last commit. Thanks @jsardid !

PR #274, opened 2021-05-07
[Minor] Advanced tokens in footer component

marcialfps:
New advanced tokens in footer component.


aweell:
## Changes

I would do the following changes:

* `socialIconSeparation` for `socialIconsGutter`

**Regarding `fontSizeBase`**

This token shouldn't be available, any `font-size` property should use the root `font-size` the browser gives. Then relative units can be changed on-demand element by element (e.g. `bottomLinksFontSize`, `copyrightFontSize`...) or even use fixed units if wanted. 

## Suggestions

**Regarding the divider of the bottom links**

In order to keep consistency between all the divider tokens and with the `bottomLink....` ones:

* `lineThickness` to `bottomLinksDividerThickness`
* `lineStyle` to `bottomLinksDividerStyle`
* `lineLinksSeparation` to `bottomLinksDividerSpacing`
* `lineColor` to `bottomLinksDividerColor`

**Regarding: `fontColor`**

Could be interesting to have separate color properties for custom content, links, and copyright as we have for other font properties?

**Regarding `customContentMinHeight`**

It's possible to play with the `padding` prop to increase the breathing space the custom content has. And to reach 15px a font size of 10px needs to be used. I'm not really confident leaving open space to edit min/max-widths/heights, but I understand it can be used to create more breathing space when there's not custom content.

aweell (approved):
:+1:

PR #276, opened 2021-05-07
[Minor] Advanced tokens in sidenav component 

GomezIvann:
Added advanced tokens for the component sidenav. They have also been implemented and tested.


aweell:
If I change the base tokens (e.g. `fontFamilyBase`, `fontStyleBase`, `fontColorBase`...) for font properties will override the specific ones? Or if I change a specific one... this value has more priority than the one used in the base?

GomezIvann:
The specific tokens (titleFontSize, subtitleFontFamily, linkFontStyle, etc) have priority over the 'Base' ones. This last tokens are for the Sidenav which doesn't have Title, Subtitle and Link compound components, that is, the default one [(i use the names as they stand here)](https://developer.dxc.com/tools/react/3/#/components/sidenav).
Maybe it is more interesting to call them defaultFontSize instead of fontSizeBase (i used the previous names).

aweell (approved):
:+1:

PR #280, opened 2021-05-11
[Minor] Advanced tokens for header component

marcialfps:
New advanced tokens in header component.


lib/src/common/variables.js
```diff
@@ -214,21 +213,49 @@ export const componentTokens = {
...
-    hoverHamburguerColor: globalTokens.mediumGreyBlack,
-    overlayColor: globalTokens.softBlack,
+    customContentFontFamily: "",
+    customContentFontStyle: globalTokens.type_normal,
+    customContentFontColor: "",
+    customContentFontSize: "",
+    customContentFontWeight: globalTokens.type_regular,
+    dropdownBackgroundColor: globalTokens.transparent,
+    dropdownHoverBackgroundColor: globalTokens.transparent,
     focusColor: globalTokens.blue,
```
aweell, line 223: Where is the `focusColor` token applied?

marcialfps: `focusColor` is applied to the hamburguer button when is clicked or focused.

aweell: Should we call it then `hamburguerFocusColor` instead? :hamburger:

marcialfps: `focusColor` renamed to `hamburguerFocusColor` in last commit.

aweell (approved):
👍🏼

PR #283, opened 2021-05-12
[Minor] Advanced tokens for heading component

marcialfps:
New advanced tokens in heading component.


aweell (changes requested):
Let's provide only one token for the `font-weight` property per level instead of having 3 variables related to the prop (e.g. `level1FontWeight`, `level2FontWeight` and so on...)

marcialfps:
> Let's provide only one token for the `font-weight` property per level instead of having 3 variables related to the prop (e.g. `level1FontWeight`, `level2FontWeight` and so on...)

Done in last commits. :v:

aweell (approved):
:metal:

PR #284, opened 2021-05-12
[Minor] Advanced tokens for Switch component

GomezIvann:
Added advanced tokens for the switch component. They had also been implemented and tested.


aweell (approved):
:+1:

PR #285, opened 2021-05-12
[Minor] Advanced tokens for Table component

GomezIvann:
- Added advanced tokens for the table component. They had also been implemented and tested.
- Added new tokens to sidenav component related to the scroll, which were available for other components.


aweell (approved):
👍🏼 http://gph.is/15ICB8K

PR #286, opened 2021-05-12
[Minor] Advanced tokens for Button component

jsuarezgonz:


aweell (commented):
In order to provide more flexibility apart from the `min-height` property, should we provide also the `padding` values? Since it's the only property available to decrease the width and height of the component.

**`min-height` set to 1px:**

Captura de pantalla 2021-05-12 a las 17 48 57

**`min-height` set to 1px and `padding` set to 16px**
Captura de pantalla 2021-05-12 a las 17 50 47

aweell (approved):
👍🏼

PR #288, opened 2021-05-13
[Minor] Advanced tokens in wizard component

Lisa159:


aweell (commented):
## Suggestions 

* Should be possible to change the enabled but not selected `stepContainerBackground`?
* At the same time we provide the possibility to change the `circleBorderColor` when enabled, maybe would be interesting to provide the token for the selected and disabled states?

Lisa159:
> ## Suggestions
> * Should be possible to change the enabled but not selected `stepContainerBackground`?
> * At the same time we provide the possibility to change the `circleBorderColor` when enabled, maybe would be interesting to provide the token for the selected and disabled states?

- Added new token stepContainerBackgroundColor
- Added new tokens for circle with selected and disabled states.

aweell (approved):
👍

PR #289, opened 2021-05-13
[Minor] Advanced tokens for input component

marcialfps:
New advanced tokens in input text component.


aweell (commented):
## Suggestions

* Should we provide tokens for suffix and prefix icon sizes?
* I think would be interesting to provide tokens to customize the input underline
* Is the error color affecting the "*" of the required fields? Should we provide tokens to change the size, weight and color?

marcialfps:
> ## Suggestions
> * Should we provide tokens for suffix and prefix icon sizes?
These tokens were added but I discarded them because user can break the component. 
Example:
image

> * I think would be interesting to provide tokens to customize the input underline
The input underline uses fontColorBase to apply a color, maybe we can create a new specific token when the user wants a different color for the underline. Apart from this token, could it be interesting to add a thickness token?

> * Is the error color affecting the "*" of the required fields? Should we provide tokens to change the size, weight and color?
No. The error color is only for the error status, not for the required status.

aweell:
We can add the following then:

* `underlineColor`
* `underlineFocusColor`
* `underlineThickness`
* `requiredMarkColor`
* `requiredMarkFontWeight`
* `requiredMarkFontSize`

marcialfps:
> We can add the following then:
> 
> * `underlineColor`
> * `underlineFocusColor`
> * `underlineThickness`

:thumbsup:

> * `requiredMarkColor`
> * `requiredMarkFontWeight`
> * `requiredMarkFontSize`

The Required mark is a common component (DxcRequired) shared with Checkbox, Radio, Select, Switch and Textarea. If we want to add these tokens we need to separate this component for each case or create a new public component that can be tokenised.

aweell (commented):
@jsardid can you give your input about this?

> The Required mark is a common component (DxcRequired) shared with Checkbox, Radio, Select, Switch and Textarea. If we want to add these tokens we need to separate this component for each case or create a new public component that can be tokenised.

PR #291, opened 2021-05-14
[Minor] Advanced tokens for Tabs component

GomezIvann:
Added advanced tokens for the Tabs component. They had also been implemented and tested. This includes the theme-builder previews.


aweell (commented):
Should we provide `min-height` and tab `padding` properties in order to make them vertically sizable?

aweell (approved):
👍🏼

PR #296, opened 2021-05-17
[Minor] Advanced tokens for Dropdown component

jsuarezgonz:


aweell (changes requested):
* Should we make a clear distinction between the `background-color` applied to the button and the one applied to the options list? (e.g. `buttonBackgroundColor`, `optionsListBackgroundColor`)
* Could be interesting to provide different font tokens for the dropdown label and the list options?
* Are the border properties applied to the whole container?

jsuarezgonz:
> * Should we make a clear distinction between the `background-color` applied to the button and the one applied to the options list? (e.g. `buttonBackgroundColor`, `optionsListBackgroundColor`)

We have two different tokens 'Background color' for the button background and 'Dropdown background color' for the list. Should we rename them?

> * Could be interesting to provide different font tokens for the dropdown label and the list options? 

Done

> * Are the border properties applied to the whole container?

Yes, borders apply to the whole container.

aweell:
> We have two different tokens 'Background color' for the button background and 'Dropdown background color' for the list. Should we rename them?

Yes, I believe they're more convenient 😉

aweell (approved):
👍🏼

PR #297, opened 2021-05-17
[Minor] Advanced tokens for Dialog component

jsuarezgonz:


aweell (commented):
Should we add more font properties for the custom content? (As we're doing for the accordion for example)

jsuarezgonz:
> Should we add more font properties for the custom content? (As we're doing for the accordion for example)

Done

aweell (approved):
👍🏼

PR #299, opened 2021-05-17
[Minor] Advanced tokens for Toggle Group component

GomezIvann:
Added advanced tokens for the Tabs component. They had also been implemented and tested. This includes the theme-builder previews.


lib/src/common/variables.js
```diff
@@ -653,24 +653,35 @@ export const componentTokens = {
...
     disabledSelectedFontColor: globalTokens.white,
-    fontFamily: globalTokens.type_sans,
-    fontSizeBase: globalTokens.type_scale_root,
-    fontSize: globalTokens.type_scale_02,
-    fontStyle: globalTokens.type_normal,
-    fontTextTransform: globalTokens.type_uppercase,
-    fontLetterSpacingWide02: globalTokens.type_spacing_wide_02,
+    disabledUnselectedBackgroundColor: globalTokens.lightWhite,
+    disabledUnselectedFontColor: globalTokens.lighterBlack,
+    disabledFontStyle: globalTokens.type_normal,
```
aweell, line 674: Why we have a `font-style` property for disabled state, but no one for other states?

aweell (commented):
Should we have a token for focus outline?

lib/src/common/variables.js
GomezIvann: I though it would be interesting to have the option to differentiate the style of the font when it is disabled. And regarding the focus outline, you mean something like a **focusColor** token as we did with other previous components?

aweell: > I thought it would be interesting to have the option to differentiate the style of the font when it is disabled.

I understand how adding the style in the disabled state can be interesting, but shouldn't be interesting too if we provide a change of `font-weight` for the selected state for example? Are we losing common criteria if we provide one property only for a specific state?

In this case, the chances of having this property edited I believed were low... but I rather leave this to your criteria (Besides, one property more does anything but harm  😉  )

> you mean something like a `focusColor` token as we did with other previous components?

Yes, I do!

GomezIvann: I completely agree with all you've said, i think we should follow the same criteria in every component, at least in this initial version, maybe for the next one :smiley:

aweell (approved):
👍🏼

PR #301, opened 2021-05-18
[Minor] Advanced tokens for resultset table component 

marcialfps:
New advanced tokens in resultset table component.


aweell (changes requested):
Should we add a `border-radius` property to change the applied to the table header?

marcialfps:
> Should we add a `border-radius` property to change the applied to the table header?

If we add `border-radius` it is added to the Table component and not to the ResultSet Table.

PR #302, opened 2021-05-18
[Minor] Advanced tokens for Textarea component

GomezIvann:
Added advanced tokens for the Textarea component. They had also been implemented and tested. This includes the theme-builder previews.


aweell (commented):
With which tokens can the font properties of the text area content be changed?

GomezIvann:
> With which tokens can the font properties of the text area content be changed?

With none of them. Should i implement specific tokens for the content or synchronise them with those of the label?

aweell (commented):
I think it would be better to create the `customContent...` tokens that we have already in other components

aweell (approved):
👍🏼

PR #303, opened 2021-05-19
[Minor] Advanced tokens for Checkbox component

jsuarezgonz:


lib/src/common/variables.js
```diff
@@ -252,15 +252,27 @@ export const componentTokens = {
...
+    disabledBorderColor:globalTokens.softBlue,
+    disabledBorderColorOnDark:globalTokens.softBlue,
+    disabledColorOnDark: "#BFBFBF",
     disabledCheckColor: globalTokens.white,
+    disabledCheckColorOnDark: "#FFFFFF",
     disabledFontColor: globalTokens.lighterBlack,
+    disabledFontColorOnDark: "#575757",
     focusColor: globalTokens.blue,
-    fontFamily: globalTokens.type_sans,
+    checkGutter: "10px"
```
aweell, line 275: I suggest changing `checkGutter` to `checkLabelSpacing` since is more semantic

aweell (commented):
Should we add tokens for the hover state? In order to provide a cue, like changing the bg or border color on hover?

aweell (changes requested):
I saw you added `backgroundHoverColorChecked`, should we have one for unchecked also?

jsuarezgonz:
> I saw you added `backgroundHoverColorChecked`, should we have one for unchecked also?

There is no token referring to the 'unchecked' component, we only have the border in that case; that´s why I created a hover token for the border and not for the background.

aweell (approved):
🤘🏼

PR #304, opened 2021-05-19
[Minor] Advanced tokens for select component

marcialfps:
New advanced tokens in select component.


aweell (changes requested):
* Should we add properties for changing the `border-radius` of the options list and the focus container?
* Should we add an option to change the `background-color` of the options list?

marcialfps:
> *Should we add properties for changing the `border-radius` of the options list and the focus container?

If we add `border-radius` token for options list other tokens like `border-color` and `border-thickness` appears. We didn't add `border-radius` for focus in any component.


> *Should we add an option to change the `background-color` of the options list?

marcialfps:
> *Should we add properties for changing the `border-radius` of the options list and the focus container?

If we add the `border-radius` token to the list of options, other tokens such as `border-color` and `border-thickness` appear. We have not added `border-radius` for focus in any of the components.

> *Should we add an option to change the `background-color` of the options list?

Done. ✌🏼

aweell:
> We have not added border-radius for focus in any of the components.

I know I opened the Pandora's box with the focus state...lets keep it as it is right now 😉 

> If we add the border-radius token to the list of options, other tokens such as border-color and border-thickness appear.

I think that if we add the border properties to the dropdown options list we should be consistent and add them to the select too.

marcialfps:
> I think that if we add the border properties to the dropdown options list we should be consistent and add them to the select too.

I think the `border-radius` is a token that makes the component look strange. `border-color`, `border-thickness` and `border-style` are fine to add.
image

aweell (approved):
👍🏼

aweell (approved):
👍🏼

PR #305, opened 2021-05-20
[Minor] Advanced tokens for Upload component

GomezIvann:
- Added advanced tokens for the Upload component. They had also been implemented and tested. This includes the theme-builder previews.
- Fixed issue #308. Now file names should display properly, with an ellipsis in case they are too long to show.


aweell (commented):
* Where the `accentColor` and the `secondaryBackgroundColor` tokens apply?

GomezIvann:
> * Where the `accentColor` and the `secondaryBackgroundColor` tokens apply?

**accentColor** is a previous token which i decided to preserve. It gives color to the border-bottom of every file in the file list:

image

Maybe a better name such us **fileUnderlineColor** would fit better. And **secondaryBackgroundColor** is applied over the two backgrounds that are not initially white (they are a kind of grey):

image

image

aweell:
* I agree with the `fileUnderlineColor` change.
* Since is applied only to the drag and drop area, we change it to `dropAreaBackgroundColor`

aweell (approved):
👍🏼

Issue #282, opened 2021-05-11
[Tag] Component specs


aweell:
**Icon size**

Source | Property | Value
-- | -- | -- |
 design guidelines  | Icon size	| 24x24px
implementation  | Icon size	| 23x23px

**Icon section background-color**

Source | Property | Value
-- | -- | -- |
design guidelines  | background-color	| #6f2c91
implementation  | background-color	| #6b4187

Icon size and background-color should be updated to match the design guidelines.

PR #306, opened 2021-05-20
[Minor] Advanced tokens for tag component

marcialfps:
New advanced tokens in tag component.


aweell (changes requested):
* Should we add a token for the `background-color` of the icon container area and another one for the label area? If this is added a new token for the `font-color` will be needed.
* Could be interesting adding tokens for the `border-radius`?
* Should we provide a way to customize the container shadow?
* How can the `font-size` of the label be changed?

marcialfps:
> * Should we add a token for the `background-color` of the icon container area and another one for the label area? If this is added a new token for the `font-color` will be needed.

Background color of the icon container is provided by a prop.
image

This component uses a Box internally, so all tokens from Box are applied in Tag. So, tokens like `font-color` are "inherited" from Box.

> * Could be interesting adding tokens for the `border-radius`?
> * Should we provide a way to customize the container shadow?
> * How can the `font-size` of the label be changed?

`border-radius`, shadows and `font-size` are tokens in Box component.


> Related issues #282

Done in last commit. :v:

aweell (approved):
Thanks for the clarifications @marcialfps. Everything seems perfect then 🤘🏼

PR #331, opened 2021-06-01
[Patch] Select - avoid label displacement in focused option

marcialfps:
Now when you focus on an option the label does not move.


aweell (approved):
🤘

PR #355, opened 2021-06-25
Fixed the displacement caused by Link's underline

GomezIvann:
Fixed problem with Link's underline which produced the surounding text to displace when it was hovered.


aweell (approved):
👍🏼

PR #440, opened 2021-11-26
[Patch] Change to use number context props.

raquelarrojo:
Remove unnecessary stateful variables for minNumber, maxNumber and stepNumber, and use the numberContext props instead.

### DIFF
--- a/lib/src/new-input-text/NewInputText.jsx
+++ b/lib/src/new-input-text/NewInputText.jsx
@@ -70,10 +70,6 @@ const DxcNewInputText = React.forwardRef(
     const [filteredSuggestions, changeFilteredSuggestions] = useState([]);
     const [visualFocusedSuggIndex, changeVisualFocusedSuggIndex] = useState(-1);
 
-    const [minNumber, setMinNumber] = useState(null);
-    const [maxNumber, setMaxNumber] = useState(null);
-    const [stepNumber, setStepNumber] = useState(null);
-
     const suggestionsRef = useRef(null);
     const inputRef = useRef(null);
     const actionRef = useRef(null);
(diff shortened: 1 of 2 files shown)


lib/src/new-input-text/NewInputText.jsx
```diff
@@ -291,24 +287,26 @@ const DxcNewInputText = React.forwardRef(
 
     const decrementNumber = () => {
       const numberValue = value || innerValue;
```
GomezIvann, line 289: Wouldn't it be better: `const numberValue = value ?? innerValue;` ? If `value` has a [_falsy value_](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), that includes 0, this variable would have the `innerValue` value instead, which i don't think is the right behaviour here, what do you think?

lib/src/new-input-text/NewInputText.jsx
```diff
@@ -318,22 +316,25 @@ const DxcNewInputText = React.forwardRef(
 
     const incrementNumber = () => {
       const numberValue = value || innerValue;
```
GomezIvann, line 318: Wouldn't it be better: `const numberValue = value ?? innerValue;` ? If `value` has a [_falsy value_](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), that includes 0, this variable would have the `innerValue` value instead, which i don't think is the right behaviour here, what do you think?

lib/src/new-input-text/NewInputText.jsx
raquelarrojo: In this case it is not totally necessary because value is a string (it is parsed to int when the changeValue function is called). So, in the case the value is "0", it will be correctly evaluated using the `||` operator. But I think it could be a good idea to change to `??` operator anyway.

lib/src/new-input-text/NewInputText.jsx
raquelarrojo: In this case, it is not totally necessary because value is a string (it is parsed to int when the changeValue function is called). So, in the case the value is "0", it will be correctly evaluated using the `||` operator. Also, `value` cannot have any other falsy value (like " ", as it is not possible to write empty space on the number component). But I think it could be a good idea to change to `??` operator anyway.

lib/src/new-input-text/NewInputText.jsx
GomezIvann: True, i was thinking as if it was a type number, I did not remember that it was a string, well thought!

PR #441, opened 2021-11-30
[Minor] First version of the New Select component

GomezIvann:
What we currently have:

- Complete layout of the component according to XD specs.
- Single selection functionality.
- Groups.
- Visual focus in the dropdown selection list.
- Search functionality. Selected option and search box duality. Clear action for cleaning the search value.
- Controlled version: onChange and value.
- Optional and required error both in onChange & onBlur events.
- Readonly, disabled and error states implemented.
- Margin, size, tabIndex and ref.
- Types in d.ts added.
- Tokens.

Remaining:
- Replace label with span and aria-labelledby.
- Multiple selection (with caret).
- Adding key events.
- Autoscroll for single selected option when dropdown menu opens.
- aria attributes.
- Use Radix popover for dropdown menu.
- Review tokens.


lib/src/new-select/NewSelect.jsx
```diff
@@ -0,0 +1,699 @@
...
+                }}
+                onMouseLeave={() => {
+                  changeVisualFocusedSuggIndex(-1);
+                }}
+              >
+                {searchable ? filteredOptions.map(mapOptionFunc) : options.map(mapOptionFunc)}
+              </OptionsList>
+            )}
+          </SelectContainer>
+          {!disabled && <Error backgroundType={backgroundType}>{error}</Error>}
```
raquelarrojo, line 300: Error span should be shown only if error exists, so we should add this check `{!disabled && error && <Error.../>}`. If not, the error span is always shown.

lib/src/new-select/NewSelect.jsx
```diff
@@ -0,0 +1,699 @@
...
+        : props.theme.disabledLabelFontColor
+      : props.backgroundType === "dark"
+      ? props.theme.labelFontColorOnDark
+      : props.theme.labelFontColor};
+
+  font-family: ${(props) => props.theme.fontFamily};
+  font-size: ${(props) => props.theme.labelFontSize};
+  font-style: ${(props) => props.theme.labelFontStyle};
+  font-weight: ${(props) => props.theme.labelFontWeight};
+  line-height: 1.715em;
```
raquelarrojo, line 349: I think we should add a token for `labelLineHeight`, as in the rest of components.

lib/src/new-select/NewSelect.jsx
```diff
@@ -0,0 +1,699 @@
...
+      : props.backgroundType === "dark"
+      ? props.theme.helperTextFontColorOnDark
+      : props.theme.helperTextFontColor};
+
+  font-family: ${(props) => props.theme.fontFamily};
+  font-size: ${(props) => props.theme.helperTextFontSize};
+  font-style: ${(props) => props.theme.helperTextFontStyle};
+  font-weight: ${(props) => props.theme.helperTextFontWeight};
+  line-height: 1.5em;
+`;
```
raquelarrojo, line 371: I think we should add a token for `helperTextLineHeight`, as in the rest of components.

lib/src/new-select/index.d.ts
```diff
@@ -0,0 +1,40 @@
...
+  placeholder: string;
+  disabled: boolean;
+  optional: boolean;
+  searchable: boolean;
+  multiple: boolean;
+  onChange: (value: string | string[]) => void;
+  onBlur: (val: { value: string | string[]; error: string }) => void;
+  error: string;
+  margin: string | Margin;
+  size: "small" | "medium" | "large" | "fillParent";
```
raquelarrojo, line 35: We can add a Size type like this:`type Size = "small" | "medium" | "large" | "fillParent";` and use if for the size property, like in the rest of components.
Comments are left but I suppose it is an initial version.

lib/src/new-select/NewSelect.jsx
```diff
@@ -0,0 +1,699 @@
...
+  padding: 0 calc(1rem * 0.5);
+  outline: none;
+  ${(props) => {
+    if (props.disabled)
+      return props.backgroundType === "dark"
+        ? `background-color: ${props.theme.disabledContainerFillColorOnDark};`
+        : `background-color: ${props.theme.disabledContainerFillColor};`;
+  }}
+  box-shadow: 0 0 0 2px transparent;
+  border-radius: 4px;
```
raquelarrojo, line 388: I think we should add a token for `selectBorderRadius`.

lib/src/new-select/NewSelect.jsx
```diff
@@ -0,0 +1,699 @@
...
+  border: 1px solid
```
raquelarrojo, line 389: We could also add tokens for selectBorderThickness and selectBorderStyle.

lib/src/new-select/NewSelect.jsx
```diff
@@ -0,0 +1,699 @@
...
+            option.options.forEach((singleOption) => {
+              if (singleOption.value === val) selectedOptionLabel = singleOption.label;
+            });
+          } else if (option.value === val) selectedOptionLabel = option.label;
+        });
+      }
+
+      return selectedOptionLabel;
+    };
+
```
raquelarrojo, line 149: I think it could be interesting to encapsulate this function in a useMemo. Now this function is being executed always (when the select is opened, when we make hover of an option, when we select an option...). Something like this `const selectedLabel = useMemo(() => getSingleSelectedOptionLabel(), [value ?? innerValue]);` (and use this `selectedLabel` in the render instead of calling the function) will make that this is only called when the value changes.

lib/src/new-select/NewSelect.jsx
```diff
@@ -0,0 +1,699 @@
...
+      );
+    };
+    let global_index = 0;
+    const mapOptionFunc = (option, index) => {
+      if (option.options) {
+        return (
+          <>
+            {option.options.length > 0 && <OptionGroupLabel>{option.label}</OptionGroupLabel>}
+            {option.options.map((singleOption) => {
+              global_index++;
```
raquelarrojo, line 228: The second param of the map function is the index of the current element. If you do `option.options.map((singleOption), i) => ...`, you can use it to pass it to the `Option` like `index={i}` and you don't need to create a counter (`global_index`).

GomezIvann: The main goal of using a global variable instead of the regular map index is that it preserves its value between `mapOptionFunc` calls. What i'm trying to achieve is that every option of each group has its own index, to be able for visual focus to identify which is been hovered when the mouse enters and paint it properly. If i use the index of map, it will always start with 0, causing this to happen: 

image

lib/src/new-select/NewSelect.jsx
GomezIvann: Great idea, I will implement it, thanks!

lib/src/new-select/index.d.ts
GomezIvann: This change respect to the other components has a purpose. If we use an in-between type, then when you press `Ctrl` key over the prop while consuming the component, you see the type Size but not its values, as this:

Sin título 2

While if you do it like above, you see them:

Sin título

I find this way more useful for the user, what do you think?

raquelarrojo (approved):

PR #469, opened 2021-12-28
[Patch] Added title to password icon and change in styles of file input.

raquelarrojo:
- Added `Show` title to the password icon if it is not visible and `Hide` if the password is visible.
- Changes in styles of the hover, focus and active of the file item's delete button.


GomezIvann (approved):

aweell:
The copy `show` or `hide` as title can be misleading or ambiguous, `show password` / `hide password` instead, describes the action the user will perform clicking the button better.

raquelarrojo:
> The copy `show` or `hide` as title can be misleading or ambiguous, `show password` / `hide password` instead, describes the action the user will perform clicking the button better.

raquelarrojo:
Title changed to `Show password` and `Hide password`.

PR #494, opened 2022-01-10
[Minor] TextInput component update and Select small fixes

GomezIvann:
Summary:

- Autosuggest list specs updated.
- TextInput tokens review.
- Tests updated to the new autosuggest implementation and included more.
- Some small fixes done in the Select along the way.


lib/src/text-input/TextInput.jsx
```diff
@@ -889,20 +880,22 @@ const Error = styled.span`
 `;
 
 const Suggestions = styled.ul`
-  background-color: ${(props) => (props.isError ? props.theme.errorMessageBackgroundColor : "#ffffff")};
   position: absolute;
   z-index: 1;
-  max-height: 160px;
-  overflow: auto;
+  max-height: 304px;
+  overflow-x: auto;
```
aweell, line 886: Why do we have an auto `overflow-x` property value? Shouldn't be `overflow-y:scroll;`?

GomezIvann: You are right Alex, changed to `overflow-y: auto`, thanks!

PR #497, opened 2022-01-10
[Minor] Change file input's token name.

raquelarrojo:
Change focusDropBackgroundColor to dragOverDropBackgroundColor in file input.


docs/src/pages/themeBuilder/themes/AdvancedTheme.json
```diff
@@ -344,13 +344,13 @@
     "disabledDropLabelFontColor": "#999999",
     "focusDropBorderColor": "#0095ff",
     "disabledDropBorderColor": "#999999",
-    "focusDropBackgroundColor": "#33AAFF",
+    "dragoverDropBackgroundColor": "#33AAFF",
```
aweell, line 347: The value #33AAFF is for `color-blue-500` token, the background color od the drop zone in `dragover` state should use `color-blue-50`

From the documentation:

| Component token                         | Element                     | Core token                  | Value         |
| :-------------------------------------- | :-------------------------- | :-------------------------- | :------------ |
| `dragoverDropBackgroundColor`           | Dnd fill:dragover           | `color-blue-50`             |   #f5fbff     |

aweell (changes requested):

docs/src/pages/themeBuilder/themes/AdvancedTheme.json
raquelarrojo: Updated value in schema too.

aweell (approved):

PR #500, opened 2022-01-11
[Minor] Select tokens review

GomezIvann:
Select

- Tokens review, new tokens were added and others were deleted.
- Fix issue regarding the selected value color when the list was opened and it was not searchable. Now it doesn't change its color.
- Selection Indicator sizes fixed.
- Stories file included.

TextInput

- New tokens added: `inputMarginTop` and `inputMarginBottom`.


lib/src/select/Select.jsx
```diff
@@ -661,14 +665,16 @@ const SelectionIndicator = styled.span`
...
+  max-height: 22px;
+  width: 46px;
 `;
 
 const SelectionNumber = styled.span`
   display: flex;
   align-items: center;
   justify-content: center;
-  width: 24px;
+  width: 22px;
```
aweell, line 676: Why the change from 24px to 22px?

aweell (approved):

PR #527, opened 2022-01-20
Types for DateInput component

GomezIvann:


lib/src/date-input/types.ts
```diff
@@ -47,21 +46,21 @@ type Props = {
...
+  /**
    * This function will be called when the user types within the input
    * element of the component. An object including the string value, the
    * error and the date value will be passed to this function.
    * If the string value is a valid date, error will
    * be null. Also, if the string value is not a valid date, date will be null.
    */
-  onChange?: (val: { value: string; error: string; date: Date }) => void;
-  /**          
+  onChange?: (val?: { value?: string; error?: string; date?: Date }) => void;
```
marcialfps, line 56: This props must be required because if it is not defined the component does not work.

GomezIvann: It's already changed but not commited, i have to discuss some things of this signature with @jsardid before uploading!

marcialfps (approved):

PR #542, opened 2022-01-24
Types for Accordion component

GomezIvann:


lib/src/accordion/Accordion.tsx
```diff
@@ -4,34 +4,31 @@ import ExpansionPanel from "@material-ui/core/ExpansionPanel";
...
-  isExpanded,
+  onChange,
   children,
   margin,
   padding,
   tabIndex = 0,
-}) => {
-  const [innerIsExpanded, setInnerIsExpanded] = React.useState(false);
-  const [isResponsive, setIsResponsive] = useState();
+}: AccordionProps) => {
```
marcialfps, line 25: We should add JSX.Element type.

GomezIvann: done, thanks!!

marcialfps (approved):

PR #546, opened 2022-01-27
Types for Tabs component

### LINKED ISSUE #544: Add types for Tabs component

marcialfps:
Closes #544

 ## Changes

- Props typing with required `tabs` prop.
- `tabs` prop has no default value.
- Removed typeof condition.
- Removed unused variables (`getTabIndex` and `tabContent`).


lib/src/tabs/Tabs.tsx
```diff
@@ -25,14 +25,9 @@ const DxcTabs = ({
     if (activeTabIndex == null) {
       setInnerActiveTabIndex(newValue);
     }
-    if (typeof onTabClick === "function") {
-      onTabClick(newValue);
-    }
+    onTabClick(newValue);
```
GomezIvann, line 28: When an event is marked as optional in _types.ts_, we should add optional method chaining when it is called: `onCloseClick?.()` or it will throw a **ReferenceError**. Be careful when removing `typeof === "function"`, this line is also controlling that is not undefined or null.

lib/src/tabs/Tabs.tsx
```diff
@@ -67,24 +62,21 @@ const DxcTabs = ({
...
-          })}
+          {tabs.map((tab, i) => (
+            <Tab
+              tabIndex={(activeTabIndex === i || innerActiveTabIndex === i) && !tab.isDisabled ? tabIndex : -1}
+              key={`tab${i}${tab.label}`}
+              label={getLabelForTab(tab)}
+              disabled={tab.isDisabled}
+              disableRipple={true}
+              onMouseEnter={() => {
+                onTabHover(i);
```
GomezIvann, line 73: When an event is marked as optional in _types.ts_, we should add optional method chaining when it is called: `onCloseClick?.()` or it will throw a **ReferenceError**. Be careful when removing `typeof === "function"`, this line is also controlling that is not undefined or null.

lib/src/tabs/Tabs.tsx
```diff
@@ -67,24 +62,21 @@ const DxcTabs = ({
...
+              }}
+              onMouseLeave={() => {
+                onTabHover(null);
```
GomezIvann, line 76: When an event is marked as optional in _types.ts_, we should add optional method chaining when it is called: `onCloseClick?.()` or it will throw a **ReferenceError**. Be careful when removing `typeof === "function"`, this line is also controlling that is not undefined or null.

GomezIvann (changes requested):

lib/src/tabs/Tabs.tsx
marcialfps: All changes done in last commit ✌🏼

GomezIvann (dismissed):

GomezIvann (approved):

PR #558, opened 2022-01-28
Accordion Group refactor for typing

GomezIvann:


lib/src/accordion-group/types.ts
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Size of the padding to be applied to the custom area ('xxsmall' | 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge').
+   * You can pass an object with 'top', 'bottom', 'left' and 'right' properties in order to specify different padding sizes.
+   */
+  padding?: Space | Padding;
+  /**
+   * The expanded panel of the accordion. This area can be used to render
+   * custom content.
+   */
+  children: React.ReactNode;
```
marcialfps, line 46: I think an accordion can be empty.

lib/src/accordion/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * This function will be called when the user clicks the accordion to expand or collapse
+   * the panel. The new state of the panel will be passed as a parameter.
+   */
+  onChange?: (isExpanded: boolean) => void;
+  /**
+   * The expanded panel of the accordion. This area can be used to render
+   * custom content.
+   */
+  children: React.ReactNode;
```
marcialfps, line 52: Same. I think this prop should be optional.

jsardid: Prop shouldn't be optional in order to enforce the design guidelines.

PR #560, opened 2022-01-28
Added textarea stories

jsuarezgonz:


raquelarrojo:
- We should add tests for margins.
- Add optional to the disabled tests (both in light and dark).
- Add tests disabled with value (both in light and dark).
- Maybe it is interesting to add an example with verticalGrow as 'auto' and value to see how it grows.
- We could test helper text, optional and value in a same test to avoid repeated tests (as these elements do not affect each other).

jsuarezgonz:
> * We should add tests for margins. **Done**
> * We could add optional to the disabled tests (in light and dark). **Done**
> * We could add tests disabled with value (in light and dark). **Done**
> * Maybe it is interesting to add an example with verticalGrow as 'auto' and value to see how it grows. **(Instead of adding the test for the verticalGrow as 'auto'(that doesn´t add anything else than the 'resizer icon' already tested in another scenario) I´ve added a test for the verticalGrow as 'manual', test case in which the scrollbar is displayed.)**
> * We could test helper text, optional and value in a same test to avoid repeated tests (as these elements do not affect each other). **Done**

raquelarrojo (approved):

PR #579, opened 2022-01-31
Types for Slider component

GomezIvann:
Summary of changes in Slider component:

- Types included.
- Updated docs.
- Added tabIndex prop.


lib/src/slider/Slider.tsx
```diff
@@ -40,32 +41,22 @@ const DxcSlider = ({
...
-    if (typeof onChange === "function") {
-      onChange(newValue);
-    }
+    const valueToCheck = value ?? innerValue;
+    valueToCheck !== newValue && setInnerValue(newValue);
+    onChange?.(newValue);
+  };
+
+  const handleSliderOnChangeCommited = (event, selectedValue) => {
+    onDragEnd?.(selectedValue);
```
marcialfps, line 50: What are these changes?

lib/src/slider/Slider.tsx
```diff
@@ -81,14 +72,15 @@ const DxcSlider = ({
             </MinLabelContainer>
           )}
           <Slider
-            value={(value != null && value >= 0 && value) || innerValue}
+            value={(value != null && value >= 0 && value) || innerValue} // === value ?? innerValue
```
marcialfps, line 75: Should we have that comment?

lib/src/slider/Slider.tsx
GomezIvann: After speaking with Raquel and Jairo we decided to change that lines because they weren't working as expected.

lib/src/slider/Slider.tsx
GomezIvann: My fault. Already removed, thanks!!

lib/src/slider/Slider.tsx
marcialfps: 👍🏼

marcialfps (approved):

PR #582, opened 2022-01-31
[Minor] Types for Footer component

Lisa159:


docs/src/pages/components/cdk-components/footer/api.jsx
```diff
@@ -68,7 +68,7 @@ const alertPropsTable = () => {
         <td>margin: string</td>
         <td></td>
         <td>
-          Size of the top margin to be applied to the footer ('xxsmall' |
```
marcialfps, line 71: We should maintain `top margin` comment.

lib/src/footer/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  left?: Space;
+  right?: Space;
+};
+type SocialLink = {
+  logo?: ImgElement;
+  logoSrc?: string;
+  href?: string;
+};
+type BottomLink = {
+  text?: string;
```
marcialfps, line 20: Required prop?

lib/src/footer/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  bottom?: Space;
+  left?: Space;
+  right?: Space;
+};
+type Padding = {
+  top?: Space;
+  bottom?: Space;
+  left?: Space;
+  right?: Space;
+};
```
marcialfps, line 13: Maybe is a good idea to unify in a unique type.

lib/src/footer/types.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  socialLinks?: SocialLink[];
+  /**
+   * An array of objects representing the links that will be rendered at
+   * the bottom part of the footer. Each object has the following
+   * properties:
+   * - text: Text for the link.
+   * - href: URL of the page the link goes to.
+   */
+  bottomLinks?: BottomLink[];
```
marcialfps, line 42: We can use `[BottomLink, ...BottomLink[]]` to avoid the use of empty arrays.

lib/src/footer/types.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  copyright?: string;
+  /**
+   * Value of the tabindex for all interactuable elements, except those
+   * inside the custom area.
+   */
+  tabIndex?: number;
+  /**
+   * Size of the margin to be applied to the component ('xxsmall' |
+   * 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge').
```
marcialfps, line 59: Try to remove the values from the comments since they are already defined in the type.

docs/src/pages/components/cdk-components/footer/api.jsx
Lisa159: Done

lib/src/footer/types.ts
Lisa159: Done

lib/src/footer/types.ts
Lisa159: Done

lib/src/footer/types.ts
Lisa159: Done

marcialfps (approved):

PR #583, opened 2022-01-31
Types for Dialog component

Jialecl:


lib/src/dialog/types.ts
```diff
@@ -0,0 +1,44 @@
...
+   * If true, the dialog will be displayed over a darker background that covers the content behind.
+   */
+  overlay?: boolean;
+  /**
+   * This function will be called when the user clicks background of the modal.
+   * The responsibility of hiding the modal lies with the user.
+   */
+  onBackgroundClick?: () => void;
+  /**
+   * Size of the padding to be applied to the component ('xxsmall' | 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge').
```
marcialfps, line 29: Try to remove the values from the comments since they are already defined in the type.

lib/src/dialog/types.ts
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * This function will be called when the user clicks background of the modal.
+   * The responsibility of hiding the modal lies with the user.
+   */
+  onBackgroundClick?: () => void;
+  /**
+   * Size of the padding to be applied to the component ('xxsmall' | 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge'). 
+   * You can pass an object with 'top', 'bottom', 'left' and 'right' properties in order to specify different padding sizes.
+   */
+  padding?: Padding;
```
marcialfps, line 32: `Padding | Space`

marcialfps (changes requested):

GomezIvann (changes requested):

lib/src/dialog/types.ts
Jialecl: Should we not add the possible values in similar cases? 
I was using the button as a template and it has the possible values in the comments.

GomezIvann: @Jialecl is right. We should add the possible values when the type doesn't explicitly indicate them. This usually happens with Margin type.

marcialfps: Why do the possible values have to be indicated if they are already defined in the type? I think it is redundant to repeat this information.

Jialecl: I think it's helpful to be able to see the possible values while hovering over the parameter

PR #587, opened 2022-02-01
Types for Table component

Lisa159:


lib/src/table/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  /**
+   * Size of the margin to be applied to the component ('xxsmall' | 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge').
+   * You can pass an object with 'top', 'bottom', 'left' and 'right' properties in order to specify different margin sizes.
+   */
+  margin?: Space | Margin;
+  /**
+   * The center section of the table. Can be used to render custom
+   * content in this area.
+   */
+  children?: React.ReactNode;
```
marcialfps, line 19: I think this prop should be required. Also, we can be more restrictive on the type of the children by forcing the user to use `HTMLTableRowElement` (https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableRowElement).

I want to know about your opinion @GomezIvann .

marcialfps (changes requested):

lib/src/table/types.ts
GomezIvann: I agree with this prop being required. About its type, I'm not so sure. I would recommend avoiding HTML types because they are more useful when you are typing the dom tree only using typescript. In React, they lose their utility and can throw errors when they should not because React marks every jsx code as `JSX.Element`, which is the most generic type possible.

marcialfps: That's right, you said so in the last conversation about the SVG type. @Lisa159 , set it as required and keep that type.

Lisa159: Yes, you are right. Changed to required.

marcialfps (approved):

PR #590, opened 2022-02-01
Types for Spinner component

Lisa159:


lib/src/spinner/types.ts
```diff
@@ -0,0 +1,35 @@
+type Space = "xxsmall" | "xsmall" | "small" | "medium" | "large" | "xlarge" | "xxlarge";
+type Margin = {
+  top?: Space;
+  bottom?: Space;
+  left?: Space;
+  right?: Space;
+};
+type Mode = "large" | "small" | "overlay";
```
jsuarezgonz, line 8: Should we avoid creating the 'Mode' type and declare 'Mode' options inline as we did with the Button component, for example?

Lisa159: Yes, you are right. It is changed.

jsuarezgonz (approved):

PR #596, opened 2022-02-02
Added progressbar stories

Lisa159:


aweell (commented):
It seems that the font is not loading correctly (Clearly evident in the helper-text example). I checked other component stories and it looks like (at least) is rendering a sans-serif. Is something we need to check?

Lisa159:
> It seems that the font is not loading correctly (Clearly evident in the helper-text example). I checked other component stories and it looks like (at least) is rendering a sans-serif. Is something we need to check?

Looks like helperText has a wrong token in the code. It is using "fontFamily" token (which doesn't exist in this component) instead of "helperTextFontFamily" token. 
I will create an issue to fix this.

raquelarrojo:
Tests without value will change all the times, so we could add value in all tests to avoid this.

Lisa159:
> Tests without value will change all the times, so we could add value in all tests to avoid this.

Added values to examples.

lib/src/progress-bar/ProgressBar.stories.jsx
```diff
@@ -0,0 +1,58 @@
...
+      <Title title="XxLarge margin" theme="light" level={4} />
+      <DxcProgressBar label="Margin xxlarge" margin="xxlarge" overlay={false} value={50} showValue/>
+    </ExampleContainer>
+  </>
+);
+
+export const ProgressBarOverlay = () => (
+  <ExampleContainer>
+    <Title title="Overlay" theme="light" level={4} />
+    <DxcProgressBar label="Overlay" helperText="Helper text" showValue value={50} />
```
raquelarrojo, line 56: Here should be `theme="dark"`.

raquelarrojo (approved):

PR #603, opened 2022-02-02
Added footer stories

Lisa159:


raquelarrojo:
It could be good to make only one test with socialLinks, bottomLinks, copyright and children, as they do not affect between each other and so we can avoid increment with unnecessary snapshots.

Lisa159:
> It could be good to make only one test with socialLinks, bottomLinks, copyright and children, as they do not affect between each other and so we can avoid increment with unnecessary snapshots.

Done

raquelarrojo (approved):

PR #604, opened 2022-02-02
Added table stories

Lisa159:


marcialfps (commented):
I think it is necessary to add the case where the table has a scrollbar.

Lisa159:
> I think it is necessary to add the case where the table has a scrollbar.

Done

lib/src/table/Table.stories.jsx
```diff
@@ -0,0 +1,284 @@
...
+  </>
+);
+
+// const Container = styled.div`
+//   display: flex;
+//   flex-direction: row;
+//   height: 400px;
+//   width: 100%;
+//   margin-bottom: 50px;
+// `;
```
marcialfps, line 284: Please delete these comments ✌🏼

marcialfps (approved):

lib/src/table/Table.stories.jsx
Lisa159: Done

PR #606, opened 2022-02-02
Added spinner stories

Lisa159:


raquelarrojo:
Default test is not necessary as it is the same as the test for the large mode.
It would be interesting to add an example for the overlay mode.

PR #613, opened 2022-02-03
Adding dialog stories

Jialecl:


lib/src/dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+        dignissim, pharetra neque molestie, molestie lectus.
+      </p>{" "}
+    </DxcDialog>
+  </ExampleContainer>
+);
+
+export const DialogWithPadding = () => (
+  <ExampleContainer>
+    <Title title="Dialog With Padding" theme="light" level={4} />
+    <DxcDialog padding={{top: "xlarge", left: "large", bottom: "medium", right: "small"}}>
```
marcialfps, line 97: It might be a better idea to pass a single value (e.g. `padding="small"`) and have a case for each possible value.

Lisa159: I agree, it would be better to add one example for each padding.

lib/src/dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+        tincidunt. Vestibulum cursus ut risus sit amet congue. Nunc luctus, urna ullamcorper facilisis Jia Le, risus
+        eros aliquam erat, ut efficitur ante neque id odio. Nam orci leo, dignissim sit amet dolor ut, congue gravida
+        enim. Donec rhoncus aliquam nisl, ac cursus enim bibendum vitae. Nunc sit amet elit ornare, malesuada urna eu,
+        fringilla mauris. Vivamus bibendum turpis est, id elementum purus euismod sit amet. Etiam sit amet maximus
+        augue. Vivamus erat sapien, ultricies fringilla tellus id, condimentum blandit justo. Praesent quis nunc
+        dignissim, pharetra neque molestie, molestie lectus.
+      </p>{" "}
+    </DxcDialog>
+  </ExampleContainer>
+);
```
marcialfps, line 92: Repeated story. Delete it pls 👍🏼

marcialfps (changes requested):

Lisa159 (approved):

marcialfps (approved):

PR #614, opened 2022-02-03
Add Switch stories and update component styles

GomezIvann:
Summary:

1. Switch stories added.
2. Updated Switch styles: label text ellipsis, new focus color and tokens value reviewed.


aweell:
badge_refactor_approved

aweell (approved):
badge_refactor_approved

lib/src/switch/Switch.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+  component: DxcSwitch,
+};
+
+export const Chromatic = () => (
+  <>
+    <ExampleContainer>
+      <Title title="With label before" theme="light" level={4} />
+      <DxcSwitch label="Switch" />
+    </ExampleContainer>
+    <ExampleContainer>
```
raquelarrojo, line 20: This test is not necessary as we are already testing the label before the rest of tests. We can avoid repeated tests if we delete it.

lib/src/switch/Switch.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+        </ExampleContainer>
+        <ExampleContainer>
+          <Title title="Disabled checked" theme="dark" level={4} />
+          <DxcSwitch label="Switch" disabled checked labelPosition="after" />
+        </ExampleContainer>
+        <ExampleContainer pseudoState="pseudo-focus">
+          <Title title="Focused" theme="light" level={4} />
+          <DxcSwitch label="Switch" />
+        </ExampleContainer>
+      </DarkContainer>
```
raquelarrojo, line 90: Here the pseudoState is not doing anything as we are overriding the focus styles from material. We can use the focus function like this: (both for light and dark tests)

```
const Switch = () => (
  <ExampleContainer>
    <Title title="Focused" theme="light" level={4} />
    <DxcSwitch label="Switch" />
  </ExampleContainer>
);

export const FocusedSwitch = Switch.bind({});
FocusedSwitch.play = async ({ canvasElement }) => {
  const canvas = within(canvasElement);
  canvas.getByRole("checkbox").focus();
};
```

lib/src/switch/Switch.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+const Switch = () => (
+  <ExampleContainer pseudoState="pseudo-focus">
+    <Title title="Focused" theme="light" level={4} />
+    <DxcSwitch label="Switch" />
+  </ExampleContainer>
+);
+export const FocusedSwitch = Switch.bind({});
+FocusedSwitch.play = async () => {
+  await userEvent.tab();
+};
```
raquelarrojo, line 154: As described in the comment above we can use the focus function instead of using the tab.

lib/src/switch/Switch.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+    </ExampleContainer>
+    <ExampleContainer>
+      <Title title="With long word" theme="light" level={4} />
+      <DxcSwitch
+        label="Large texttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt"
+        labelPosition="after"
+        size="medium"
+      />
+    </ExampleContainer>
+    <ExampleContainer>
```
raquelarrojo, line 40: We should test the ellipsis for medium size also. We can add these ellipsis tests in the sizes test section as in the button component, for example.

lib/src/switch/Switch.stories.tsx
GomezIvann: This was a residual example I forgot to remove. I will add a new story for dark focused Switch!

lib/src/switch/Switch.stories.tsx
GomezIvann: I agree!

lib/src/switch/Switch.stories.tsx
```diff
@@ -0,0 +1,160 @@
...
+      <Title title="Large size" theme="light" level={4} />
+      <DxcSwitch label="Large" size="large" />
+    </ExampleContainer>
```
raquelarrojo, line 121: We could also add an example with large size and ellipsis.

raquelarrojo: Not needed, with ellipsis in one size it's tested!

raquelarrojo (approved):

PR #617, opened 2022-02-03
Added spinner stories

Lisa159:


raquelarrojo:
We could add a test for small mode with value.

raquelarrojo:
We also need a dark example.

Lisa159:
> We could add tests for both modes with value 100% to see the entire spinner, and other with other values. Remove the tests without value because the snapshots are going to change all the times.

Done

Lisa159:
> We also need dark examples.

Overlay mode should be the dark examples.

raquelarrojo (approved):

PR #620, opened 2022-02-03
Added tag stories

raquelarrojo:
- Added tag stories.
- Fix fillParent size.


jsuarezgonz (commented):
I think it might be good to have a test of a Tag with a larger image as an icon to test that it scales properly.

raquelarrojo:
> I think it might be good to have a test of a Tag with a larger image as an icon to test that it scales properly.
Done!

jsuarezgonz (approved):

PR #626, opened 2022-02-04
Added button types

raquelarrojo:


docs/src/pages/components/cdk-components/button/api.jsx
```diff
@@ -29,7 +29,7 @@ const buttonPropsTable = () => {
         <td>Text to be placed next to the button.</td>
       </tr>
       <tr>
-        <td>icon: node</td>
```
GomezIvann, line 32: I would continue using `node` as the type of the svgs, because, even if we are typing it, React treats this prop as a `JSX.Element` type, so anything can go inside. In the future, when types at doc are auto-generated we can change this.

raquelarrojo: I agree! It is changed to `node` again.

Lisa159 (approved):

GomezIvann (approved):

PR #631, opened 2022-02-04
Add DateInput stories

GomezIvann:


jsuarezgonz (commented):
- We should add tests for margins.
- We should add tests for sizes.

Could it be interesting to add a test by setting the focus on one of the calendar days in the same way that now the focus is set on the year by the tab function? And on the arrows to change the month?

jsuarezgonz (approved):

jsuarezgonz (changes requested):
Chromatic screenshots are not working properly, we need to increase the height of the body to capture them correctly.

jsuarezgonz (approved):

PR #635, opened 2022-02-07
Added Dropdown stories

### LINKED ISSUE #541: Add visual regression tests for Dropdown component

marcialfps:
Closes #541

### DIFF
--- a/lib/src/dropdown/Dropdown.stories.tsx
+++ b/lib/src/dropdown/Dropdown.stories.tsx
@@ -1,0 +1,185 @@
+import React from "react";
+import { userEvent, within } from "@storybook/testing-library";
+import DxcDropdown from "./Dropdown";
+import Title from "../../.storybook/components/Title";
+import ExampleContainer from "../../.storybook/components/ExampleContainer";
+
+export default {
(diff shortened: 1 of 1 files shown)


marcialfps:
I decided to separate the different cases in other stories, so snapshots are prepared to fix the bug with the ClickAwayListener. A new issue is opened to fix the detected bug (#679).

Can we continue with the review of this component @jsardid ?

jsuarezgonz (changes requested):
- We should add an example with ellipsis in the dropdown label and another one with ellipsis in one of the options.

- Add an example using an icon bigger than 24x24 to check that it scales properly.

- Example with icon after each option.

- In the 'Hovered/Actived Options' examples we should leave only one option. (As we did in select or text input).

marcialfps:
> * We should add an example with ellipsis in the dropdown label and another one with ellipsis in one of the options.

Impossible to reproduce the second case, currently the options have no ellipsis.

marcialfps:
The ellipsis is not included in the design guidelines, so I have opened an issue: https://github.com/dxc-technology/halstack-style-guide/issues/734

PR #637, opened 2022-02-07
Added Sidenav stories

Jialecl:


marcialfps (commented):
Nice!!! 💯 I think we can add a case where the sidenav link is focused to verify the `linkFocusColor` theme input. 
image

Lisa159 (approved):

PR #644, opened 2022-02-07
Added box types

raquelarrojo:
- Added box types.
- Added tabIndex to box.


lib/src/box/types.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  shadowDepth?: 0 | 1 | 2;
+  /**
+   * Changes the display CSS property of the box div.
+   */
+  display?: string;
+  /**
+   * Custom content that will be placed in the box component.
+   */
+  children?: React.ReactNode;
```
marcialfps, line 27: It should be required.

marcialfps (changes requested):

marcialfps (approved):

PR #645, opened 2022-02-07
Added Slider stories

GomezIvann:


lib/src/slider/Slider.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+    </ExampleContainer>
+    <ExampleContainer>
+      <Title title="Xxlarge" theme="light" level={4} />
+      <DxcSlider label="Xxlarge" margin="xxlarge" />
+    </ExampleContainer>
+    <Title title="Sizes" theme="light" level={2} />
+    <ExampleContainer>
+      <Title title="Large" theme="light" level={4} />
+      <DxcSlider label="Large" size="medium" />
+    </ExampleContainer>
```
raquelarrojo, line 162: Title and slider label here should be `Medium`.

lib/src/slider/Slider.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+    <ExampleContainer>
+      <Title title="Xlarge" theme="light" level={4} />
+      <DxcSlider label="Xlarge" size="large" />
+    </ExampleContainer>
```
raquelarrojo, line 166: Title and slider label here should be `Large`.

lib/src/slider/Slider.stories.tsx
```diff
@@ -0,0 +1,172 @@
...
+    <ExampleContainer>
+      <Title title="Xxlarge" theme="light" level={4} />
+      <DxcSlider label="Xxlarge" size="fillParent" />
+    </ExampleContainer>
```
raquelarrojo, line 170: Title and slider label here should be `FillParent`.

raquelarrojo:
We could also add a test for `labelFormatCallback`.

lib/src/slider/Slider.stories.tsx
GomezIvann: Done!

lib/src/slider/Slider.stories.tsx
GomezIvann: Done!

lib/src/slider/Slider.stories.tsx
GomezIvann: Done!

raquelarrojo:
> Yes, you're right. I think it's a non valuable test.

PR #646, opened 2022-02-07
Added wizard stories

Lisa159:


Lisa159:
> I think it might be interesting to have a test with an icon larger than 24x24 to prove that it rescales well.

Done.

Lisa159:
When all steps are disabled, the selected one is not taking correct styles. Open issue -> #653

jsuarezgonz (approved):

PR #649, opened 2022-02-08
Added Slider stories

GomezIvann:
`labelFormatCallback` type is also fixed in this PR.


GomezIvann:
Following here [this conversation](https://github.com/dxc-technology/halstack-react/pull/645#issuecomment-1032715626):

image

PR #650, opened 2022-02-08
Fixed Slider limit values size and storie added for this scenario

GomezIvann:


raquelarrojo (commented):
After this change, the max value label is not totally visible. Seems to be something related to the `overflow`.

image

GomezIvann:
> After this change, the max value label is not totally visible. Seems to be something related to the `overflow`.
> 
> image

Fixed!

raquelarrojo (approved):

PR #660, opened 2022-02-10
Added Header stories

### LINKED ISSUE #565: Add visual regression tests for Header component

marcialfps:
Closes #565


jsuarezgonz (commented):
- I think it might be a good idea to add some text after the Header to see how the margin prop affects it.
- All the examples have a dropdown as content, I think we could leave only one example with content and leave the others without it

PR #665, opened 2022-02-10
Added textInput types.

raquelarrojo:
- Added textInput types.
- Modified input value of the slider to be passed as a string instead of a number.

### DIFF
--- a/docs/src/pages/components/cdk-components/text-input/api.jsx
+++ b/docs/src/pages/components/cdk-components/text-input/api.jsx
@@ -49,7 +49,9 @@ const textInputPropsTable = () => {
           Action to be shown in the input. This is an object composed of an
           onClick function and an icon, being the latter either an inline svg or
           a URL to the image. An example of this object is: {"{ "}
-          <code>onClick: onClickFunction, icon: string | svgIcon</code>
+          <code>
+            onClick: onClickFunction, icon: string | svgIcon, title: string
+          </code>
           {" }"}.
         </td>
       </tr>
--- a/docs/src/pages/components/cdk-components/text-input/api.jsx
+++ b/docs/src/pages/components/cdk-components/text-input/api.jsx
@@ -49,9 +49,7 @@ const textInputPropsTable = () => {
           Action to be shown in the input. This is an object composed of an
           onClick function and an icon, being the latter either an inline svg or
           a URL to the image. An example of this object is: {"{ "}
-          <code>
-            onClick: onClickFunction, icon: string | svgIcon, title: string
-          </code>
+          <code>onClick: function, icon: string | svgIcon, title: string</code>
           {" }"}.
         </td>
       </tr>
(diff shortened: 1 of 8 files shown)


lib/src/number-input/NumberInputContext.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+interface INumberInputContext {
```
GomezIvann, line 3: I think we should use a type instead of an interface, as we are currently only using types.

docs/src/pages/components/cdk-components/text-input/api.jsx
```diff
@@ -49,7 +49,9 @@ const textInputPropsTable = () => {
           Action to be shown in the input. This is an object composed of an
           onClick function and an icon, being the latter either an inline svg or
           a URL to the image. An example of this object is: {"{ "}
-          <code>onClick: onClickFunction, icon: string | svgIcon</code>
+          <code>
+            onClick: onClickFunction, icon: string | svgIcon, title: string
```
GomezIvann, line 53: For consistency, I think it would be better to use `function` instead of `onClickFunction`, as it is its type in javascript and the one used in the rest of the props of this kind.

lib/src/text-input/types.tsx
```diff
@@ -67,14 +76,14 @@ type Props = {
...
+  onChange?: (val: { value: string; error: string | null }) => void;
   /**
    * This function will be called when the input element loses the focus.
    * An object including the input value and the error (if the value
-   * entered is not valid) will be passed to this function. If there is no error, 
+   * entered is not valid) will be passed to this function. If there is no error,
    * error will be null.
    */
-  onBlur?: (obj: { value: string; error: string }) => void;
+  onBlur?: (obj: { value: string; error: string | null }) => void;
```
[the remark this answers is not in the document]
raquelarrojo: Done.

docs/src/pages/components/cdk-components/text-input/api.jsx
raquelarrojo: Done.

lib/src/number-input/NumberInputContext.tsx
raquelarrojo: Done.

lib/src/text-input/types.tsx
```diff
@@ -87,25 +96,25 @@ type Props = {
...
-    * reached, the error parameter of both events will be null.
+   * Specifies the minimun and maximum length allowed by the input.
+   * This will be checked both when the input element loses the
+   * focus and while typing within it. If the string entered does not
+   * comply the length, the onBlur and onChange functions will be called
+   * with the current value and an internal error informing that the value
+   * length does not comply the specified range. If a valid length is
+   * reached, the error parameter of both events will be null.
    */
   length?: { min?: number; max?: number };
```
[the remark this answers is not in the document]
raquelarrojo: Done!

GomezIvann (approved):

PR #672, opened 2022-02-10
Added toggleGroup types

### LINKED ISSUE #594: Add types for ToggleGroup component

jsuarezgonz:
Closes #594


lib/src/toggle-group/ToggleGroup.tsx
```diff
@@ -1,30 +1,30 @@
...
-  options = [],
+  options,
   margin,
   multiple = false,
   tabIndex = 0,
-}) => {
+}: ToogleGroupPropsType): JSX.Element => {
   const colorsTheme = useTheme();
-  const [selectedValue, setSelectedValue] = useState(multiple ? [] : null);
+  const [selectedValue, setSelectedValue] = useState(multiple ? ([] as string[]) : "");
```
GomezIvann, line 20: Is this cast really necessary? I think it would work with just `useState(multiple ? [] : "");`

lib/src/toggle-group/ToggleGroup.tsx
```diff
@@ -35,7 +35,7 @@ const DxcToggleGroup = ({
         setSelectedValue(newSelectedOptions);
       } else setSelectedValue(selectedOption === selectedValue ? null : selectedOption);
     } else if (multiple) {
-      newSelectedOptions = value.map((v) => v);
+      newSelectedOptions = Array.isArray(value) ? value.map((v) => v) : value;
```
GomezIvann, line 38: If it is multiple, `value` should be an array, what's the purpose of iterating throw it to just not change anything (`map(v => v)`)? I would just put `newSelectedOptions = value` and not check anything.

lib/src/toggle-group/ToggleGroup.tsx
```diff
@@ -1,30 +1,30 @@
...
+  const [selectedValue, setSelectedValue] = useState(multiple ? ([] as string[]) : "");
   const [toggleGroupId] = useState(`toggle-group-${uuidv4()}`);
 
   const handleToggleChange = (selectedOption) => {
     let newSelectedOptions;
 
     if (value == null) {
-      if (multiple) {
+      if (multiple && Array.isArray(selectedValue)) {
         newSelectedOptions = selectedValue.map((value) => value);
```
GomezIvann, line 28: Why are we iterating to not change anything? I think its better to just write `selectedOptions = selectedValue;`.

lib/src/toggle-group/types.ts
```diff
@@ -0,0 +1,73 @@
...
+  left?: Space;
+  right?: Space;
+};
+type SVG = React.SVGProps<SVGSVGElement> | React.FunctionComponent<React.SVGProps<SVGSVGElement>>;
+
+type Option={
+    /**
+    * Number with the option inner value.
+    */
+    value?: string;
```
GomezIvann, line 14: I think `value` should be mandatory.

lib/src/toggle-group/types.ts
```diff
@@ -0,0 +1,73 @@
...
+    /**
+    * String with the option display value.
+    */
+    label?: string;
```
GomezIvann, line 18: I consider `label` or `icon` a mandatory prop. I mean, an Option must have one of them, or both, but never any of them. For further understanding: 

image

You have an example at the Link component with `icon` and `label` props.

lib/src/toggle-group/types.ts
```diff
@@ -0,0 +1,73 @@
...
+   * it must be one unique value. If the component allows multiple selection, value must be an array. 
+   * If undefined, the component will be uncontrolled and the value will be managed internally by the component.
+   */
+  value?: string | string[];
+  /**
+   * This function will be called every time the selection changes. The number with the key of the selected 
+   * value will be passed as a parameter to this function. 
+   * If multiple selection is allowed, an array of keys will be passed.
+   */
+  onChange?:(optionIndex: number | number[]) => void;
```
GomezIvann, line 49: I think we should talk with @jsardid and @aweell about the `onChange`. Perhaps it makes sense to follow the same approach as Select and use the `value: string | string[]` as the parameter of this event.

GomezIvann (changes requested):

lib/src/toggle-group/ToggleGroup.tsx
jsuarezgonz: Done.

lib/src/toggle-group/ToggleGroup.tsx
jsuarezgonz: I think we can leave this change for a future revision of the component.

lib/src/toggle-group/ToggleGroup.tsx
jsuarezgonz: I think we can leave this change for a future revision of the component.

lib/src/toggle-group/types.ts
jsuarezgonz: Done.

lib/src/toggle-group/types.ts
jsuarezgonz: Done.

lib/src/toggle-group/ToggleGroup.tsx
GomezIvann: Perfect! I agree.

lib/src/toggle-group/ToggleGroup.tsx
GomezIvann: Agree.

PR #677, opened 2022-02-11
Adding App Layout types

Jialecl:


docs/src/pages/components/cdk-components/applicationLayout/api.jsx
```diff
@@ -15,16 +15,29 @@ const SidenavApplicationLayoutPropsTable = () => {
...
+      </tr>
+      <tr>
+        <td>padding: string | object </td>
+        <td></td>
+        <td>
+          Size of the padding to be applied to the custom area ('xxsmall' |
+          'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge'). You
+          can pass an object with 'top', 'bottom', 'left' and 'right' properties
+          in order to specify different padding sizes.
+        </td>
```
marcialfps, line 40: This prop is from DxcSidenav. Maybe is not necessary to indicate it since we link to DxcSidenav props.

lib/src/layout/types.ts
```diff
@@ -0,0 +1,62 @@
...
+   * In lower resolutions the mode will always be overlay.
+   */
+  mode: "push" | "overlay";
+};
+
+type AppLayoutPropsType = {
+  /**
+   * The area inside the sidenav. This area can be used to render custom content.
+   */
+  children: React.ReactNode;
```
marcialfps, line 59: Maybe we can be more restrective indicating children types like in the Accordion group. 
Example: 
`children: React.ReactElement<AppLayoutHeaderPropsType> | React.ReactElement<AppLayoutMainPropsType> | React.ReactElement<AppLayoutFooterPropsType> | React.ReactElement<AppLayoutSidenavPropsType>;`

marcialfps (changes requested):
I'm not sure about the way we use children prop. @GomezIvann can u take a look? I think is similar to the Accordion group. 🤔

lib/src/layout/types.ts
GomezIvann: I agree with the idea of restricting this prop, but the type I don't think is correct. I suggest something like this:

```
type AppLayoutChildrenType = AppLayoutHeaderPropsType | AppLayoutMainPropsType | AppLayoutFooterPropsType | AppLayoutSidenavPropsType;
type AppLayoutPropsType = {
  children: React.ReactElement<AppLayoutChildrenType> | React.ReactElement<AppLayoutChildrenType>[];
};
```

GomezIvann:
> I'm not sure about the way we use children prop. @GomezIvann can u take a look? I think is similar to the Accordion group. 🤔

Above I give my suggestion for this case, agree with Marcial.

GomezIvann (dismissed):

docs/src/pages/components/cdk-components/applicationLayout/api.jsx
marcialfps: What about this @Jialecl ?

marcialfps (approved):

PR #681, opened 2022-02-14
Added result set table types

Lisa159:
New prop **showGoToPage** will be added after merging this PR.


lib/src/resultsetTable/types.ts
```diff
@@ -0,0 +1,84 @@
...
+  top?: Space;
+  bottom?: Space;
+  left?: Space;
+  right?: Space;
+};
+type Column = {
+  /**
+   * Column display value.
+   */
+  displayValue;
```
marcialfps, line 12: React.ReactNode?

lib/src/resultsetTable/types.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+type Row = {
+  /**
+   * Value to be displayed in the cell.
+   */
+  displayValue;
```
marcialfps, line 18: React.ReactNode?

lib/src/resultsetTable/ResultsetTable.tsx
```diff
@@ -54,15 +54,15 @@ const BothArrows = () => (
   </svg>
 );
 
-const DxcResultsetTable = ({
+const DxcResultsetTable: Overload = ({
```
marcialfps, line 57: Why do u convert JSX.Element to Overload type?

lib/src/resultsetTable/types.ts
Lisa159: Done

lib/src/resultsetTable/types.ts
Lisa159: Done

lib/src/resultsetTable/ResultsetTable.tsx
Lisa159: Removed pattern.

lib/src/resultsetTable/types.ts
```diff
@@ -9,28 +9,34 @@ type Column = {
...
-  isSortable: boolean;
-};
-type RowSortable = Row & {
+  displayValue: React.ReactNode;
   /**
    * Value to be used when sorting the table by that
    * column. If not indicated displayValue will be used for sorting.
    */
-  sortValue: boolean;
+  sortValue?: boolean;
```
marcialfps, line 27: This prop is not boolean, I think it must be `string` type.

marcialfps (approved):

PR #683, opened 2022-02-14
Added textInput stories

raquelarrojo:
- Added textInput stories.
- Added ellipsis to input and suggestions.


lib/src/text-input/TextInput.stories.tsx
```diff
@@ -0,0 +1,418 @@
...
+import DxcTextInput from "./TextInput";
+
+export default {
+  title: "Text input",
+  component: DxcTextInput,
+};
+
+const action = {
+  onClick: () => {
+    console.log("Copy that!");
```
jsuarezgonz, line 17: I think we can remove this 'console.log' and leave an empty function.

jsuarezgonz (commented):
- It might be interesting to add an example to see how the disabled state affects the prefix and suffix.

- Add an example with a bigger icon in an action to check that it scales properly.

lib/src/text-input/TextInput.stories.tsx
raquelarrojo: Agree! Done!

raquelarrojo:
> * It might be interesting to add an example to see how the disabled state affects the prefix and suffix.
> * Add an example with a bigger icon in an action to check that it scales properly.

Done!

jsuarezgonz (approved):

PR #691, opened 2022-02-15
File-input visual fixes

GomezIvann:
Summary
- Updated clear focus.
- Error message fixed in single file selection.
- Single file selection alignment fixed


raquelarrojo (approved):

raquelarrojo (changes requested):
After removing min-height of error, the select in paginator is not aligned.

image

aweell:
> After removing min-height of error, the select in paginator is not aligned.
> 
> 
> 
> image
> 
> 

How the file-input changes affect the select in the paginator?

GomezIvann:
> > After removing min-height of error, the select in paginator is not aligned.
> > image
> 
> How the file-input changes affect the select in the paginator?

This issue is caused by the change we did at the Select while correcting the `line-height`. By removing the min-height of the error message its reserved space now is gone. I fix it.

image

raquelarrojo (approved):

PR #692, opened 2022-02-15
[Minor] Added new prop showGoToPage in Resultset Table

Lisa159:


docs/src/pages/components/cdk-components/resultsetTable/api.jsx
```diff
@@ -48,6 +48,13 @@ const resultsetTablePropsTable = () => {
           </ul>
         </td>
       </tr>
+      <tr>
+        <td>showGoToPage: boolean</td>
+        <td>
+          <code>true</code>
+        </td>
+        <td>Show page navigation select.</td>
```
GomezIvann, line 56: I think we should extend the explanation. Something like: "If true, a Select for navigation between pages will be displayed".

Lisa159: Done.

GomezIvann (approved):

PR #696, opened 2022-02-15
Added Accordion stories

GomezIvann:
Summary:

- Labels are now mandatory in both Accordion and Accordion Group Accordions.
- Added Accordion stories.


jsuarezgonz (commented):
- We could add an example of an open and disabled accordion to test how its content is affected.
- We could add an example with a bigger icon to test that the icon is resized properly.

jsuarezgonz (approved):

PR #702, opened 2022-02-17
Change array type in some components

GomezIvann:
Changing old type for arrays (`[T, ...T]`), which was causing problems, to a less restrictive one (`T[]`).

### DIFF
--- a/lib/src/dropdown/types.ts
+++ b/lib/src/dropdown/types.ts
@@ -31,7 +31,7 @@ type Props = {
   /**
    * An array of objects representing the options.
    */
-  options: [Option, ...Option[]];
+  options: Option[];
   /**
    * In case options include icons, whether the icon should appear
    * after or before the label.
--- a/lib/src/footer/types.ts
+++ b/lib/src/footer/types.ts
@@ -5,35 +5,43 @@ type Size = {
   left?: Space;
   right?: Space;
 };
+type SVG = React.SVGProps<SVGSVGElement> | React.FunctionComponent<React.SVGProps<SVGSVGElement>>;
 type SocialLink = {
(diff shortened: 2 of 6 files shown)


GomezIvann:
> Some "errors" I have found:
> 
> * Rename footer and wizard stories to `.tsx`.
> * The are some type errors in Toggle group stories.
> 
> I know these "errors" are not from your changes, but I think is a good opportunity to fix all of them.

I agree with you Martial, thanks for the effort. I will address these issues but I think we should do it in another branch to avoid mixing topics, what do you think?

marcialfps:
> > Some "errors" I have found:
> > 
> > * Rename footer and wizard stories to `.tsx`.
> > * The are some type errors in Toggle group stories.
> > 
> > I know these "errors" are not from your changes, but I think is a good opportunity to fix all of them.
> 
> I agree with you Martial, thanks for the effort. I will address these issues but I think we should do it in another branch to avoid mixing topics, what do you think?

Of course!

PS, my name is Marcial, no Martial!! 🤣

marcialfps (approved):

GomezIvann:
> > > Some "errors" I have found:
> > > 
> > > * Rename footer and wizard stories to `.tsx`.
> > > * The are some type errors in Toggle group stories.
> > > 
> > > I know these "errors" are not from your changes, but I think is a good opportunity to fix all of them.
> > 
> > 
> > I agree with you Martial, thanks for the effort. I will address these issues but I think we should do it in another branch to avoid mixing topics, what do you think?
> 
> Of course!
> 
> PS, my name is Marcial, no Martial!! 🤣

For Grammarly app you are from now on Martial

image

PR #704, opened 2022-02-17
[Minor] Added title & aria-label to textInput actions.

raquelarrojo:


lib/src/text-input/TextInput.tsx
```diff
@@ -471,6 +471,7 @@ const DxcTextInput = React.forwardRef<RefType, TextInputPropsType>(
                 }}
                 backgroundType={backgroundType}
                 tabIndex={tabIndex}
+                title="Clear"
```
GomezIvann, line 474: I suggest being a little more explicit. Maybe something like "Clear text" would be more useful to the user.

lib/src/text-input/TextInput.tsx
```diff
@@ -487,6 +488,7 @@ const DxcTextInput = React.forwardRef<RefType, TextInputPropsType>(
                   }}
                   backgroundType={backgroundType}
                   tabIndex={tabIndex}
+                  title="Decrement"
```
GomezIvann, line 491: Same as above. Maybe something like "Decrement number" would be more useful to the user.

lib/src/text-input/TextInput.tsx
```diff
@@ -500,6 +502,7 @@ const DxcTextInput = React.forwardRef<RefType, TextInputPropsType>(
                   }}
                   backgroundType={backgroundType}
                   tabIndex={tabIndex}
+                  title="Increment"
```
GomezIvann, line 505: Same as above. Maybe something like "Increment number" would be more useful to the user.

lib/src/text-input/TextInput.tsx
```diff
@@ -515,6 +518,7 @@ const DxcTextInput = React.forwardRef<RefType, TextInputPropsType>(
                     event.stopPropagation();
                   }}
                   title={action.title ?? action.title}
```
GomezIvann, line 520: Why are we checking nullish/undefined values? This would behave the same as `title={action.title}` because when the prop is not defined, title will also be undefined and that is what we are trying to achieve.

GomezIvann (changes requested):

lib/src/text-input/TextInput.tsx
raquelarrojo: What do you think @aweell? Regarding the titles of the clear, increment and decrement actions, a one simple title or a more detailed one?

aweell: I agree that we need to follow the formula: `[verb (action)] + [sustantive (the recipient of the action]`

For the input-text clear action: * "Clear field"
For the input-number stepper buttons:

* "Increment value"
* "Decrement value"

raquelarrojo: Perfect! I change this!

lib/src/text-input/TextInput.tsx
raquelarrojo: Change by "Decrement value".

lib/src/text-input/TextInput.tsx
raquelarrojo: Changed by "Increment value".

lib/src/text-input/TextInput.tsx
raquelarrojo: Agree. Changed!

GomezIvann (approved):

PR #710, opened 2022-02-18
Types for ThemeContext and BackgroundColorContext

GomezIvann:
Summary:

1. Adding types for these three files.
2. Removing .js extension from every import that referred these files.


lib/src/ThemeContext.tsx
```diff
@@ -274,14 +274,28 @@ const parseTheme = (theme) => {
...
-  );
+type ThemeProviderCommonProps = {
+  children: React.ReactNode;
+};
+type ThemeProviderDefaultTheme = ThemeProviderCommonProps & {
+  theme?: object;
+  advancedTheme?: never;
+};
+type ThemeProviderAdvancedTheme = ThemeProviderCommonProps & {
+  theme?: never;
```
marcialfps, line 285: I don't understand the use of `never` type.

GomezIvann: `never` is a special type used to represent impossible combinations. In this case, I'm using it to guarantee that an advanced theme is never passed at the same time as a default theme, and the same in the opposite case. For example:

```
type ThemeProviderDefaultTheme = ThemeProviderCommonProps & {
  theme?: object;
  advancedTheme?: never;
};
```

This type represents the situation when the ThemeProvider receives a regular `theme`. By giving `advancedTheme` the type `never` I'm saying that this prop should be received in this case.

marcialfps: And why did you discard the first solution? Is it not the same?
```
type ThemeProviderDefaultTheme = ThemeProviderCommonProps & {
  theme: object;
};
```

marcialfps (approved):

marcialfps:
Please, resolve the conflict if possible. Thanks! ✌🏼

GomezIvann:
> Please, resolve the conflict if possible. Thanks! ✌🏼

Resolved!

marcialfps (approved):

marcialfps:
We have another conflict lol

GomezIvann:
> We have another conflict lol

Ready again! We need to merge this quickly to avoid conflicts

image

marcialfps (approved):

PR #723, opened 2022-02-23
[Minor] Fixing dialog doc padding default value

Jialecl:
It seems that when no padding was used, the component was using spaces["small"] as the default value.
The documentation has been updated to reflect that.


GomezIvann:
Why are we adding a default value for padding? The same scenario happens in the Accordion component and it doesn't have a value by default.

Jialecl:
> Why are we adding a default value for padding? The same scenario happens in the Accordion component and it doesn't have a value by default.


The default value used in the css is spaces["small"].
This is the same as having small padding as default.

padding: ${(props) =>
      props.padding && typeof props.padding !== "object" ? spaces[props.padding] : spaces["small"]};

GomezIvann:
> > Why are we adding a default value for padding? The same scenario happens in the Accordion component and it doesn't have a value by default.
> 
> The default value used in the CSS is spaces["small"]. This is the same as having small padding as default.
> 
> padding: ${(props) => props.padding && typeof props.padding !== "object" ? spaces[props.padding] : spaces["small"]};
> 
> I think this should be documented as such.

My question goes a bit further, why it was set as `small` by default? I don't think we should establish a default padding, what do you think @aweell?

Jialecl:
I understand your concerns but I think that can be a different discussion. 
This change is aimed to update the documentation to reflect how the component is working now.

GomezIvann:
> My question goes a bit further, why it was set as `small` by default? I don't think we should establish a default padding, what do you think @aweell?

Totally agree with this Jiale!

GomezIvann (approved):

PR #725, opened 2022-02-24
[Minor] Wizard stories file extension changed to .tsx and removed iconSrc prop

GomezIvann:


lib/src/wizard/Wizard.tsx
```diff
@@ -2,9 +2,50 @@ import React, { useState } from "react";
 import styled, { ThemeProvider } from "styled-components";
 import { spaces } from "../common/variables.js";
 import useTheme from "../useTheme";
-import { validIcon, invalidIcon } from "./Icons";
 import WizardPropsType from "./types";
 
+const icons = {
```
marcialfps, line 7: Maybe it is better to keep all these icons out of the component?

marcialfps (approved):

lib/src/wizard/Wizard.tsx
GomezIvann: I thought they were small enough to have them inside the component and not in an external file!

marcialfps: I think the opposite, there are too many icons and they take too much space in the code. Anyway, I will merge, we can easily change this in the future. ✌🏼

PR #729, opened 2022-02-24
Added textarea types

### LINKED ISSUE #517: Add types for Textarea component

jsuarezgonz:
Closes #517


raquelarrojo (changes requested):
Error in onChange and onBlur should admit also null value:
  `onChange?: (val: { value: string; error: string | null }) => void;`
  `onBlur?: (val: { value: string; error: string | null }) => void;`

Size type should be put like this to follow the same convention as in other components:
`size?: "small" | "medium" | "large" | "fillParent";`

In the table props size should be:
`size: string`  instead of `size: string | object`

raquelarrojo (approved):

PR #731, opened 2022-02-25
[Minor] Added disabled prop in tag component

Lisa159:


aweell:
What's the difference between the Storybook's examples "with icon" and "with large icon"?

Screen Shot 2022-02-25 at 3 40 19 PM

Lisa159:
> What's the difference between the Storybook's examples "with icon" and "with large icon"?
> 
> Screen Shot 2022-02-25 at 3 40 19 PM

Large icon is an icon bigger than the icon size from the specs. It is added to check it is escalating well (from 48px to 24px):

image

lib/src/tag/Tag.stories.tsx
```diff
@@ -56,6 +56,36 @@ export const Chromatic = () => (
...
+      <DxcTag
+        label="Tag"
+        icon={icon}
+        onClick={() => {
+          console.log("click");
+        }}
+        disabled={true}
+      />
+    </ExampleContainer>
+    <ExampleContainer pseudoState="pseudo-focus">
```
GomezIvann, line 74: I think this test is functional rather than visual, we should have this in the jest set of tests.

GomezIvann: With a disabled test for the basic tag, tag with action and tag with the link I think it's enough.

GomezIvann (changes requested):

GomezIvann:
More things to add to my review:

- Disabled Tag with `linkHref`, which displays as an anchor, works when it is clicked.
- Focus now doesn't work as usual, it displays the default HTML focus instead of our custom one. Also, the focus works with the disabled state:

image

image

As expected:

aweell:
> Does the disabled state make sense when the component is not clickable (anchor or link) and doesn't have hover or focus? What do you think about this @aweell?

If there's no interaction doesn't make much sense to have a disabled state.

GomezIvann (changes requested):

lib/src/tag/Tag.stories.tsx
GomezIvann: My proposal:

- Visual tests (storybook): Disabled as action, disabled as Link. We can keep having disabled + focused visual tests
- Jest tests: disabled tag is not clickable, both as a link and as an action (button).

PR #733, opened 2022-02-28
[Patch] Updated copyright and added default value

Lisa159:


lib/src/footer/Footer.tsx
```diff
@@ -10,7 +10,7 @@ import FooterPropsType from "./types";
 const DxcFooter = ({
   socialLinks,
   bottomLinks,
-  copyright = "",
+  copyright = "© DXC Technology 2022. All rights reserved.",
```
GomezIvann, line 13: We keep having the same problem as before with this approach (update every year the copyright message). My proposal: 

Sin título

And as the default value for the prop don't give anyone.

docs/src/pages/overview/Overview.jsx
```diff
@@ -368,7 +368,7 @@ function Overview() {
             { text: "Facebook", href: "http://www.google.com" },
             { text: "Instagram", href: "http://www.google.com" },
           ]}
-          copyright="© DXC Technology 2020. All rights reserved."
+          copyright="© DXC Technology 2022. All rights reserved."
```
GomezIvann, line 371: Remove this line. This Footer (and every other one in our CDK must use the preset value).

docs/src/pages/components/Components.jsx
```diff
@@ -156,7 +156,7 @@ function Components() {
             { text: "Facebook", href: "http://www.google.com" },
             { text: "Instagram", href: "http://www.google.com" },
           ]}
-          copyright="© DXC Technology 2020. All rights reserved."
+          copyright="© DXC Technology 2022. All rights reserved."
```
GomezIvann, line 159: Remove this line. This Footer (and every other one in our CDK must use the preset value).

GomezIvann (changes requested):

GomezIvann:
Adding to the review one more thing:

- Remove every `copyright` appearance in the CDK docs, it must now use the preset value.

image

lib/src/footer/Footer.tsx
Lisa159: Done

docs/src/pages/components/Components.jsx
Lisa159: Done

docs/src/pages/overview/Overview.jsx
Lisa159: Done

GomezIvann (approved):

PR #734, opened 2022-02-28
[Patch] Footer logos revision

GomezIvann:
Summary:

- Update logos HTML structure as in the other components.
- Update the` logo` token value to fix theme-builder issue (was not rendering and throwing an error). This also happened in the Header examples, whose tokens are also updated (`logo` and `logoResponsive`).


marcialfps (approved):
💯

PR #736, opened 2022-03-02
Types for Select component

GomezIvann:
Also, changed the `label` prop in the Checkbox component from mandatory to optional, as we internally use it in the multiple select.


marcialfps (changes requested):
Change in props table the size type to `string`.

GomezIvann:
> Change in props table the size type to `string`.

Fixed!

marcialfps (approved):

PR #746, opened 2022-03-10
[Minor] Initial version of new component Radio Group

GomezIvann:
Summary:

- Define the structure of the components (main Radio Group and Radio(s)).
- Establish the first version of the HTML structure.
- First iteration of the component states (hover, active, focus, error, etc).
- First iteration of the component prop types.
- First iteration of the roving tabindex. Now, when the radio group gains the focus, the focused radio is the checked one and, in the case there's no one, it goes to the first option.
- Props already implemented (some of them may lack some functionality):
```
  - label
  - name
  - helperText
  - options
  - disabled (may be incomplete)
  - optional
  - optionalOptionLabel (set to "None" by default)
  - stacking = "column"
  - error
  - defaultValue
```


marcialfps (approved):
🔥 🔥

PR #748, opened 2022-03-11
Several fixes in stories

raquelarrojo:
- Fix dark test of the accordion.
- Added `expanded` prop to the `ExampleContainer` since some components were not seen in its entirety in the snapshots.
- Change some `smmall` titles to `small`.
- Added `< hr/>` in some stories in order to get the snapshot of the whole component with margin.
- Added `focus` to the arrow of the application layout.
- Added `focus` to the sortable heading in resultset table.
- Change in `Heading` `weight` tests.

### DIFF
--- a/lib/src/accordion/Accordion.stories.tsx
+++ b/lib/src/accordion/Accordion.stories.tsx
@@ -301,6 +301,7 @@ export const Chromatic = () => (
           lobortis eget.
         </div>
       </DxcAccordion>
+      <hr />
     </ExampleContainer>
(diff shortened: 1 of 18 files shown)


aweell:
As a suggestion, in the `heading` stories when comparing weights, we should maintain the same heading level since it's easier to compare them with an equal `font-size`.

Screen Shot 2022-03-14 at 5 49 52 PM
`

aweell (approved):
👍🏼

PR #765, opened 2022-03-16
Radio Group updates

GomezIvann:
Summary:

- Focus improvements. Now ignores disabled options.
- Radio Input label fix. The first click on a radio label of a radio group, without previously having the focus (the focus enters for the first time in the radio group), was not focusing its radio.
- Added new prop `readonly`. This includes both styling and functionality.
- Now Enter/Space keys select an option if it is focused but not checked.
- New tokens added.
- New aria's added.


marcialfps (approved):
🔥 🔥

PR #774, opened 2022-03-17
Changes in Stack, Row and List

raquelarrojo:
- Stack, Row and List types moved to an external file.
- Change in Stack, Row and List stories.

### DIFF
--- a/lib/src/bleed/Bleed.tsx
+++ b/lib/src/bleed/Bleed.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import styled from "styled-components";
-import BleedProps from "./types";
+import BleedPropsType from "./types";
 
 export default function Bleed({
   space,
@@ -11,7 +11,7 @@ export default function Bleed({
   bottom,
   left,
   children,
-}: BleedProps): JSX.Element {
+}: BleedPropsType): JSX.Element {
   return (
     <StyledBleed
       space={space}
@@ -54,8 +54,8 @@ function getSpacingValue(spacingName) {
   }
 }
 
-const StyledBleed = styled.div<BleedProps>`
-  ${({ space, horizontal, vertical, top, right, bottom, left }) => `
+const StyledBleed = styled.div<BleedPropsType>`
+  ${({ space, horizontal, vertical, top, right, bottom, left }: BleedPropsType) => `
     margin: -${getSpacingValue(top || vertical || space)} -${getSpacingValue(
     right || horizontal || space
   )} -${getSpacingValue(bottom || vertical || space)} -${getSpacingValue(left || horizontal || space)};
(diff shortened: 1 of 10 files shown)


lib/src/bleed/Bleed.tsx
```diff
@@ -54,8 +54,8 @@ function getSpacingValue(spacingName) {
   }
 }
 
-const StyledBleed = styled.div<BleedProps>`
-  ${({ space, horizontal, vertical, top, right, bottom, left }) => `
+const StyledBleed = styled.div<BleedPropsType>`
+  ${({ space, horizontal, vertical, top, right, bottom, left }: BleedPropsType) => `
```
marcialfps, line 58: It is not necessary to indicate `BleedPropsType` when passing to styled components.

lib/src/row/Row.tsx
```diff
@@ -19,9 +18,9 @@ export default function Row({ gutter = "none", align, justify, wrap = true, reve
 }
 const StyledRow = styled.div`
   display: flex;
-  flex-direction: ${({ reverse }: RowProps) => (reverse ? "row-reverse" : "row")};
-  flex-wrap: ${({ wrap }: RowProps) => (wrap ? "wrap" : "nowrap")};
-  align-items: ${({ align }: RowProps) => {
+  flex-direction: ${({ reverse }: RowPropsType) => (reverse ? "row-reverse" : "row")};
+  flex-wrap: ${({ wrap }: RowPropsType) => (wrap ? "wrap" : "nowrap")};
+  align-items: ${({ align }: RowPropsType) => {
```
marcialfps, line 23: It is not necessary to indicate `RowPropsType` in all cases if you indicate it to styled components.

lib/src/stack/Stack.tsx
```diff
@@ -32,7 +25,7 @@ const Divider = styled.div`
 const StyledStack = styled.div`
   display: flex;
   flex-direction: column;
-  align-items: ${({ align }: StackProps) => {
+  align-items: ${({ align }: StackPropsType) => {
```
marcialfps, line 28: Same.

marcialfps (changes requested):

lib/src/bleed/Bleed.tsx
raquelarrojo: Getting this error if I only indicate `BleedPropsType` in the styled component:

image

The same for `Row` and `Stack`.

marcialfps: I don't have that error. 
image

We are using the same approach in the radio group and is working fine...

raquelarrojo: Done!

lib/src/stack/Stack.tsx
raquelarrojo: Done!

lib/src/row/Row.tsx
raquelarrojo: Done!

marcialfps (approved):
💯

PR #776, opened 2022-03-18
[Patch] Fixed responsive behavior in footer component

Lisa159:


lib/src/footer/Footer.tsx
```diff
@@ -150,11 +143,10 @@ const BottomLinks = styled.div`
   & > span:last-child span {
     display: none;
   }
-  margin: ${(props) => (props.refSize <= responsiveSizes.mobileLarge ? "40px 0 40px 0" : "")};
```
marcialfps, line 153: Why do you deleted this margin?

lib/src/footer/Footer.tsx
```diff
@@ -150,11 +143,10 @@ const BottomLinks = styled.div`
...
+  align-self: center;
 `;
 
 const ChildComponents = styled.div`
-  min-height: 15px;
```
marcialfps, line 157: `min-height` is 16px according to the guidelines.

aweell: 47e160bdd040410030944251f12c5d53

lib/src/footer/Footer.tsx
Lisa159: Because it is adding a lot of space:
image

Without margin: 
image

lib/src/footer/Footer.tsx
Lisa159: This min-height is not doing anything. However, it is applying a margin-top of 16px (line 134).

Lisa159: Done

marcialfps (approved):

PR #779, opened 2022-03-18
[Minor] Remove iconSrc from Tag

marcialfps:
### Changes
- Remove iconSrc in Tag component.
- Update doc and examples.
- Fix stories and test.

### DIFF
--- a/lib/src/tag/types.ts
+++ b/lib/src/tag/types.ts
@@ -5,17 +5,13 @@ type Margin = {
   left?: Space;
   right?: Space;
 };
-type SVG = React.SVGProps<SVGSVGElement> | React.FunctionComponent<React.SVGProps<SVGSVGElement>>;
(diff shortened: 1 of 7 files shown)


lib/src/tag/types.ts
```diff
@@ -5,17 +5,13 @@ type Margin = {
...
   /**
-   * Element used as the icon that will be placed next to the label.
+   * Element or path used as the icon that will be placed next to the label.
    */
-  icon?: SVG;
-  /**
-   * @deprecated URL of the icon.
-   */
-  iconSrc?: string;
+  icon?: string | SVG;
```
GomezIvann, line 14: Does a Tab make sense without an `icon`? The same question goes for the `label`.

GomezIvann (commented):
I think we should add an storie with and url passed to the `icon` prop, just to test if it resizes and it is positioned correctly.

PR #780, opened 2022-03-18
Radio Group update

GomezIvann:


aweell (commented):
Remove the `cursor: pointer` property when the state of the radio group is `readonly` in order to distinguish the default interaction and this state.

marcialfps (approved):

PR #782, opened 2022-03-21
[Patch] Fixed focus sort in resultset table

Lisa159:


aweell (commented):
Does it seem that maybe another outline is being applied?

Screen Shot 2022-03-21 at 9 49 15 AM

It actually works well visually since the blue and the header purple blend a little. I would leave like that just knowing where these styles are coming from (or even applying the consciously in our component implementation) if we need to change them in the future...

Lisa159:
> Does it seem that maybe another outline is being applied?
> 
> Screen Shot 2022-03-21 at 9 49 15 AM
> 
> It actually works well visually since the blue and the header purple blend a little. I would leave like that just knowing where these styles are coming from (or even applying the consciously in our component implementation) if we need to change them in the future...

Looks like it is applying the white border when outline style is auto or revert. If we change it to another value, it will not have rounded corners. Maybe changing outline to box-shadow?

aweell:
> > Does it seem that maybe another outline is being applied?
> > Screen Shot 2022-03-21 at 9 49 15 AM
> > It actually works well visually since the blue and the header purple blend a little. I would leave like that just knowing where these styles are coming from (or even applying the consciously in our component implementation) if we need to change them in the future...
> 
> Looks like it is applying the white border when outline style is auto or revert. If we change it to another value, it will not have rounded corners. Maybe changing outline to box-shadow?

Use `solid` instead

lib/src/resultsetTable/ResultsetTable.tsx
```diff
@@ -197,9 +198,10 @@ const HeaderContainer = styled.div`
       : props.theme.headerTextAlign === "right"
       ? "flex-end"
       : "flex-start"};
-  width: 100%;
+  width: fit-content;
   :focus {
-    outline: #0095ff auto 1px;
+    outline: ${(props) => (props.isSortable ? "#0095ff solid 2px" : "")};
```
marcialfps, line 203: I think this is better: `${(props) => (props.isSortable && "outline: #0095ff solid 2px;")}`

lib/src/resultsetTable/ResultsetTable.tsx
```diff
@@ -200,8 +200,8 @@ const HeaderContainer = styled.div`
       : "flex-start"};
   width: fit-content;
   :focus {
-    outline: ${(props) => (props.isSortable ? "#0095ff solid 2px" : "")};
-    outline-offset: 4px;
+    ${(props) => props.isSortable && "outline: #0095ff solid 2px;"}
+    ${(props) => props.isSortable && "outline-offset: 4px;"}
```
marcialfps, line 204: Perhaps unify the condition?

```
${(props) => props.isSortable && 
      "outline: #0095ff solid 2px;
       outline-offset: 4px;"
}
```

Lisa159: Done.

marcialfps (approved):
💯

Issue #781, opened 2022-03-18
[Accordion-group] Box shadow in disabled state


aweell:
## Issue description

The current disabled state of the accordion-group component is displaying the `box-shadow` of each single accordion element.

<img width="1322" alt="Screen Shot 2022-03-18 at 2 47 07 PM" src="https://user-images.githubusercontent.com/44420072/159014662-f0bd325c-cda4-4892-85ec-3dcb43308d00.png">

_Fig. 1 - Enabled state_

<img width="1314" alt="Screen Shot 2022-03-18 at 2 47 00 PM" src="https://user-images.githubusercontent.com/44420072/159014657-e264d9f5-0ea5-4380-93dd-2992aafc59cd.png">

_Fig. 2 - Disabled state_

The box-shadow should stay at bottom as shown in fig. 3:

<img width="870" alt="Screen Shot 2022-03-18 at 2 52 07 PM" src="https://user-images.githubusercontent.com/44420072/159015268-75ad0d3e-abcb-4bf3-85f1-62f585ebf1e8.png">

_Fig. 3 - Disabled state with box-shadow fix_

PR #788, opened 2022-03-21
[Patch] Fix box shadow in disabled accordion group

raquelarrojo:
Removed `background-color: transparent` in disabled accordion. Also removed unnecessary `disabledBackgroundColor` accordion's token.

Now in a disabled accordion group, the box shadow is only at the bottom:

image

Closes [781](https://github.com/dxc-technology/halstack-react/issues/781).


aweell (commented):
😎

aweell (approved):

PR #797, opened 2022-03-23
Types for Badge inner component

### LINKED ISSUE #768: [Badge] Types remaining
Badge inner component has not been updated to typescript as the other components.

marcialfps:
Closes #768


lib/src/badge/types.ts
```diff
@@ -0,0 +1,5 @@
+type Props = {
+  notificationText?: boolean | number;
```
raquelarrojo, line 2: I think it should be also `string` type, as in the tabs (where the badge is being used), if the notificationNumber is > 99, `+99` will be put as notificationText in the badge.

raquelarrojo (changes requested):

raquelarrojo (approved):

PR #801, opened 2022-03-24
Radio Group update

GomezIvann:
Summary:

- New prop `tabIndex` added.
- More jest tests added.
- More stories added.
- Code improvements and some hotfixes that were found during testing.


marcialfps (approved):

aweell (commented):
Why the `:focus` state in chromatic is displayed as:

Screen Shot 2022-03-28 at 11 21 15 AM

?

In storybook seems to be correct
Screen Shot 2022-03-28 at 11 21 55 AM

aweell (commented):
A couple of accessibility questions:

Should the disabled and read-only statuses have an `aria-required="true"`?

Shouldn't the read-only radio group have an `aria-readonly="true"` included in the `role="radiogroup"` container?

GomezIvann:
> A couple of accessibility questions:
> 
> Should the disabled and read-only statuses have an `aria-required="true"`?


First of all, thanks for the feedback! Answering your questions:

> Shouldn't the read-only radio group have an `aria-readonly="true"` included in the `role="radiogroup"` container?
Yes, the `aria-readonly` is missing at the Radio Group level.

> Should the disabled and read-only statuses have an `aria-required="true"`?
I refactor this question, is there any scenario where a disabled radio group, that is also required, makes sense?

GomezIvann:
> Why the `:focus` state in chromatic is displayed as:
> 
> Screen Shot 2022-03-28 at 11 21 15 AM
> 
> ?
> 
> In storybook seems to be correct Screen Shot 2022-03-28 at 11 21 55 AM

Seems to be a persistent problem with Chromatic. This is also happening in the Link component, maybe a problem with `border-radius`, I'm not sure...

aweell:
> Seems to be a persistent problem with Chromatic. This is also happening in the Link component, maybe a problem with `border-radius`, I'm not sure...

Okay, since it has no impact on the component behavior we can move on and review this chromatic issue separately 🙈 . I accepted the new changes as baseline.

marcialfps (approved):

PR #804, opened 2022-03-25
Fix #777: New aria attributes for error and disabled states

### LINKED ISSUE #777: [A11y] Update aria labels in input, select and text area
## Issue description

The current implementation doesn't use `aria-errormessage` and `aria-live: "assertive"` to indicate assistive technologies that an error has occurred in the text input, textarea and select component.

It's needed also to indicate using `aria-disabled`, the disabled states of the previously mentioned components.

### Related tasks

- [x] [A11y] Text-input - Update aria labels
- [x] [A11y] Textarea - Update aria labels
- [x] [A11y] Select - Update aria labels

### References

* [MDN Docs - `aria-errormessage`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage)
* [MDN Docs - `aria-live` regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)
* [MDN Docs - `aria-disabled`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled)

GomezIvann:
Fixes #777. Summary (Select, Textarea and TextInput components):

- New aria attributes `aria-disabled` for disabled state.
- New aria attributes `aria-errormessage` and `aria-live` for error state.
- Tests for checking correct aria labelling.

### DIFF
--- a/lib/src/select/Select.tsx
+++ b/lib/src/select/Select.tsx
@@ -188,6 +188,7 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
   ): JSX.Element => {
     const [selectId] = useState(`select-${uuidv4()}`);
     const selectLabelId = `label-${selectId}`;
+    const errorId = `error-${selectId}`;
     const optionsListId = `${selectId}-listbox`;
     const [innerValue, setInnerValue] = useState(multiple ? [] : "");
     const [searchValue, setSearchValue] = useState("");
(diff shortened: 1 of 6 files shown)


aweell (approved):
This is lit! 🔥

PR #808, opened 2022-03-28
Select code and performance improvements

GomezIvann:
Summary:

- Listbox options have been extracted to an inner component called `Option`. This cleans a little bit of code from the Select and enhances its performance, since the options are now memorized by React to avoid unnecessary re-renders.
- Icons are now in a separate file.
- New `aria-orientation` attribute for the Listbox helps better understand the component's anatomy.
- `aria-required` is now set to `false` when the component is disabled.


lib/src/select/Select.tsx
```diff
@@ -360,6 +281,12 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
       setSearchValue("");
     };
 
+    const handleOptionOnClick = useCallback(() => {
+      handleSelectChangeValue(optionalEmptyOption);
+      !multiple && closeOptions();
+      setSearchValue("");
+    }, [handleSelectChangeValue, closeOptions, multiple]);
+
```
raquelarrojo, line 289: What if we make this function generic to avoid repetition of code?
We can pass the option as an argument like:

```
const handleOptionOnClick = useCallback((option) => {
      handleSelectChangeValue(option);
      !multiple && closeOptions();
      setSearchValue("");
    }, [handleSelectChangeValue, closeOptions, multiple]);
```

lib/src/select/Select.tsx
```diff
@@ -432,15 +314,52 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
...
+                  return (
+                    <Option
+                      id={`option-${globalIndex}`}
+                      option={singleOption}
+                      onClick={() => {
+                        handleSelectChangeValue(singleOption);
+                        !multiple && closeOptions();
+                        setSearchValue("");
+                      }}
+                      multiple={multiple}
```
raquelarrojo, line 326: Following the comment above, we can call `handleOptionOnClick(singleOption)` here.

lib/src/select/Select.tsx
```diff
@@ -432,15 +314,52 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
...
+          <Option
+            key={`option-${option.value}`}
+            id={`option-${globalIndex}`}
+            option={option}
+            onClick={() => {
+              handleSelectChangeValue(option);
+              !multiple && closeOptions();
+              setSearchValue("");
+            }}
+            multiple={multiple}
```
raquelarrojo, line 354: Following the comment above, we can call `handleOptionOnClick(option)` here.

lib/src/select/Select.tsx
```diff
@@ -562,14 +481,31 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
...
                   </OptionsSystemMessage>
                 ) : (
-                  optional && !multiple && <Option option={optionalEmptyOption} index={0} />
+                  optional &&
+                  !multiple && (
+                    <Option
+                      id={`option-${0}`}
+                      option={optionalEmptyOption}
+                      onClick={handleOptionOnClick}
+                      multiple={multiple}
```
raquelarrojo, line 498: Following the comment above, we can call `handleOptionOnClick(optionalEmptyOption)` here.

lib/src/select/Select.tsx
GomezIvann: Yes! I started with the strategy you mention in the comments (a common function surrounded by a `useCallback` to not cause unnecessary re-renders) but, for some reason, i discontinued it.

raquelarrojo (approved):

PR #809, opened 2022-03-28
Fix in theme documentation and ThemeContext

raquelarrojo:
- Fix in `Custom Themes` documentation.
- Added `dateInput` and `select` to DefaultTheme.json.
- Added `textInput` and `select` to Default.schema.json.
- Added `select` to ThemeContext.
- Change `selectedBaseColor` to `selectedFontColor` in toggle's `selectedDisabledFontColor` token in ThemeContext.


docs/src/pages/overview/Overview.jsx
```diff
@@ -109,8 +109,8 @@ function CustomThemes() {
         customize, using the next structure:
         <SyntaxHighlighter language="javascript" style={docco}>
           {`componentName: {
-  themeInput: "value", 
-}`}
+            themeInput: "value", 
+          }`}
```
marcialfps, line 113: Remove tabs.
image

docs/src/pages/overview/Overview.jsx
```diff
@@ -125,125 +125,123 @@ function CustomThemes() {
 
       <SyntaxHighlighter language="javascript" style={docco}>
         {`
-  import { DxcButton, ThemeProvider } from "@dxc-technology/halstack-react";
+          import { DxcButton, ThemeProvider } from "@dxc-technology/halstack-react";
```
marcialfps, line 128: I think we should remove the first tab, looks very alienated on the right.

marcialfps (changes requested):

docs/src/pages/overview/Overview.jsx
raquelarrojo: Done!

docs/src/pages/overview/Overview.jsx
raquelarrojo: Done!

marcialfps (approved):
🔥 🔥 🔥

PR #870, opened 2022-04-04
New prop 'defaultValue' for Date Input

GomezIvann:
New prop `defaultValue` added for the component Date Input. In this case, it benefits from the already existing prop of the Text Input.
Also, the examples from the Date Input have been updated to match the other form components.


lib/src/date-input/DateInput.tsx
```diff
@@ -11,30 +11,32 @@ import useTheme from "../useTheme";
...
       onChange,
       onBlur,
       error,
-      autocomplete = "off",
+      autocomplete,
       margin,
-      size = "medium",
-      tabIndex = 0,
+      size,
+      tabIndex,
```
marcialfps, line 35: Why have you removed these default values?

docs/src/pages/components/cdk-components/date-input/examples/basicDateInput.js
```diff
@@ -0,0 +1,40 @@
...
+const code = `() => {
+  const options = [
+    { label: "Option 01", value: "1" },
+    { label: "Option 02", value: "2" },
+    { label: "Option 03", value: "3" },
+    { label: "Option 04", value: "4" },
+  ];
+
+  return (
+    <div style={{ display: "flex" }}>
```
marcialfps, line 13: Maybe is a good time to start using our layout components in these examples.

marcialfps (changes requested):

lib/src/date-input/DateInput.tsx
GomezIvann: The Date Input uses internally the Text Input, so the default values for each prop are already there, we were rewriting the same code.

GomezIvann: I only did this to the props that are passed directly to the text input and already have a default value (and match the value).

docs/src/pages/components/cdk-components/date-input/examples/basicDateInput.js
GomezIvann: We can take this into account when we change the Example component, what do you think?

GomezIvann: Since we are going to change the way the examples are presented we can do this in the future.

marcialfps (approved):

PR #871, opened 2022-04-04
Stack component documentation

marcialfps:
Documentation and examples for Stack component.


docs/src/pages/components/cdk-components/stack/api.jsx
```diff
@@ -0,0 +1,50 @@
...
+        <td>children: node</td>
+        <td></td>
+        <td>Custom content inside the stack.</td>
+      </tr>
+      <tr>
+        <td>divider: boolean</td>
+        <td>
+          <code>false</code>
+        </td>
+        <td>If true a divider is shown between children.</td>
```
GomezIvann, line 34: I think we need to add a comma for better reading: "If true, a divider is shown between children".

lib/src/stack/types.tsx
```diff
@@ -10,9 +13,21 @@ type Props = {
...
+  /**
+   * If true a divider is shown between children.
+   */
   divider?: boolean;
+  /**
+   * Alignment applied to children.
+   */
   align?: "start" | "center" | "end" | "baseline" | "stretch";
+  /**
+   * Component type or primitive that is rendered as the type of the root element.
```
GomezIvann, line 25: If you made any change based on the feedback, remember to repeat it here 🚀.

marcialfps: Done!! ✌🏼

PR #872, opened 2022-04-04
New prop 'defaultValue' for Number Input 

GomezIvann:
New prop `defaultValue` added to the Number Input. In this case, it benefits from the already existing prop of the Text Input.
Also, the examples from the Number Input have been updated to match the other form components.

### DIFF
--- a/docs/src/pages/components/cdk-components/bleed/Bleed.jsx
+++ b/docs/src/pages/components/cdk-components/bleed/Bleed.jsx
@@ -1,0 +1,32 @@
+import React from "react";
+import { DxcHeading } from "@dxc-technology/halstack-react";
+import ComponentDoc from "../../common/ComponentDoc";
+import Section from "../../common/Section";
+import Example from "../../common/Example";
+import ComponentHeader from "../../common/ComponentHeader";
+import BleedPropsTable from "./api.jsx";
+import space from "./examples/space";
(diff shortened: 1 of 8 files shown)


docs/src/pages/components/cdk-components/number-input/examples/customErrorsNumberInput.js
```diff
@@ -4,29 +4,30 @@ import { useState } from "react";
 const code = `() => {
   const [value, setValue] = useState("");
   const [errorMessage, setErrorMessage] = useState("");
-  const onChange = ({ value }) => {
+  const onChange = ({ value, error }) => {
```
marcialfps, line 7: I find this example a bit difficult to understand, perhaps it is better to separate the two cases.

GomezIvann: I agree, but this is example is in every other input and textarea component. Since we have to review the error prop and how we handle errors in the documentation (empty string issue), we can discuss this there. But. for now, I think I prefer consistency rather than a better example...

marcialfps: I agree. 👍🏼

marcialfps (approved):

PR #881, opened 2022-04-06
Row component documentation

marcialfps:
Documentation and examples for Row component.


docs/src/pages/components/cdk-components/row/api.jsx
```diff
@@ -0,0 +1,58 @@
...
+          <code>false</code>
+        </td>
+        <td>If true, children are shown in reverse order.</td>
+      </tr>
+      <tr>
+        <td>wrap: boolean</td>
+        <td>
+          <code>true</code>
+        </td>
+        <td>If true, children will wrap onto multiple rows.</td>
```
GomezIvann, line 52: Do you find it interesting to explain what happens when it is false or is not necessary? Something like: "If false, children will overflow when they reach the container limits."

marcialfps: It is a good idea to explain it. Thank you!

GomezIvann (approved):

PR #888, opened 2022-04-07
[Major] onChange and onBlur undefined error parameter in Textarea

### LINKED ISSUE #887: Change `onChange` and `onBlur` event in Textarea

marcialfps:
Closes #887


GomezIvann (changes requested):
Some things missing:

- Update the description of the prop `error`, I recommend changing it to this: _"If it is a defined value and also a truthy string, the component will change its appearance, showing the error below the radio group. If the defined value is an empty string, it will reserve a space below the component for a future error, but it would not change its look. In case of being undefined or null, both the appearance and the space for the error message would not be modified."_. This is the existing description of the Radio Group component, which I think helps to better understand how the error now works with empty strings and undefined/null values.
- Update the controlled examples which can display errors, to avoid unexpected jumps in transitions between error states and non-error states. They are three:
  - **Controlled textarea.** Since the are specific examples for the error, I suggest removing it from here.
  - **Textarea with pattern constraint.**
  - **Textarea with length constraint.**
  - **Textarea with custom error messages.**
The easiest way is to do it like the example _"Handling errors"_ from the Radio Group. It would be like this (Pattern example):

image

marcialfps:
>     The easiest way is to do it like the example _"Handling errors"_ from the Radio Group. It would be like this (Pattern example):
> 
> image

What do you think of this way? I think it is easier:

image

GomezIvann:
> > ```
> > The easiest way is to do it like the example _"Handling errors"_ from the Radio Group. It would be like this (Pattern example):
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > image
> 
> What do you think of this way? I think it is easier:
> 
> image


It was done like that to show the user how to manage the `undefined`/`null` value in only one place. Imagine if the use case is to treat pattern error at the `onChange` and the length error at the `onBlur`. If we do it like you've done the undefined/null value should be managed in each method, while with the other approach, it's only a few more characters when you pass the prop!

For further knowledge, I didn't use the nullish coalescing operator (??) since our code example environment doesn't support the latest version of javascript.

GomezIvann (approved):

PR #890, opened 2022-04-07
[Major] onChange and onBlur undefined error parameter in Text, Number and Password

### LINKED ISSUE #889: Change `onChange` and `onBlur` event in Text Input

### LINKED ISSUE #896: Change `onChange` and `onBlur` event in Number Input

### LINKED ISSUE #897: Change `onChange` and `onBlur` event in Password Input

marcialfps:
- onChange and onBlur undefined error parameter in Number, Password and Text input.
- Improved description of `error` prop.
- Avoid jumps and improvements in examples.
- Fix Textarea error prop description in types file.

Closes #889, closes #896, closes #897


docs/src/pages/components/cdk-components/number-input/examples/minMaxStepNumberInput.js
```diff
@@ -3,7 +3,7 @@ import { useState } from "react";
 
 const code = `() => {
   const [value, setValue] = useState("");
-  const [error, setError] = useState("");
```
GomezIvann, line 6: In this particular example, you don't need to remove the initial value `""` of the `error` variable, since we are using the predefined error message. It also adds some value, since if they are using typescript it will infer the type of the stateful variable (`string` in this case, not `undefined` as it currently suggests).

docs/src/pages/components/cdk-components/password-input/api.jsx
```diff
@@ -64,17 +64,19 @@ const PasswordInputPropsTable = () => {
           entered is not valid) will be passed to this function. An example of
           this object is: {"{ "}
           <code>value: value, error: error</code>
-          {" }"}. If there is no error, error will be null.
+          {" }"}. If there is no error, error will not be defined.
         </td>
       </tr>
       <tr>
         <td>error: string</td>
```
GomezIvann, line 71: `pattern`, `minLength` and `maxLength` prop descriptions need to be updated too (_"... the error parameter of both events will be null"_).

docs/src/pages/components/cdk-components/password-input/examples/patternPasswordInput.js
```diff
@@ -3,7 +3,7 @@ import { useState } from "react";
 
 const code = `() => {
   const [value, setValue] = useState("");
-  const [error, setError] = useState("");
```
GomezIvann, line 6: In this particular example, you don't need to remove the initial value `""` of the `error` variable, since we are using the predefined error message. It also adds some value, since if they are using typescript it will infer the type of the stateful variable (`string` in this case, not `undefined` as it currently suggests).

docs/src/pages/components/cdk-components/password-input/examples/lengthPasswordInput.js
```diff
@@ -3,7 +3,7 @@ import { useState } from "react";
 
 const code = `() => {
   const [value, setValue] = useState("");
-  const [error, setError] = useState("");
```
GomezIvann, line 6: In this particular example, you don't need to remove the initial value "" of the error variable, since we are using the predefined error message. It also adds some value, since if they are using typescript it will infer the type of the stateful variable (string in this case, not undefined as it currently suggests).

docs/src/pages/components/cdk-components/text-input/api.jsx
```diff
@@ -119,16 +119,19 @@ const textInputPropsTable = () => {
           entered is not valid) will be passed to this function. An example of
           this object is: {"{ "}
           <code>value: value, error: error</code>
-          {" }"}. If there is no error, error will be null.
+          {" }"}. If there is no error, error will not be defined.
         </td>
       </tr>
       <tr>
         <td>error: string</td>
```
GomezIvann, line 126: `pattern`, `minLength` and `maxLength` prop descriptions need to be updated too (_"... the error parameter of both events will be null"_).

docs/src/pages/components/cdk-components/text-input/examples/lengthTextInput.js
```diff
@@ -3,7 +3,7 @@ import { useState } from "react";
 
 const code = `() => {
   const [value, setValue] = useState("");
-  const [error, setError] = useState("");
```
GomezIvann, line 6: In this particular example, you don't need to remove the initial value `""` of the `error` variable, since we are using the predefined error message. It also adds some value, since if they are using typescript it will infer the type of the stateful variable (`string` in this case, not `undefined` as it currently suggests).

docs/src/pages/components/cdk-components/text-input/examples/patternTextInput.js
```diff
@@ -3,7 +3,7 @@ import { useState } from "react";
 
 const code = `() => {
   const [value, setValue] = useState("");
-  const [error, setError] = useState("");
```
GomezIvann, line 6: In this particular example, you don't need to remove the initial value `""` of the `error` variable, since we are using the predefined error message. It also adds some value, since if they are using typescript it will infer the type of the stateful variable (`string` in this case, not `undefined` as it currently suggests).

lib/src/password-input/types.ts
```diff
@@ -31,19 +31,22 @@ type Props = {
...
-   * the error below the password input component. If it is not defined, the
-   * error messages will be managed internally, but never displayed on its own.
+   * If it is a defined value and also a truthy string, the component will
+   * change its appearance, showing the error below the password input
+   * component. If the defined value is an empty string, it will reserve a
+   * space below the component for a future error, but it would not change
+   * its look. In case of being undefined or null, both the appearance and
+   * the space for the error message would not be modified.
    */
   error?: string;
```
GomezIvann, line 51: Remember to also update here `pattern`, `minLength` and `maxLength` prop type descriptions.

lib/src/text-input/types.ts
```diff
@@ -78,20 +78,23 @@ type Props = {
...
-   * the error below the input component. If it is not defined, the error
-   * messages will be managed internally, but never displayed on its own.
+   * If it is a defined value and also a truthy string, the component will
+   * change its appearance, showing the error below the input component. If
+   * the defined value is an empty string, it will reserve a space below
+   * the component for a future error, but it would not change its look. In
+   * case of being undefined or null, both the appearance and the space for
+   * the error message would not be modified.
    */
   error?: string;
```
GomezIvann, line 99: Remember to also update here `pattern`, `minLength` and `maxLength` prop type descriptions.

lib/src/textarea/types.ts
```diff
@@ -68,9 +68,12 @@ type Props = {
...
-   * the error below the textarea. If it is not defined, the error
-   * messages will be managed internally, but never displayed on its own.
+   * If it is a defined value and also a truthy string, the component will 
+   * change its appearance, showing the error below the textarea. If the 
+   * defined value is an empty string, it will reserve a space below the
+   * component for a future error, but it would not change its look. In 
+   * case of being undefined or null, both the appearance and the space for 
+   * the error message would not be modified.
    */
   error?: string;
```
GomezIvann, line 78: Remember to also update here `pattern`, `minLength` and `maxLength` prop type descriptions. In the case of the Textarea, we also need to update the documentation of the component... 😢

lib/src/number-input/types.ts
```diff
@@ -74,20 +74,23 @@ type Props = {
...
-   * the error below the input component. If it is not defined, the error
-   * messages will be managed internally, but never displayed on its own.
+   * If it is a defined value and also a truthy string, the component will
+   * change its appearance, showing the error below the input component. If
+   * the defined value is an empty string, it will reserve a space below
+   * the component for a future error, but it would not change its look. In
+   * case of being undefined or null, both the appearance and the space for
+   * the error message would not be modified.
    */
   error?: string;
```
GomezIvann, line 95: Remember to also update here `min` and `max` prop type descriptions.

GomezIvann (changes requested):

docs/src/pages/components/cdk-components/number-input/examples/minMaxStepNumberInput.js
marcialfps: I'm not quite sure I understand your comment... But in the case of the type, you should note that the `error` variable will also be set to undefined when there is no error, so I think it will be the same to set initially to empty string or undefined.

GomezIvann: When there's no error, our current code will set it to `""` (`error={error == undefined ? "" : error}`), that's why I suggest setting it to the empty string by default, since it would never be `undefined` again. This only applies to this kind of example, with the custom error message examples we should keep it as it stands.

GomezIvann (approved):

PR #898, opened 2022-04-08
New prop `defaultIsExpanded` for Accordion

GomezIvann:
New prop `defaultIsExpanded` added to the Accordion component. It lets the user change the initial state of the accordion's panel when it is uncontrolled. The prop has been documented and tested.


lib/src/accordion/Accordion.test.js
```diff
@@ -5,23 +5,24 @@ import DxcAccordion from "./Accordion";
...
-
     fireEvent.click(getByText("Accordion"));
     expect(onChange).toHaveBeenCalled();
   });
-
   test("Controlled accordion", () => {
     const onChange = jest.fn();
     const { getByText } = render(<DxcAccordion label="Accordion" onChange={onChange} isExpanded={true} />);
-
     fireEvent.click(getByText("Accordion"));
```
marcialfps, line 26: We should add `expect(getByRole("button").getAttribute("aria-expanded")).toBe("true");` to this test.

GomezIvann: Agree!

marcialfps (approved):

marcialfps (approved):

PR #904, opened 2022-04-11
New prop `defaultValue` for Slider

GomezIvann:
New prop `defaultValue` added to the Slider component. It lets the user change the initial value of the component when it is uncontrolled. The prop has been documented and tested.
Also, some unnecessary examples of the component have been removed from its documentation.


docs/src/pages/components/cdk-components/slider/examples/uncontrolled.js
```diff
@@ -13,6 +13,7 @@ const code = `() => {
     <DxcSlider
       label="Select a value"
       helperText="Helper text"
+      defaultvalue={25}
```
marcialfps, line 16: Not working, change to `defaultValue`.

marcialfps (approved):
💯

PR #914, opened 2022-04-12
New prop `defaultActiveTabIndex` for Tabs

GomezIvann:
New prop `defaultActiveTabIndex` added. It has been documented and tested.


marcialfps (approved):
💯

PR #921, opened 2022-04-19
Fix Button `z-index` and Select `listbox` refactor

GomezIvann:
Summary:

1. Fix issue regarding Button's label when a component with defined z-index was over the mentioned component.

_Previous behaviour:_
image
_After fix:_
image

2. **Select component refactor.** Listbox element extracted to an inner component for performance and code improvements.
3. Now each element of the listbox has a single key to avoid React warnings.

### DIFF
--- a/lib/src/select/Listbox.tsx
+++ b/lib/src/select/Listbox.tsx
@@ -25,7 +25,56 @@ const Listbox = React.forwardRef<ListboxRefType, ListboxProps>(
     ref
   ): JSX.Element => {
     const colorsTheme = useTheme();
+
     let globalIndex = optional && !multiple ? 0 : -1; // index for options, starting from 0 to options.length -1
+    const mapOptionFunc = (option, mapIndex) => {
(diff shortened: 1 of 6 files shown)


lib/src/select/Listbox.tsx
```diff
@@ -0,0 +1,169 @@
...
+                  option={option}
+                  onClick={handleOptionOnClick}
+                  multiple={multiple}
+                  visualFocused={visualFocusIndex === globalIndex}
+                  isLastOption={lastOptionIndex === globalIndex}
+                  isSelected={multiple ? currentValue.includes(option.value) : currentValue === option.value}
+                />
+              );
+            }
+          })}
```
marcialfps, line 112: I think is better to have this part in a separate function called `mapOptionFunc` as before.

GomezIvann: Done!

marcialfps (approved):

PR #922, opened 2022-04-19
[Patch] Prevent mouse wheel from modifying number input value

### LINKED ISSUE #460: [Number Input] Mouse wheel modifies the value of the component
Number Input component modifies its value when the cursor is over it and the mouse wheel scrolls. This is inherited from the native `<input type="number" />` and Reacts own implementation of the `wheel` event. The solution would be to remove the default behaviour in future iterations of the component.

marcialfps:
Closes #460


marcialfps:
This functionality should be discussed.

PR #927, opened 2022-04-21
[Patch] Wizard margin overflow

### LINKED ISSUE #924: [Wizard] Margin overflow
Adding a value to the prop `margin` of the component Wizard makes it overflow its container:

image

This is being caused by a css rule `width: 100%` in the main container of the component.

marcialfps:
Closes #924

### DIFF
--- a/lib/src/wizard/Wizard.tsx
+++ b/lib/src/wizard/Wizard.tsx
@@ -86,7 +86,7 @@ const DxcWizard = ({
               aria-current={renderedCurrent === i ? "step" : "false"}
               tabIndex={tabIndex}
             >
-              <StepHeader>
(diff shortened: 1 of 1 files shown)


GomezIvann (commented):
While using the wizard I saw this: 

image

Seems that each step is reserving the space for the valid icon and that makes the focus a bit strange. The look in the documentation fits the step:

image

marcialfps:
> While using the wizard I saw this:
> 
> image
> 
> Seems that each step is reserving the space for the valid icon and that makes the focus a bit strange. The look in the documentation fits the step:
> 
> image

Done!

GomezIvann (approved):

PR #935, opened 2022-04-25
[Major] Replace required prop with optional in Switch

GomezIvann:
1. Removed `required` prop from Switch component.
2. Added `optional` prop to Switch component.

### DIFF
--- a/docs/src/pages/components/cdk-components/switch/api.jsx
+++ b/docs/src/pages/components/cdk-components/switch/api.jsx
@@ -64,9 +64,8 @@ const switchPropsTable = () => {
           <code>false</code>
         </td>
         <td>
-          If true, the component will be optional, showing{" "}
-          <code>(Optional)</code> next to the label. Otherwise, it should be
-          considered required and display an error if it is convenient.
+          If true, the component will display <code>(Optional)</code> next to
+          the label.
         </td>
       </tr>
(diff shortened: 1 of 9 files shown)


Jialecl (approved):

docs/src/pages/components/cdk-components/switch/api.jsx
```diff
@@ -59,13 +59,14 @@ const switchPropsTable = () => {
...
+        <td>optional: boolean</td>
         <td>
           <code>false</code>
         </td>
         <td>
-          If true, the switch will change its appearence, showing that the value
-          is required.
+          If true, the component will be optional, showing{" "}
+          <code>(Optional)</code> next to the label. Otherwise, it should be
+          considered required and display an error if it is convenient.
```
Jialecl, line 69: Maybe this is not correct since the component has no way of showing errors

GomezIvann: That's why I say "...if it is convenient." and "...it should be considered..." in the prop description. Maybe I need to rephrase it like this:

- If true, the component will be optional, showing `(Optional)` next to the label. Otherwise, it should be considered required and, if it is convenient, display an error message.

What do you think?

Jialecl (approved):

PR #953, opened 2022-05-06
[Patch] Fixed bug error of button without onClick prop

Lisa159:


lib/src/button/Button.tsx
```diff
@@ -47,9 +47,7 @@ const DxcButton = ({
           disableRipple
           aria-disabled={disabled}
           tabIndex={disabled ? -1 : tabIndex}
-          onClick={() => {
```
GomezIvann, line 50: We need to undo this change. It was been called like that to prevent react from sending the `event` parameter to the user's `onClick` function. If we write `onClick={onClick}` and the user sends us a function like `(event) => {console.log(event);}` it will work and print the value, and we don't want that.

image

lib/src/button/Button.tsx
```diff
@@ -47,9 +47,7 @@ const DxcButton = ({
...
-            onClick();
-          }}
+          onClick={onClick}
```
jsardid, line 50: These two things are not exactly the same. The old version doesn't send the event object to the handler, and the new one does.

I think we decided not to send anything to the onClick handler, do you remember this @GomezIvann??

GomezIvann (changes requested):

lib/src/button/Button.tsx
Lisa159: I see, I was trying to resolve the error showing when a button doesn't have onClick prop defined. The examples in the themeBuilder don't have it and locally it is crashing all the page.

GomezIvann: Yes, you are completely right, it was made on purpose. To solve this issue #953 we have to options: 

1. Give the prop a default value: `() => {}`.
2. Change the `onClick` event to mandatory, since I can't find out a scenario where it makes sense to not pass it 🤔 .

Lisa159: Changed onClick to mandatory as I am also agree with the second point.

GomezIvann: I also suggest to not mark the prop as mandatory, just to avoid changes in the api of the component.

Lisa159: Done as @GomezIvann suggested.

GomezIvann (approved):

PR #954, opened 2022-05-06
Link supports client-side navigation

### LINKED ISSUE #822: Review Link implementation for client side navigation

marcialfps:
### New features in DxcLink
- Removed `text` prop.
- Added `children` as prop. Can be a `string` or `ReactNode`.
- Updated Link tests and stories.
- Updated Link documentation and changed some examples.

### Other changes
- Updated `docs` and `app` with DxcLink changes.
- Use of React Router combined with DxcLink in `docs`.

Closes #822

### DIFF
--- a/docs/src/pages/components/cdk-components/applicationLayout/examples/default.js
+++ b/docs/src/pages/components/cdk-components/applicationLayout/examples/default.js
@@ -1,18 +1,17 @@
 const code = `() => {
   return (
-    <iframe
-      src="https://codesandbox.io/embed/without-sidenav-v7ou0?fontsize=14&hidenavigation=1&theme=dark"
-      style={{
-        width: "100%",
-        height: "500px",
-        border: "0",
-        borderRadius: "4px",
-        overflow: "hidden",
+    <iframe src="https://codesandbox.io/embed/without-sidenav-k7xjqf?fontsize=14&hidenavigation=1&theme=dark"
+     style={{
+       width:"100%",
+       height:"500px",
+       border:"0",
+       borderRadius: "4px",
+       overflow:"hidden",
       }}
-      title="Without sidenav"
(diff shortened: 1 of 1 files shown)


docs/src/pages/components/cdk-components/link/api.jsx
```diff
@@ -63,6 +58,16 @@ const linkPropsTable = () => {
           be called when the user clicks the link.
         </td>
       </tr>
+      <tr>
+        <td>children: string | node</td>
+        <td></td>
+        <td>
+          Content to display in the link. To use the link with other libraries
```
GomezIvann, line 65: I think this description is a bit verbose, maybe reformulate it like: "Content of the link. To use the component with other libraries (such as React Router or Next.js Link) pass the custom link as a child. The component will apply the styles to the custom link.", what do you think?

docs/src/pages/overview/Overview.jsx
```diff
@@ -62,8 +59,9 @@ function UseComponents() {
           href="https://developer.dxc.com/design/components"
```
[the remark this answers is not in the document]
marcialfps: Fixed. Changed to `/components`.

docs/src/pages/components/cdk-components/link/api.jsx
marcialfps: Yes, now is better 👍🏼

lib/src/link/Link.tsx
```diff
@@ -78,6 +78,7 @@ const LinkContainer = styled.div<LinkContainerProps>`
     padding: 0;
     cursor: pointer;
     font-family: inherit;
+    font-size: inherit;
```
GomezIvann, line 81: Shouldn't this rule have the token value? Same question goes for font-family, font-style, font-weight, etc.

GomezIvann: Currently they are at `LinkText` but they won't affect every version of the link, no? 🤔

marcialfps: Done! ✌🏼

GomezIvann (approved):

PR #960, opened 2022-05-11
Fix no previous or next page in Specification pages

marcialfps:
Previous or next pages were not shown at the footer in the specification pages.


raquelarrojo (changes requested):
This was firstly done as in your change. But we realised that if we use `startsWith` and if the name of two components starts with the same word it is not working fine. This happens for example in the radio and radio group. In the radio group, the footer is still showing Radio group as the next page.

PR #965, opened 2022-05-12
Fix doc footer from the new website

GomezIvann:
Fixing the issue of the documentation footer regarding components whose path start with the same string, such as radio and radio-group ("/components/radio" and "/components/radio-group").


marcialfps (approved):
💯

PR #971, opened 2022-05-17
[Major] Changed tokens and fixed width for small screens in paginator

Lisa159:
Fixed to be able to see correctly select components in small screens.


lib/src/paginator/Paginator.tsx
```diff
@@ -155,6 +155,7 @@ const DxcPaginatorContainer = styled.div`
 `;
 
 const SelectContainer = styled.div`
+  min-width: 82px;
```
aweell, line 158: I recommend removing the `max-width` to enable large numbers when there's available space and changing the `min-width` to 5.25rem in order to increase the value depending on the root font size.

aweell (changes requested):

lib/src/paginator/Paginator.tsx
Lisa159: Done

raquelarrojo (approved):

aweell (approved):

PR #976, opened 2022-05-19
[Minor] Internationalization of the hardcoded strings

jsuarezgonz:
- A new property is added to HalstackProvider to receive the translation of the component literals.

- The default definition of the strings used by each component is created in the variables file, which will be used if no translation is received.

- The components that had hardcoded strings are modified so that they use the new internationalization structure.


app/src/pages/Paginator.jsx
```diff
@@ -13,6 +13,13 @@ function App() {
     },
   };
 
+  const labels = {
+    paginator: {
+      itemsPerPage: " Num. items",
+      pageOfText: (a, b) => `Pagina: ${a} de ${b}`,
```
raquelarrojo, line 19: I think it should be `itemsPerPageText` instead of `itemsPerPage`.

lib/src/select/Select.tsx
```diff
@@ -131,6 +132,7 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
     const selectOptionsListRef = useRef(null);
```
raquelarrojo, line 132: We could delete `getNotOptionalErrorMessage` since it is now not used.

raquelarrojo (changes requested):
We could also add the `Contents` title of the QuickNav to be internationalised.

raquelarrojo (approved):

raquelarrojo (approved):

PR #992, opened 2022-05-25
Adding code tab to box page

Jialecl:


marcialfps (changes requested):

website/screens/components/box/code/examples/Basic.ts
```diff
@@ -0,0 +1,27 @@
...
+                Box Content
+            </DxcBox>
+            <DxcBox shadowDepth={1} padding="medium">
+                Box Content
+            </DxcBox>
+            <DxcBox shadowDepth={2} padding="medium">
+                Box Content
+            </DxcBox>
+          </DxcStack>
+        </DxcInset>
```
raquelarrojo, line 17: Something strange is happening here, since the box default size is fitContent but its showing like fillParent.

raquelarrojo (changes requested):

website/screens/components/box/code/examples/Basic.ts
Jialecl: Yes, this is caused by the DxcStack.

Jialecl: It can be fixed if instead  of `flex` it is `inline-flex`

website/screens/components/box/code/BoxCodePage.tsx
```diff
@@ -4,7 +4,7 @@ import QuickNavContainer from "@/common/QuickNavContainer";
 import Code from "@/common/Code";
 import DocFooter from "@/common/DocFooter";
 import Example from "@/common/example/Example";
-import basicUsage from "./examples/BasicUsage";
+import basic from "./examples/Basic";
```
raquelarrojo, line 8: This file should be `basicUsage` instead of `Basic`.

website/screens/components/box/usage/BoxUsagePage.tsx
```diff
@@ -3,7 +3,7 @@ import QuickNavContainerLayout from "@/common/QuickNavContainerLayout";
 import QuickNavContainer from "@/common/QuickNavContainer";
 import Code from "@/common/Code";
 import DocFooter from "@/common/DocFooter";
-import basic from "./examples/Basic"
+import basicUsage from "./examples/BasicUsage";
 import Example from "@/common/example/Example";
```
raquelarrojo, line 7: This file should be `basicUsage` instead of `BasicUsage`.

raquelarrojo (changes requested):

raquelarrojo (approved):

PR #999, opened 2022-05-26
Adding code tab to date component

Jialecl:


website/screens/components/date-input/code/DateInputCodePage.tsx
```diff
@@ -0,0 +1,224 @@
...
+          <>
+            <DxcText as="p">
+              To indicate that the date is optional, you can use the <Code>optional</Code> prop.
+            </DxcText>
+            <Example example={optional} defaultIsVisible />
+          </>
+        ),
+      },
+      {
+        title: "Error management",
```
GomezIvann, line 199: I think **'Error handling'** title fits the use case better.

website/screens/components/date-input/code/examples/errorManagement.ts
```diff
@@ -0,0 +1,39 @@
+import { DxcDateInput, DxcInset } from "@dxc-technology/halstack-react";
+import { useState } from "react";
+
+const code = `() => {
+    const [value, setValue] = useState("");
+    const [error, setError] = useState();
+    const onChange = ({ value }) => {
+    setValue(value);
```
GomezIvann, line 8: Indent this line.

website/screens/components/date-input/code/examples/errorManagement.ts
```diff
@@ -0,0 +1,39 @@
...
+    };
+    const onBlur = ({ value, error }) => {
+    setValue(value);
```
GomezIvann, line 11: Indent this and the following lines.

website/screens/components/date-input/code/examples/optional.ts
```diff
@@ -0,0 +1,20 @@
+import { DxcDateInput, DxcInset } from "@dxc-technology/halstack-react";
+
+const code = `() => {
+    return(
+        <DxcInset space="large">
+            <DxcDateInput
+            label="Default"
```
GomezIvann, line 7: Indent this and the following lines.

website/screens/components/date-input/code/examples/basic.ts
```diff
@@ -0,0 +1,32 @@
+import { DxcDateInput, DxcStack, DxcInset } from "@dxc-technology/halstack-react";
+
+const code = `() => {
+    return(
+        <DxcInset space="large">
+            <DxcStack gutter="small">
+                <DxcDateInput
+                label="Default"
```
GomezIvann, line 8: Indent this and the following lines.

GomezIvann (changes requested):

GomezIvann:
Controlled and uncontrolled examples have been removed, we should add them.

website/screens/components/date-input/code/examples/basic.ts
```diff
@@ -0,0 +1,32 @@
...
+                placeholder="Placeholder"
+                />
+                <DxcDateInput
```
GomezIvann, line 11: Remove this example, since this should be on a `Usage` tab showcase.

website/screens/components/date-input/code/examples/basic.ts
```diff
@@ -0,0 +1,32 @@
...
+                label="Error"
+                placeholder="Placeholder"
+                error="Error message"
+                />
+                <DxcDateInput
```
GomezIvann, line 16: Remove this example, since this should be on a `Usage` tab showcase.

GomezIvann (changes requested):

website/screens/components/date-input/code/examples/basic.ts
Jialecl: I removed the examples with error and disabled and only left the most basic one.

website/screens/components/date-input/code/examples/basic.ts
```diff
@@ -1,25 +1,10 @@
...
-                <DxcDateInput
-                label="Disabled"
-                placeholder="Placeholder"
-                disabled
-                />
-            </DxcStack>
-        </DxcInset>
+  return (
+    <DxcInset space="large">
+        <DxcDateInput label="Default" placeholder="Placeholder" />
```
GomezIvann, line 6: I suggest adding size="fillParent" to the example to improve the look of the component:

image

website/screens/components/date-input/code/examples/optional.ts
```diff
@@ -1,14 +1,10 @@
...
-        <DxcInset space="large">
-            <DxcDateInput
-            label="Default"
-            placeholder="Placeholder"
-            optional
-            />
-        </DxcInset>
+  return (
+    <DxcInset space="large">
+      <DxcDateInput label="Default" placeholder="Placeholder" optional />
```
GomezIvann, line 6: I don't think this example gives any kind of value or extra explanation to the user... Maybe we should remove it 🤔.

website/screens/components/date-input/code/DateInputCodePage.tsx
```diff
@@ -196,13 +198,27 @@ const sections = [
...
       {
-        title: "Error management",
+        title: "Error handling",
         content: (
           <>
-            <DxcText as="p">
-              Example of error management when it does not match the format.
-            </DxcText>
-            <Example example={errorManagement} defaultIsVisible />
+            <DxcText as="p">Example of error handling when it does not match the format.</DxcText>
```
GomezIvann, line 204: We should explain a bit more how error handling works, since it does more than it seems. Look at [Radio Group's example](https://developer.dxc.com/tools/react/next/#/components/radioGroup) for Error handling.

website/screens/components/date-input/code/examples/errorHandling.ts
```diff
@@ -0,0 +1,39 @@
...
+  const onBlur = ({ value, error }) => {
+    setValue(value);
+    setError(error);
+  };
+
+  return (
+    <DxcInset space="large">
+      <DxcDateInput
+        label="Formatted"
+        helperText="If the typed date doesn't match the defined format, an error will be displayed"
```
GomezIvann, line 19: We can add this helper text here as a regular text for the example, to give the user an explanation of how uncontrolled components work.

GomezIvann (changes requested):

website/screens/components/date-input/code/examples/optional.ts
Jialecl: I think it helps some teams since many of them are more used to required components than optional, but I also see that it could be unnecessary if they read all the props.

But I like to give a little bit more attention to this.

website/screens/components/date-input/code/examples/basic.ts
GomezIvann: After the poll, we need to do the change to have every example look the same.

website/screens/components/date-input/code/DateInputCodePage.tsx
```diff
@@ -0,0 +1,239 @@
+import { DxcStack, DxcTable, DxcText } from "@dxc-technology/halstack-react";
+import QuickNavContainer from "@/common/QuickNavContainer";
+import QuickNavContainerLayout from "@/common/QuickNavContainerLayout";
+import DocFooter from "@/common/DocFooter";
+import Code from "@/common/Code";
+import Example from "@/common/example/Example";
+import basic from "./examples/basic";
+import optional from "./examples/optional";
```
GomezIvann, line 8: Remove unused import which is causing problems.

website/screens/components/date-input/code/DateInputCodePage.tsx
```diff
@@ -187,21 +187,17 @@ const sections = [
         ),
       },
       {
-        title: "Optional",
+        title: "Error handling",
         content: (
           <>
             <DxcText as="p">
-              To indicate that the date is optional, you can use the <Code>optional</Code> prop.
+              Example of error handling when it does not match the format. If the typed date does
```
GomezIvann, line 194: The first line "Example of error handling when it does not match the format." is a bit redundant with the second one "If the typed date does not match the defined format, an error will be displayed". Also, this can be a bit misleading, since the example manages any kind of errors, not just the format ones (for example, required).

website/screens/components/date-input/code/examples/uncontrolled.ts
```diff
@@ -4,8 +4,9 @@ const code = `() => {
   return (
     <DxcInset space="large">
       <DxcDateInput
+        size="fillParent"
         label="Uncontrolled"
-        helperText="The 'defaultValue' prop only works with uncontrolled date inputs"
+        helperText="Uncontrolled date input"
```
GomezIvann, line 9: The word **"Uncontrolled"** is three times in this example, I suggest at least removing one of them (maybe changing the helper text to "Helper text" or simply removing it).

GomezIvann (changes requested):

website/screens/components/date-input/code/examples/uncontrolled.ts
Jialecl: done

website/screens/components/date-input/code/DateInputCodePage.tsx
Jialecl: done

website/screens/components/date-input/code/DateInputCodePage.tsx
Jialecl: done

GomezIvann (changes requested):

website/screens/components/date-input/code/examples/uncontrolled.ts
```diff
@@ -0,0 +1,22 @@
+import { DxcDateInput, DxcInset } from "@dxc-technology/halstack-react";
+
+const code = `() => {
```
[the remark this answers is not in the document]
Jialecl: I feel that it can also make the example more confusing since what we want to demonstrate is the use of an uncontrolled date as neither other example adds any interaction with other components. 
Also we might be implying that this is how we want forms to be implemented.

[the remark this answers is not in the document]
GomezIvann: This example was agreed upon a few months ago and it's the same for every input/select/textarea component. The main goal here is not to showcase how you should do a form, instead, it teaches the new users how to extract the input value of an uncontrolled component using the `ref` prop. In this case, it is necessary to use another component since you can achieve that demostration with just an input.

website/screens/components/date-input/code/DateInputCodePage.tsx
```diff
@@ -0,0 +1,238 @@
...
+            <Example example={basic} defaultIsVisible />
+          </>
+        ),
+      },
+      {
+        title: "Error handling",
+        content: (
+          <>
+            <DxcText as="p">
+              Example of error handling, an error will be displayed whenever an error is sent as a
```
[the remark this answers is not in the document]
GomezIvann: We can simply remove that line since it doesn't add too much value and, instead, write ONLY: "The component behaviour varies depending on the value of the `error` prop. We recommend reading the description of the prop carefully to fully understand the following example."

website/screens/components/date-input/code/examples/uncontrolled.ts
```diff
@@ -1,22 +1,42 @@
...
-      <DxcDateInput
-        size="fillParent"
-        label="Uncontrolled"
-        helperText="Helper text"
-        defaultValue="10-08-1998"
-        clearable
-      />
-    </DxcInset>
+    <>
+      <DxcInset space="large">
```
GomezIvann, line 14: Update the returned JSX code to match the current uncontrolled example in other components:
```
  return (
    <DxcInset space="large">
      <DxcStack gutter="large" align="start">
        <DxcSelect
          label="Label"
          defaultValue="4"
          options={options}
          size="fillParent"
          ref={selectRef}
        />
        <DxcButton
          onClick={handleSubmit}
          label="Submit"
        />
      </DxcStack>
    </DxcInset>
  );
```

GomezIvann (changes requested):

website/screens/components/date-input/code/examples/uncontrolled.ts
```diff
@@ -12,7 +12,7 @@ const code = `() => {
   return (
     <>
```
GomezIvann, line 13: Remove `React.Fragment`, since it does nothing here.

website/screens/components/date-input/code/DateInputCodePage.tsx
```diff
@@ -190,12 +190,9 @@ const sections = [
         content: (
           <>
             <DxcText as="p">
-              Example of error handling, an error will be displayed whenever an error is sent as a
-              parameter.
-            </DxcText>
-            <DxcText as="p">
-              Take into account that it behaves differently based on the value of error (check the
-              prop above for more information).
+              The component behaviour varies depending on the value of the error prop. We recommend
```
GomezIvann, line 193: I suggest changing a little bit the first sentence to avoid the repetition of "prop" word: "The component behaviour varies depending on the value of `<Code>error</Code>`." 🤔

GomezIvann (changes requested):

GomezIvann (approved):

PR #1025, opened 2022-05-31
Changed code to Code component in Checkbox

aidamag:
Changed code to Code component in Checkbox and removed unused image.

### DIFF
--- a/website/screens/components/checkbox/code/CheckboxCodePage.tsx
+++ b/website/screens/components/checkbox/code/CheckboxCodePage.tsx
@@ -119,7 +119,7 @@ const sections = [
     title: "Examples",
     subSections: [
       {
-        title: "Basic Usage",
+        title: "Basic usage",
         content: (
           <>
             <Example example={basic} defaultIsVisible />


GomezIvann:
Some comments about the checkbox examples:

- By default, the code is only opened for the first example and not for the rest. Everyone should have the code opened from the beginning.
- I don't find very interesting the example _Label position_, the explanation provided for the prop is enough, don't you think?
- Code indentation is not 2 spaces as we are currently doing on the other examples.
- Rename the 'Basic' example to 'Basic Usage' as it is done in other components.
- Remove every margin from the examples. Try to have the same space around for every component.
- For the first example ('Basic Usage'), here is a way to do the same but with less code:

```
() => {
  return (
    <DxcInset space="xlarge">
      <DxcRow gutter="large">
        <DxcCheckbox label="Default" />
        <DxcCheckbox label="Label" optional />
        <DxcCheckbox label="Disabled" disabled />
      </DxcRow>
    </DxcInset>
  );
}
```
Also, I don't think we should do examples for optional and disabled, since they don't require further explanation than what the prop description says, in terms of code of course (maybe in Usage tab...).

GomezIvann (changes requested):
Some comments about the checkbox examples:

- By default, the code is only opened for the first example and not for the rest. Everyone should have the code opened from the beginning.
- I don't find very interesting the example _Label position_, the explanation provided for the prop is enough, don't you think?
- Code indentation is not 2 spaces as we are currently doing on the other examples.
- Rename the 'Basic' example to 'Basic usage' as it is done in other components.
- Remove every margin from the examples. Try to have the same space around for every component.
- For the first example ('Basic usage'), here is a way to do the same but with less code:

```
() => {
  return (
    <DxcInset space="xlarge">
      <DxcRow gutter="large">
        <DxcCheckbox label="Default" />
        <DxcCheckbox label="Label" optional />
        <DxcCheckbox label="Disabled" disabled />
      </DxcRow>
    </DxcInset>
  );
}
```
Also, I don't think we should do examples for optional and disabled, since they don't require further explanation than what the prop description says, in terms of code of course (maybe in Usage tab...).

PR #1049, opened 2022-06-06
Add Code tab to Radio page

Lisa159:


website/screens/components/radio/code/RadioCodePage.tsx
```diff
@@ -0,0 +1,160 @@
...
+          <td>
+            Size of the component ('small' | 'medium' | 'large' | 'fillParent' |
+            'fitContent').
+          </td>
+        </tr>
+      </DxcTable>
+    ),
+  },
+  {
+    title: "Examples",
```
Jialecl, line 108: I think that having only the default and the radio group as examples is enough.

Lisa159: Done!

Jialecl (approved):

PR #1053, opened 2022-06-07
Add Code tab to Progress Bar page

Lisa159:


website/screens/components/progress-bar/usage/ProgressBarUsagePage.tsx
```diff
@@ -65,12 +66,7 @@ const sections = [
     title: "Determinate or indeterminate",
     content: (
       <>
-        <Figure caption="Determinate vs undeterminate progress bar">
-          <Image
-            src={determinateImage}
-            alt="Determinate vs undeterminate progress bar"
-          />
-        </Figure>
+        <Example example={determinate_indeterminate} />
```
Jialecl, line 69: I think the same could be done with the variants example.

Use a button to open the overlay progress bar, because an image does not depict the component correctly.

Also some sentences in Do's and Don'ts could be rephrased to make it clearer

Jialecl, line 69: I think the same could be done with the variants example.

Use a button to open the overlay progress bar, because an image does not depict the component correctly.

Also some sentences in Do's and Don'ts could be rephrased to make it clearer

Lisa159: Done

Jialecl (dismissed):

PR #1066, opened 2022-06-08
Fixing links and upgrading halstack-react

Jialecl:


jsuarezgonz (changes requested):
Could you try using DxcInset instead of <p> in the examples?

jsuarezgonz (approved):

PR #1068, opened 2022-06-09
Add Code tab to Tag page

Lisa159:


Jialecl:
There are two sentences in usage that needs to be changed:

Use the tag component for highlight properties of a section that want to remark. (change the "for" to "to")
If there are multiple tags, the space between the should be at least 16px. (the second "the" should be "them")

Jialecl (approved):

PR #1084, opened 2022-06-13
Update examples from Select code page 

GomezIvann:
Update the examples of select's code page after some discussions.


marcialfps (approved):
💯

PR #1094, opened 2022-06-16
Review components pages

raquelarrojo:


website/screens/components/header/code/HeaderCodePage.tsx
```diff
@@ -120,12 +120,12 @@ const sections = [
...
       {
-        title: "With custom content",
-        content: <Example example={withCustomContent} defaultIsVisible />,
+        title: "Custom content",
+        content: <Example example={customContent} defaultIsVisible />,
       },
       {
-        title: "With dropdown",
-        content: <Example example={withDropdown} defaultIsVisible />,
+        title: "Dropdown Header",
```
Lisa159, line 127: Change Header to header.

raquelarrojo: It refers to Header component and we are putting the components in upper case, so to be consistent here I would keep it as Header.

Lisa159: So in that case we should check the other references to the header in this page? Because they are all in lower case. It is true that they can have a second meaning, referring to our component or just a normal header... How do you see it?

website/screens/components/wizard/code/examples/uncontrolled.ts
```diff
@@ -1,44 +1,30 @@
...
-            },
-         ]}
-        ></DxcWizard>
-      </DxcInset>
-    );
-  }`;
+  return (
+    <DxcInset space="large">
+      <DxcWizard
+        defaultCurrentStep={1}
```
Lisa159, line 11: We could remove `defaultCurrentStep` to make the code cleaner.

website/screens/components/header/code/HeaderCodePage.tsx
raquelarrojo: For me it is the same case as here: 
image

We are specifying that the Accordion is the Accordion component and that's why it is in upper case. In this case we want to say that it is an example of the Dropdown of the Header. Maybe we could change it to `Dropdown` only?

website/screens/components/wizard/code/examples/uncontrolled.ts
raquelarrojo: I think it could be a good idea to add `defaultValue` to all uncontrolled examples. It would show how it works.

Lisa159: Yes, but why only that prop and not any other? I mean, it is just a prop that doesn't really have any difficulty to add, not like in in the case of onChange or onBlur that needs to create an additional function.

raquelarrojo: `defaultValue` prop is used only for uncontrolled components, for giving them a default value without being a controlled one. That's why it is useful in an uncontrolled example.

website/screens/components/header/code/HeaderCodePage.tsx
raquelarrojo: I would put `Dropdown` only, as we put `Icons` for the examples with icons. Here if we put `Dropdown`, we are saying that it is an example with a dropdown. If we put `Dropdown child` maybe it could be doubtful.

Lisa159: Okay 👍👍

website/screens/components/wizard/code/examples/uncontrolled.ts
Lisa159: Fair enough, you are right.

website/screens/components/header/code/HeaderCodePage.tsx
raquelarrojo: Done!

PR #1114, opened 2022-06-21
[Major] Select listbox now uses Radix UI popover

GomezIvann:
Select component is now supported by Radix UI Primitive: [Popover](https://www.radix-ui.com/docs/primitives/components/popover). This has some major improvements:

1. **The component now has collision detection.** This changes the listbox position based on the current scroll and select positioning to help the user with an always visible listbox.

https://user-images.githubusercontent.com/44321109/174848668-947588e7-96e0-4d1e-9d1e-607fb85fbf47.mp4

2. **Fixes z-index issues with material components.** The listbox now appears within a Portal with absolute positioning, so it won't be stucked behind other components such us Dialog, Button, Checkbox or anyone with a high z-index value.

https://user-images.githubusercontent.com/44321109/174847761-908ec900-34b9-43d5-9b62-991811490ad6.mp4

### DIFF
--- a/lib/src/select/Listbox.tsx
+++ b/lib/src/select/Listbox.tsx
@@ -22,11 +22,13 @@ const Listbox = React.forwardRef<ListboxRefType, ListboxProps>(
       optionalItem,
       searchable,
       handleOptionOnClick,
+      styles,
     },
     ref
   ): JSX.Element => {
     const colorsTheme = useTheme();
     const translatedLabels = useTranslatedLabels();
+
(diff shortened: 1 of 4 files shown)


Lisa159:
What about the z-index with the header? Do you see fine that it is not behind the header?

image

GomezIvann:
> What about the z-index with the header? Do you see fine that it is not behind the header?
> 
> image

This is not a problem with the z-index, Portals operate a little bit different since they are outside of the regular DOM tree, so they will always appear on top of every part of the layout (unless it is another portal, like the Dialog, there it would a z-index issue).

Lisa159 (approved):

PR #1116, opened 2022-06-22
New DxcRows component

### LINKED ISSUE #1105: Implement DxcRow and DxcRows components

marcialfps:
Closes #1105


lib/src/rows/Rows.stories.tsx
```diff
@@ -0,0 +1,314 @@
...
+import DxcRows from "./Rows";
+
+export default {
+  title: "Rows",
+  component: DxcRows,
+};
+
+export const Chromatic = () => (
+  <>
+    <Title title="Default" theme="light" level={4} />
```
GomezIvann, line 13: Is not the default value `strech`?

lib/src/rows/Rows.stories.tsx
```diff
@@ -0,0 +1,314 @@
...
+          <Placeholder width="medium" height="medium" />
+          <Placeholder width="small" height="small" />
+        </DxcRows.Row>
+        <DxcRows.Row>
+          <Placeholder width="large" height="large" />
+          <Placeholder width="small" height="small" />
+        </DxcRows.Row>
+      </DxcRows>
+    </Container>
+    <Title title="AlignY = stretch" theme="light" level={4} />
```
GomezIvann, line 134: The Placeholder is overflowing with `alignY="strech"`.

image

lib/src/rows/Rows.tsx
```diff
@@ -0,0 +1,65 @@
...
+  alignX = "stretch",
+  alignY = "stretch",
+  as = "div",
+  divider = false,
+  gutter = "0rem",
+  reverse = false,
+  children,
+}: RowsProps): JSX.Element => {
+  return (
+    <RowsContainer as={as} alignX={alignX} gutter={gutter} reverse={reverse}>
```
GomezIvann, line 15: I suggest calling this container just Rows to be more semantic.

lib/src/rows/Rows.tsx
```diff
@@ -0,0 +1,65 @@
...
+          </>
+        );
+      })}
+    </RowsContainer>
+  );
+};
+
+const DxcRow = ({ height = "auto", alignY, children }: RowProps): JSX.Element => {
+  return (
+    <StyledRow height={height} alignY={alignY}>
```
GomezIvann, line 30: Same as above, I suggest calling this container just Row to be more semantic.

lib/src/rows/Rows.stories.tsx
```diff
@@ -0,0 +1,314 @@
...
+        </DxcRows.Row>
+        <DxcRows.Row>
+          <Placeholder width="medium" />
+        </DxcRows.Row>
+        <DxcRows.Row>
+          <Placeholder width="large" />
+        </DxcRows.Row>
+      </DxcRows>
+    </Container>
+    <Title title="AlignX = stretch" theme="light" level={4} />
```
GomezIvann, line 69: This should be AlignX = start example instead of stretch.

lib/src/rows/Rows.tsx
```diff
@@ -0,0 +1,65 @@
...
+  height: 1px;
+  background-color: #999999;
+`;
+
+const StyledRow = styled.div<RowProps>`
+  ${({ height }) =>
+    height === "content"
+      ? "height: fit-content"
+      : height === "auto"
+      ? "flex-grow: 1;"
```
GomezIvann, line 58: Should we use `flex: 1` instead of `flex-grow: 1`? I'm not sure about the impact of `flex-basis` being set to auto with our current option...

lib/src/rows/Rows.tsx
```diff
@@ -0,0 +1,65 @@
...
+DxcRows.Row = DxcRow;
+
+const RowsContainer = styled.div<RowsProps>`
+  display: flex;
+  ${({ alignX, gutter, reverse }) => `
+    flex-direction: ${reverse ? "column-reverse" : "column"};
+    align-items: ${alignX};
+    gap: ${gutter};
+  `}
+  flex-grow: 1;
```
GomezIvann, line 45: Should we use `flex: 1` instead of `flex-grow: 1`? I'm not sure about the impact of `flex-basis` being set to auto with our current option...

GomezIvann (changes requested):
The component should allow this kind of syntax:
```
      <DxcRows>
        <Placeholder width="small" />
        <Placeholder width="medium" />
        <Placeholder width="large" />
      </DxcRows>
```

lib/src/rows/Rows.stories.tsx
marcialfps: Yes, but this example have a fixed width.

```
<DxcRows>
  <DxcRows.Row>
    <Placeholder width="small" />
  </DxcRows.Row>
  <DxcRows.Row>
    <Placeholder width="medium" />
  </DxcRows.Row>
  <DxcRows.Row>
    <Placeholder width="large" />
  </DxcRows.Row>
</DxcRows>
```

lib/src/rows/Rows.stories.tsx
marcialfps: Done!

lib/src/rows/Rows.tsx
marcialfps: Done!

lib/src/rows/Rows.tsx
marcialfps: Done!!

lib/src/rows/Rows.stories.tsx
marcialfps: Done!

lib/src/rows/Rows.tsx
marcialfps: I have tested it and it works exactly the same. It seems that the `flex-basis` value has no impact.

GomezIvann (commented):
1. Do you find it interesting to add an example like the default one but with a divider to check the overlapping between divider and row?

image

Code:
```
    <Container>
      <DxcRows divider>
        <Placeholder width="auto" />
        <Placeholder width="auto" />
        <Placeholder width="auto" />
      </DxcRows>
    </Container>
```

2. Every example from the storybook has three rows, I suggest changing some of them to test other numbers of rows (maybe one row, ten rows, etc).

lib/src/rows/types.tsx
```diff
@@ -0,0 +1,42 @@
...
+export type RowProps = {
+  /**
+   * Fixed width of the single row.
+   */
+  height?: "auto" | "content" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8";
+  /**
+   * Custom content inside the row.
+   */
+  children: React.ReactNode;
+  alignY?: "start" | "end" | "center" | "baseline" | "stretch";
```
GomezIvann, line 41: As we are passing `alignY` as a prop, anyone from the outside can use it, both in Rows and Row. We need to pass the value from father to child by using a context to avoid external usage.

image

lib/src/rows/Rows.tsx
```diff
@@ -0,0 +1,69 @@
...
+  );
+};
+
+DxcRows.Row = DxcRow;
+
+const Rows = styled.div<RowsProps>`
+  display: flex;
+  ${({ alignX, gutter, reverse }) => `
+    flex-direction: ${reverse ? "column-reverse" : "column"};
+    align-items: ${alignX};
```
GomezIvann, line 46: I didn't think about this the first time, but, why are we using `align-items` (which aligns items on the cross axis) to align items on the main axis? Shouldn't we be using `justify-content` for that purpose?

image

from: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Aligning_Items_in_a_Flex_Container

GomezIvann: I know it's working but I find it a bit confusing...

GomezIvann (changes requested):

lib/src/rows/Rows.tsx
```diff
@@ -0,0 +1,69 @@
...
+const Row = styled.div<RowProps>`
+  ${({ height }) =>
+    height === "content"
+      ? "height: fit-content"
+      : height === "auto"
+      ? "flex-grow: 1;"
+      : `height: ${(parseInt(height) / 8) * 100}%`};
+
+  display: flex;
+  align-items: ${({ alignY }) => alignY};
```
GomezIvann, line 66: We should take into account that our values "start" and "end" are not the same as "flex-start" or "flex-end", respectively. This has some connotations that maybe can affect our layout components:

image

Other is that `flex-start` takes into account the presence of the "-reverse" values of the `flex-direction`, while `start` does not. This only happens at Firefox browser.

See: [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content), [css-tricks](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#aa-justify-content) and this question at [stack-overflow](https://stackoverflow.com/questions/50919447/flexbox-flex-start-self-start-and-start-whats-the-difference#:~:text=flex%2Dstart%20takes%20into%20account,direction%2C%20while%20start%20does%20not.).

GomezIvann: What do you think? Should we go for the more generic approach (as it stands right now) or since we are only using flexbox containers (our layout is based on flex) should we use `flex-XXX` properties?

GomezIvann (changes requested):

lib/src/rows/types.tsx
marcialfps: Done

lib/src/rows/Rows.tsx
marcialfps: We are using a flex with column orientation, so I think this changes the order of the axes: `align-items` for the main axis and `justify-content` for the cross axis.

lib/src/rows/Rows.tsx
marcialfps: I think is better to use the corresponding values for flex.

lib/src/rows/Rows.tsx
GomezIvann: You are right!

GomezIvann (approved):

PR #1118, opened 2022-06-22
New DxcColumns component

### LINKED ISSUE #1117: Implement DxcColumn and DxcColumns components

marcialfps:
Closes #1117

### DIFF
--- a/lib/src/columns/Columns.stories.tsx
+++ b/lib/src/columns/Columns.stories.tsx
@@ -331,6 +331,42 @@ export const Chromatic = () => (
         </DxcColumns.Column>
       </DxcColumns>
     </Container>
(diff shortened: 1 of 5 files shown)


GomezIvann (changes requested):
1. Every example from the storybook has three columns, I suggest changing some of them to test other number of columns (maybe one column, ten columns, etc)
2. Do you find it interesting to add an example like the default one but with a divider to check the overlapping between divider and column?

image

Code:
```
    <Container>
      <DxcColumns divider>
        <Placeholder width="auto" />
        <Placeholder width="auto" />
        <Placeholder width="auto" />
      </DxcColumns>
    </Container>
```

3. Each column has the border issue we had with the rows (overlapping with the next one and going outside in the last one) 

image

PR #1122, opened 2022-06-22
[Major] Sidenav Content Style 

### LINKED ISSUE #949: Change sidenav content styles

Jialecl:
Closes #949


lib/src/sidenav/Sidenav.stories.tsx
```diff
@@ -3,180 +3,271 @@ import styled from "styled-components";
...
-const StyledContainer = styled["div"]`
-  display: flex;
-  flex-direction: row;
-  height: 500px;
-`;
-
 export const Chromatic = () => (
   <>
     <ExampleContainer>
       <Title title="Deafult sidenav" theme="light" level={4} />
```
GomezIvann, line 16: Typo, "Default Sidenav" instead.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,148 @@
...
+        </SidenavGroupTitleButton>
+      ) : (
+        title && (
+          <SidenavGroupTitle>
+            {icon}
+            {title}
+          </SidenavGroupTitle>
+        )
+      )}
+      {(!collapsable || (collapsable && !collapsed)) && children}
```
GomezIvann, line 90: This line can be synthesized to just `!collapsed && children` since the `collapsed` variable is never going to be true unless the `SidenavGroupTitleButton` changes its value, whose presence is determined by `collapsable`, so...

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,148 @@
...
+const Group = ({ children, title, collapsable = false, icon }: SidenavGroupPropsType): JSX.Element => {
+  const [collapsed, setCollapsed] = useState(false);
   return (
-    <SidenavMenuLink tabIndex={tabIndex} href={href} onClick={handleClick}>
-      {children}
-    </SidenavMenuLink>
+    <SidenavGroup>
+      {collapsable && title ? (
+        <SidenavGroupTitleButton aria-expanded={collapsed} onClick={() => setCollapsed(!collapsed)}>
+          <span>
```
GomezIvann, line 76: This must be a styled component. `SidenavGroupTitle` and this span share the same CSS rules, so it can be reused here and removed the common CSS rules from `SidenavGroup` and its `span {}` selector, padding, box-sizing, font rules, etc.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,148 @@
...
 
+const DxcSidenav = ({ children }: SidenavPropsType): JSX.Element => {
+  const colorsTheme = useTheme();
   return (
     <ThemeProvider theme={colorsTheme.sidenav}>
-      <SidenavContainer padding={padding}>
-        <BackgroundColorProvider color={colorsTheme.sidenav.backgroundColor}>{children}</BackgroundColorProvider>
+      <SidenavContainer>
+        <BackgroundColorProvider color={colorsTheme.sidenav.backgroundColor}>
+          {React.Children.map(children, (child: ReactElement<any>) => {
```
GomezIvann, line 43: Remove `<any>`, it is not necessary.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,148 @@
...
+            if (child.type === Title) {
+              return child;
+            }
+          })}
+          <DxcStack divider={true} gutter="small">
+            {React.Children.map(children, (child: ReactElement<any>) => {
```
GomezIvann, line 49: Remove `<any>`, it is not necessary.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,148 @@
...
+      <SidenavLink
+        selected={selected}
+        href={href ? href : undefined}
+        target={href ? (newWindow ? "_blank" : "_self") : undefined}
+        ref={ref}
+        onClick={handleClick}
+        tabIndex={tabIndex}
+        {...otherProps}
+      >
+        <span>
```
GomezIvann, line 124: This should be a style-component, like this:

```
const SidenavLinkContent = styled.span`
  display: flex;
  align-items: center;
  
  svg {
    width: 16px;
    fill: currentColor;
    margin-right: 8px;
  }
`;
```

lib/src/sidenav/types.ts
```diff
@@ -1,55 +1,78 @@
-type Space = "xxsmall" | "xsmall" | "small" | "medium" | "large" | "xlarge" | "xxlarge";
+type SVG = React.SVGProps<SVGSVGElement>;
```
GomezIvann, line 1: Why are we only accepting SVG types? Shouldn't we allow them to also pass URLs (`icon: SVG | string`)?

lib/src/sidenav/types.ts
```diff
@@ -1,55 +1,78 @@
...
    */
   children: React.ReactNode;
+  /**
+   * The title of the sidenav group.
+   */
+  title?: string;
+  /**
+   * If true the sidenav group title will be considered a button and the group will be collapsable.
+   */
+  collapsable: boolean;
```
GomezIvann, line 40: `collapsable` shouldn't it be an optional prop? (`false` by default).

GomezIvann: Is currently throwing errors in the stories because of it.

GomezIvann (changes requested):
Update /docs examples from Sidenav page:

image

lib/src/sidenav/Sidenav.tsx
```diff
@@ -67,49 +159,145 @@ const SidenavContainer = styled.div<SidenavPropsType>`
...
+  padding: 7px 24px;
+  span {
+    display: flex;
+    align-items: center;
+    svg {
+      width: 16px;
+      margin-right: 8px;
+    }
+  }
+  &:focus {
```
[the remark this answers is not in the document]
GomezIvann: This same question goes for the hover of each Sidenav Link. This feels a bit strange from Alex's point of view since the new site has hover and active styles, apart from focus, for each Sidenav link. Am I missing something?

lib/src/sidenav/Sidenav.tsx
Jialecl: I already tried that scenario and unfortunately, because of the presence of the collapsable indicator, it does not work correctly sharing the same css. The few similarities are only in the font but they also affect the collapsable indicator.

lib/src/sidenav/Sidenav.tsx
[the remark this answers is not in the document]
Jialecl: I think that Alex did not have time to add the styles to the group title button, since it was added later.
The group title with a dark background were planned to indicate that the group has a link that is selected.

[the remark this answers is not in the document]
Jialecl: We are also missing the selected link hover.

lib/src/sidenav/Sidenav.tsx
Jialecl: Thinking more regarding this I think I understand a bit more about this and I have also applied it to the linkContent mentioned below.

lib/src/sidenav/Sidenav.stories.tsx
```diff
@@ -3,180 +3,195 @@ import styled from "styled-components";
...
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>{" "}
-      </DxcSidenav>
-    </ExampleContainer>
+export const FocusedSidenav = () => (
+  <ExampleContainer pseudoState="pseudo-focus">
+    <Title title="Deafult sidenav" theme="light" level={4} />
```
GomezIvann, line 80: Typo, "Default Sidenav" instead.

lib/src/sidenav/Sidenav.stories.tsx
```diff
@@ -3,180 +3,195 @@ import styled from "styled-components";
...
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>{" "}
-      </DxcSidenav>
-    </ExampleContainer>
+const HoverSidenav = () => (
+  <ExampleContainer pseudoState="pseudo-hover">
+    <Title title="Deafult sidenav" theme="light" level={4} />
```
GomezIvann, line 146: Typo, "Default Sidenav" instead.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,140 @@
...
   );
 };
 
-const Title = ({ children }: SidenavTitlePropsType): JSX.Element => <SidenavMenuTitle>{children}</SidenavMenuTitle>;
-
-const Subtitle = ({ children }: SidenavSubtitlePropsType): JSX.Element => (
-  <SidenavMenuSubTitle>{children}</SidenavMenuSubTitle>
+const Title = ({ children, icon }: SidenavTitlePropsType): JSX.Element => (
+  <SidenavTitle>
+    {typeof icon === "string" ? <img src={icon} /> : icon}
```
GomezIvann, line 55: Sidenav Title Images are not being resize or have any kind of styles at all. This should be fixed:

image

lib/src/sidenav/Sidenav.tsx
```diff
@@ -67,49 +151,148 @@ const SidenavContainer = styled.div<SidenavPropsType>`
...
+    color: #ffffff;
+    background: #4d4d4d;
+    outline: 2px solid #0095ff;
+    outline-offset: -2px;
   }
 `;
 
-DxcSidenav.Title = Title;
-DxcSidenav.Subtitle = Subtitle;
+const SidenavSpan = styled.span`
```
GomezIvann, line 278: Try to avoid using the HTML tag in the styled-components naming. Instead, use names that make sense with the context or their placement inside the component. In this case, for example, `SidenavContent` would fit a little bit better.

GomezIvann (changes requested):
Component documentation in the old site is not updated. Most of the new props are not there or event the compound components (Groups). We need to fix this.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,140 @@
...
+  SidenavGroupPropsType,
+  SidenavLinkPropsType,
+  SidenavSectionPropsType,
+  SidenavTitlePropsType,
+} from "./types.js";
+import DxcStack from "../stack/Stack";
 
-const DxcSidenav = ({ padding, children }: SidenavPropsType): JSX.Element => {
-  const colorsTheme = useTheme();
+const collapsableIcon = (
```
GomezIvann, line 16: The icons of a collapsable group are exchanged:

image

This should be all the way around, as Alex had in the design and we are currently doing in the Select.

image

lib/src/sidenav/Sidenav.tsx
```diff
@@ -67,49 +151,148 @@ const SidenavContainer = styled.div<SidenavPropsType>`
...
+  display: flex;
+  align-items: center;
+
+  svg {
+    width: 16px;
+    margin-right: 8px;
+  }
+`;
+
+const SidenavIconImg = styled.img`
```
GomezIvann, line 288: Same as the `SidenavSpan`, avoid using HTML tags in the styled-components. With `SidenavIcon` it's enough.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,140 @@
...
-      {children}
-    </SidenavMenuLink>
+    <SidenavGroup>
+      {collapsable && title ? (
+        <SidenavGroupTitleButton aria-expanded={collapsed} onClick={() => setCollapsed(!collapsed)}>
+          <SidenavSpan>
+            {typeof icon === "string" ? <SidenavIconImg src={icon} /> : icon}
+            {title}
+          </SidenavSpan>
+          {collapsed ? collapsedIcon : collapsableIcon}
```
GomezIvann, line 72: Swipe these icons. Also, they are currently bigger than what Alex thought for them:

image

Alex design: 

image

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,140 @@
...
+const DxcSidenav = ({ children, title }: SidenavPropsType): JSX.Element => {
+  const colorsTheme = useTheme();
   return (
     <ThemeProvider theme={colorsTheme.sidenav}>
-      <SidenavContainer padding={padding}>
-        <BackgroundColorProvider color={colorsTheme.sidenav.backgroundColor}>{children}</BackgroundColorProvider>
+      <SidenavContainer>
+        <BackgroundColorProvider color={colorsTheme.sidenav.backgroundColor}>
+          {title}
+          <DxcStack divider={true} gutter="small">
```
GomezIvann, line 44: Sidenav's divider is going for the whole width of the component. Alex's design wasn't like it at all:

image

Our implementation:

image

GomezIvann (changes requested):

lib/src/sidenav/Sidenav.tsx
Jialecl: Yes I also saw this issue and I asked him directly regarding this. And having it as 16x16 it was too small so it was increased to 24x24.

lib/src/sidenav/Sidenav.tsx
Jialecl: Yes, I also talked with Alex regarding this issue because it also affects the with of each link and group title that is collapsable.

Reducing the section width won't allow us to take the whole width for the links and groups.

lib/src/sidenav/Sidenav.tsx
Jialecl: One of the possibilities that we thought about was maybe even having only an image in the title.

lib/src/sidenav/Sidenav.tsx
GomezIvann: I keep thinking that 24x24 is too big, maybe we can reach an intermediate solution while trying to preserve the whole size of 24x24:

image

const SidenavCollapseIcon = styled.span`
  display: flex;
  flex-wrap: wrap;
  align-content: center;
  padding: 3px;
  height: 18px;
  width: 18px;
`;

(Forget the margin of the image)

lib/src/sidenav/Sidenav.tsx
GomezIvann: That's because we are currently using our layout components. First, I suggest not using the DxcStack since it's not a final version and will change shortly. And in case it is 100% necessary, to make the content of a section overflow the width set by the divider you can use the component DxcBleed

lib/src/sidenav/Sidenav.tsx
GomezIvann: Letting the user decide their image size is far more problematic than establishing ourselves a max-width and max-height at least.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -67,49 +156,156 @@ const SidenavContainer = styled.div<SidenavPropsType>`
...
-  font-weight: ${(props) => props.theme.subtitleFontWeight};
-  color: ${(props) => props.theme.subtitleFontColor};
-  letter-spacing: ${(props) => props.theme.subtitleFontLetterSpacing};
-  text-transform: ${(props) => props.theme.subtitleFontTextTransform};
-  margin-bottom: 4px;
+const SidenavGroup = styled.div`
+  width: 100%;
+  a {
+    padding: 7px 24px 7px 36px;
+  }
```
raquelarrojo, line 180: Why this value for the padding? We never use odd values, like 7px.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -67,49 +156,156 @@ const SidenavContainer = styled.div<SidenavPropsType>`
...
+  font-style: ${(props) => props.theme.groupTitleFontStyle};
+  font-weight: ${(props) => props.theme.groupTitleFontWeight};
+  font-size: ${(props) => props.theme.groupTitleFontSize};
+  line-height: 19px;
+
+  display: flex;
+  align-items: center;
+  margin: 0px;
+  padding: 7px 24px;
+`;
```
raquelarrojo, line 201: Same here. Why this value for the padding? We never use odd values, like 7px.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -67,49 +156,156 @@ const SidenavContainer = styled.div<SidenavPropsType>`
...
+  font-style: ${(props) => props.theme.groupTitleFontStyle};
+  font-weight: ${(props) => props.theme.groupTitleFontWeight};
+  font-size: ${(props) => props.theme.groupTitleFontSize};
+  line-height: 19px;
+
+  display: flex;
+  align-items: center;
+  margin: 0px;
+  padding: 7px 24px;
+  &:focus {
```
raquelarrojo, line 220: Same here. Why this value for the padding? We never use odd values, like 7px.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -67,49 +156,156 @@ const SidenavContainer = styled.div<SidenavPropsType>`
...
   text-decoration: ${(props) => props.theme.linkTextDecoration};
-  margin: ${(props) =>
-    `${props.theme.linkMarginTop} ${props.theme.linkMarginRight} ${props.theme.linkMarginBottom} ${props.theme.linkMarginLeft}`};
+
+  font-family: ${(props) => props.theme.linkFontFamily};
+  font-style: ${(props) => props.theme.linkFontStyle};
+  font-weight: ${(props) => props.theme.linkFontWeight};
+  font-size: ${(props) => props.theme.linkFontSize};
+  line-height: 19px;
+
```
raquelarrojo, line 251: Same here. Why this value for the line-height? We never use odd values, like 19px.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -67,49 +156,156 @@ const SidenavContainer = styled.div<SidenavPropsType>`
...
+      : `color: ${props.theme.linkFontColor}; background: transparent;`}
+
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+
+  box-shadow: 0 0 0 2px transparent;
+
+  padding: 7px 24px;
+
```
raquelarrojo, line 264: Same here. Why this value for the padding? We never use odd values, like 7px.

raquelarrojo (changes requested):
It is necessary to update also the sidenav in the themebuilder: 
image

lib/src/sidenav/Sidenav.tsx
Jialecl: Yes, you are right I checked the wrong measurements. I added 1 px from the outline in the design.

I will also change the values from px to rem but maintaining the same value and change the 7px.

lib/src/sidenav/Sidenav.test.js
```diff
@@ -1,45 +1,19 @@
...
-        <DxcSidenav.Title>test-title</DxcSidenav.Title>
-        <DxcSidenav.Subtitle>test-subtitle</DxcSidenav.Subtitle>
-        <DxcSidenav.Link href="#">test-link</DxcSidenav.Link>
+        <DxcSidenav.Section>
+          <p>nav-content-test</p>
+          <DxcSidenav.Link href="#">Link</DxcSidenav.Link>
+        </DxcSidenav.Section>
       </DxcSidenav>
     );
-    expect(getByText("test-title")).toBeTruthy();
```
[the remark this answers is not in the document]
Jialecl: I added a test to check the collapsed state, but the tests mentioned are already being tested in the visual tests

docs/src/pages/components/cdk-components/sidenav/Sidenav.jsx
```diff
@@ -28,10 +31,27 @@ function Sidenav() {
           .
         </p>
         <DxcHeading text="DxcSidenav.Title" level={4} weight="bold" />
-        <p>The content will be showed as the main title in the sidenav.</p>
```
raquelarrojo, line 31: `DxcSidenav.Title` is not a children of `DxcSidenav`. The prop `title` should be used instead.

docs/src/pages/components/cdk-components/sidenav/Sidenav.jsx
```diff
@@ -21,7 +21,11 @@ function Sidenav() {
         <SidenavPropsTable />
       </Section>
       <Section>
-        <DxcHeading level={3} text="Children" margin={{ bottom: "small" }} />
+        <DxcHeading
+          level={3}
+          text="Compound components"
+          margin={{ bottom: "small" }}
+        />
         <p>
```
raquelarrojo, line 29: I don't think this is clear in the documentation. I would keep children (because they are children and should be documented like this), but I would put `DxcSidenav.Title` above, outside of children.

raquelarrojo (changes requested):

raquelarrojo:
When an item is selected in the sidenav, there is one moment when the background is not colored. I think that this should not be happening.


https://user-images.githubusercontent.com/10975289/178670267-ad029a8e-ea7d-4e7a-9f0f-a08b07a47e84.mov

docs/src/pages/components/cdk-components/sidenav/Sidenav.jsx
Jialecl: I thought about that and having DxcSidenav.Title above the children it would be in a different hierarchy unless we add another title at the same level as the children title.

raquelarrojo: Yes, when I say that DxcSidenav.Title should not be inside Children, I refer that it is not a children. But we can put a title for it.

raquelarrojo: Old docs not updated.

lib/src/sidenav/types.ts
```diff
@@ -1,55 +1,80 @@
-type Space = "xxsmall" | "xsmall" | "small" | "medium" | "large" | "xlarge" | "xxlarge";
+import React from "react";
```
marcialfps, line 1: You don't need to import React.

lib/src/sidenav/Sidenav.tsx
```diff
@@ -1,56 +1,147 @@
...
   );
 };
 
-const Title = ({ children }: SidenavTitlePropsType): JSX.Element => <SidenavMenuTitle>{children}</SidenavMenuTitle>;
+const Title = ({ children }: SidenavTitlePropsType): JSX.Element => <SidenavTitle>{children}</SidenavTitle>;
 
-const Subtitle = ({ children }: SidenavSubtitlePropsType): JSX.Element => (
-  <SidenavMenuSubTitle>{children}</SidenavMenuSubTitle>
+const Section = ({ children }: SidenavSectionPropsType): JSX.Element => (
+  <DxcBleed left="1rem" right="1rem">
```
marcialfps, line 64: Simplify to:
`horizontal="1rem"`

lib/src/sidenav/Sidenav.tsx
```diff
@@ -67,49 +158,163 @@ const SidenavContainer = styled.div<SidenavPropsType>`
...
+
+  svg {
+    width: 18px;
+    height: auto;
+  }
+`;
+
+type StyledLinkProps = {
+  selected: boolean;
+};
```
marcialfps, line 248: I think it is better to remove this type and indicate it directly in the styled component:
`styled.a<{ selected: boolean; }>`

lib/src/sidenav/Sidenav.stories.tsx
```diff
@@ -3,180 +3,193 @@ import styled from "styled-components";
...
-    <ExampleContainer>
-      <Title title="Sidenav with medium padding" theme="light" level={4} />
-      <DxcSidenav padding="medium">
-        {" "}
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>
-        <p>Lorem ipsum</p>{" "}
+      <Title title="Default sidenav" theme="light" level={4} />
```
marcialfps, line 41: Add cases where icons are used in Groups and Links.

marcialfps (changes requested):

lib/src/sidenav/Sidenav.stories.tsx
```diff
@@ -3,180 +3,179 @@ import styled from "styled-components";
 import DxcSidenav from "./Sidenav";
```
marcialfps, line 3: Remove unused import from styled-components.

marcialfps (approved):

lib/src/sidenav/Sidenav.tsx
Jialecl: As decided the user should be responsible of the image size

GomezIvann (approved):

raquelarrojo (approved):

PR #1137, opened 2022-06-24
[Patch] Button size fit custom label in File Input

### LINKED ISSUE #1082: [File-Input] Button size is not fitting the custom label
**Describe the bug**
When the prop `buttonLabel` has as value a string longer than what it can fit, the button doesn't increase its size, it applies a text-overflow. The button seems to have a size="medium" in the code.
The `filedrop` and `dropzone` modes work fine, which makes it all the more strange.

**Expected behaviour**
The upload button should fit its content when `mode="file"`

**Screenshots**

image

image

marcialfps:
Closes #1082

### DIFF
--- a/lib/src/file-input/FileInput.tsx
+++ b/lib/src/file-input/FileInput.tsx
@@ -197,7 +197,7 @@ const DxcFileInput = ({
               }
               onClick={handleClick}
               disabled={disabled}
-              size="medium"
+              size="fitContent"
               tabIndex={tabIndex}
             />
             {files.map((file) => {


Jialecl (approved):

lib/src/file-input/FileInput.tsx
```diff
@@ -197,7 +197,7 @@ const DxcFileInput = ({
               }
               onClick={handleClick}
               disabled={disabled}
-              size="medium"
+              size="fitContent"
```
Jialecl, line 200: This is not working as intended. It´s working as fillParent.

PR #1138, opened 2022-06-27
[Patch] Focus outline fits Card size

### LINKED ISSUE #970: Card focus doesn't fit the size of the component
The bottom line of the focus doesn't fit the size of the component.

image

Seems to be an issue with the `<a>` tag when acting as a link 🤔

image

marcialfps:
Closes #970


Lisa159:
When a card doesn't have action or link, it should not have focus but it has it. 

Could you remove it?

image

Lisa159 (approved):

PR #1149, opened 2022-06-29
[Major] Stack component review

### LINKED ISSUE #1126: Implement DxcStack component

marcialfps:
Closes #1126


lib/src/stack/Stack.tsx
```diff
@@ -14,60 +20,22 @@ export default function Stack({ gutter, divider, align, as = "div", children }:
...
-        return "0";
-    }
-  }};
-  margin: 0;
-  padding: 0;
+  ${({ alignX, gutter, reverse }) => `
+    flex-direction: ${reverse ? "column-reverse" : "column"};
+    align-items: ${alignX === "start" || alignX === "end" ? `flex-${alignX}` : alignX};
+    gap: ${gutter};
+  `}
```
GomezIvann, line 38: `flex-wrap: wrap;` is remaining. Columns must be wrapped when reached the end its the container.

GomezIvann (dismissed):
Rermaining tests (check Inline stories for more information, they are more less the same):
- Test the alignment (alignX) when the items are wrapped.
- Default wrapped items test.
- Every example has the same number of items, vary this.

lib/src/bleed/Bleed.stories.tsx
```diff
@@ -23,7 +23,7 @@ export const Chromatic = () => (
     </Container>
     <Title title="Space = xxxsmall" theme="light" level={4} />
```
Jialecl, line 24: Update the titles with the new value of the prop

Jialecl (commented):
We should consider how the divider works when the alignX is not set to stretch.

marcialfps:
> We should consider how the divider works when the alignX is not set to stretch.

Good question. The divider is not shown when alignX is different from stretch. I think the width value should always be 100%.

Jialecl (approved):

PR #1150, opened 2022-06-30
Fixed quick nav container width

Lisa159:
Changed width from quick nav container to solve the problem with tables. Now tables can be scrollabled.

Before the change:
image

After the change:
image

### DIFF
--- a/website/screens/common/HeaderCell.tsx
+++ b/website/screens/common/HeaderCell.tsx
@@ -1,7 +1,0 @@
-import styled from "styled-components";
-
-const HeaderCell = styled.th`
-  min-width: 300px;
-`;
-
-export default HeaderCell;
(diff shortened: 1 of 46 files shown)


website/screens/common/QuickNavContainer.tsx
```diff
@@ -77,7 +77,8 @@ const Container = styled.div`
 const ContentContainer = styled.div`
   display: flex;
   flex-direction: column;
-  width: 800px;
+  width: 100%;
+  max-width: 800px;
```
Jialecl, line 81: The fix works fine but the table becomes difficult to read due to the amount of text and small space for the description column.
Since we already have the scroll making the column wider will help with readability.

website/screens/common/HeaderCell.tsx
```diff
@@ -0,0 +1,7 @@
+import styled from "styled-components";
+
+const HeaderCell = styled.th`
```
Jialecl, line 3: Maybe include "Description" in the name to imply a more specialized use/scenario.

Lisa159: Done

Jialecl (approved):

PR #1154, opened 2022-07-01
Inline component documentation

marcialfps:


website/screens/components/inline/code/InlineCodePage.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+const sections = [
+  {
+    title: "Props",
+    content: (
+      <DxcTable>
+        <tr>
+          <th>Name</th>
+          <th>Default</th>
+          <th>Description</th>
```
Lisa159, line 18: Change `th` from Description to `HeaderDescriptionCell`.

marcialfps: Done!

website/screens/components/inline/code/examples/basicUsage.ts
```diff
@@ -0,0 +1,24 @@
+import { DxcInline } from "@dxc-technology/halstack-react";
+import Placeholder from "./Placeholder";
+
+const code = `() => {
+  return (
+    <DxcInline gutter="0.5rem" divider>
```
Lisa159, line 6: What do you think about adding DxcInset to wrap the example?

From this:
image

To this:
image

marcialfps: Yes, it is a good idea!

Lisa159 (approved):

PR #1156, opened 2022-07-05
Fixed examples overflow

Lisa159:
Added overflow to examples container to be able to scroll on them when the screen is small.

image


Jialecl:
it seems that there are still some issues 
image

Jialecl (approved):

PR #1161, opened 2022-07-06
Fixed too long subtitles

Lisa159:
When subtitles are too long, the website can not be seen correctly in small devices:

image

Added `word-break: break-word`  to be more readable:

image


marcialfps (approved):
💯

PR #1168, opened 2022-07-11
Inline component fixes

marcialfps:
### Changes
- Fix gap value when showing the divider.
- Fix divider height when using alignY.
- Changes in story examples.

### DIFF
--- a/lib/src/inline/Inline.stories.tsx
+++ b/lib/src/inline/Inline.stories.tsx
@@ -56,7 +56,7 @@ export const Chromatic = () => (
     </Container>
     <Title title="AlignY = end" theme="light" level={4} />
     <Container>
-      <DxcInline alignY="end">
+      <DxcInline alignY="end" divider>
         <Placeholder height="small" width="small" />
         <Placeholder height="medium" width="medium" />
         <Placeholder height="large" width="large" />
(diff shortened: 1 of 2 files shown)


lib/src/inline/Inline.stories.tsx
```diff
@@ -103,7 +103,7 @@ export const Chromatic = () => (
     </Container>
     <Title title="AlignX = end" theme="light" level={4} />
```
Lisa159, line 104: Add `with divider` in the title.

lib/src/inline/Inline.stories.tsx
```diff
@@ -56,7 +56,7 @@ export const Chromatic = () => (
     </Container>
     <Title title="AlignY = end" theme="light" level={4} />
```
Lisa159, line 57: Add `with divider` in the title.

lib/src/inline/Inline.stories.tsx
marcialfps: Done!

lib/src/inline/Inline.stories.tsx
marcialfps: Done!

Lisa159 (approved):

PR #1183, opened 2022-07-18
[Minor] New prop wrap for Inline component

GomezIvann:
As we discussed, a new prop `wrap` for the Inline component, so the user can decide whether items should be wrapped onto one line or jump to a new one we they can be fitted in a single row.

### DIFF
--- a/lib/src/inline/Inline.stories.tsx
+++ b/lib/src/inline/Inline.stories.tsx
@@ -38,7 +38,7 @@ export const Chromatic = () => (
     </FlexContainer>
     <Title title="Wrap" theme="light" level={4} />
     <Container customWidth>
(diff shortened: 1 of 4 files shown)


lib/src/inline/Inline.stories.tsx
```diff
@@ -38,7 +38,7 @@ export const Chromatic = () => (
     </FlexContainer>
     <Title title="Wrap" theme="light" level={4} />
     <Container customWidth>
-      <DxcInline>
+      <DxcInline wrap>
```
marcialfps, line 41: Add another example without `wrap`.

marcialfps (approved):
💯

PR #1189, opened 2022-07-21
Fixed links width on quick nav

Lisa159:


marcialfps (commented):
Please, add new test cases that cover focus, hover, and long links.

marcialfps (approved):

PR #1199, opened 2022-07-25
Remove material from button component

raquelarrojo:


lib/src/button/Button.tsx
```diff
@@ -31,22 +44,14 @@ const DxcButton = ({
...
-        type={type}
-        margin={margin}
-        mode={mode !== "primary" && mode !== "secondary" && mode !== "text" ? "primary" : mode}
-        disabled={disabled}
-        iconPosition={iconPosition}
-        size={size}
-        backgroundType={backgroundType}
-        icon={icon}
-      >
+      <ButtonContainer margin={margin} size={size} backgroundType={backgroundType}>
```
Jialecl, line 47: I don't think this container is longer necessary since the margin and the size can be applied directly to the button.

lib/src/button/Button.tsx
```diff
@@ -118,43 +85,44 @@ const DxCButton = styled.div`
...
+    box-shadow: 0 0 0 2px
+      ${(props) =>
+        props.backgroundType === "dark" ? props.theme.focusBorderColorOnDark : props.theme.focusBorderColor};
+  }
 
-    ${(props) => {
-      const { mode, backgroundType } = props;
-      if (mode === "primary") {
-        return `
+  ${(props) => {
```
Jialecl, line 122: What do you think about dividing this conditional into three components that inherit from Button?
I think that will make this a bit more readable.

lib/src/button/Button.tsx
raquelarrojo: Done!

lib/src/button/Button.tsx
raquelarrojo: This it nos possible in Button so I made a refactor to avoid the 3 conditionals depending on the button mode and reduce code.

lib/src/button/Button.tsx
raquelarrojo: Done!

lib/src/button/Button.tsx
```diff
@@ -31,84 +44,39 @@ const DxcButton = ({
...
-  fitContent: "unset",
-};
-
-const calculateWidth = (margin, size) => {
-  if (size === "fillParent") {
-    return `calc(${sizes[size]} - ${getMargin(margin, "left")} - ${getMargin(margin, "right")})`;
-  }
-  return sizes[size];
+type ButtonProps = {
+  mode?: "primary" | "secondary" | "text";
```
Jialecl, line 73: props like mode will always have a value I don´t think it should be optional.

raquelarrojo: Done!

Jialecl (approved):

Issue #987, opened 2022-05-24
Wizard focus extends to the full width of the component


GomezIvann:
In vertical mode, the Wizard component extends the step focus to the whole width of the component instead of adjusting to step size.

![image](https://user-images.githubusercontent.com/44321109/169994657-2fdc7aa9-8559-4090-a356-1e43bdc224b4.png)

This can also led to unexpected behaviours, since the whole width of the component is clickable and can change the wizard's current step.

PR #1208, opened 2022-08-01
[Patch] Update Wizard focus behaviour

GomezIvann:
Fixes [987](https://github.com/dxc-technology/halstack-react/issues/987). Also, removed the `@ts-nocheck` flag in the component.

### DIFF
--- a/lib/src/wizard/Wizard.tsx
+++ b/lib/src/wizard/Wizard.tsx
@@ -195,10 +195,7 @@ const Step = styled.button<StepProps>`
   }
 `;
 
-type StepHeaderProps = {
-  validityIcon: boolean;
-};
-const StepHeader = styled.div<StepHeaderProps>`
+const StepHeader = styled.div<{ validityIcon: boolean }>`
   position: relative;
   display: inline-flex;
   ${(props) => props.validityIcon && "padding-bottom: 4px;"}
@@ -348,13 +345,9 @@ const Description = styled.p<StateProps>`
   margin: 0;
 `;
 
-type StepSeparatorProps = {
-  mode: Mode;
-};
-const StepSeparator = styled.div<StepSeparatorProps>`
-  width: ${(props) => (props.mode === "horizontal" ? "" : "0")};
-  height: ${(props) => (props.mode === "horizontal" ? "0" : "")};
-  ${(props) => (props.mode === "vertical" ? "margin: 0 18px;" : "")}
+const StepSeparator = styled.div<{ mode: Mode }>`
+  ${(props) => (props.mode === "horizontal" ? "height: 0;" : "width: 0;")};
+  ${(props) => props.mode === "vertical" && "margin: 0 18px;"}
   border: ${(props) =>
     `${props.theme.separatorBorderStyle} ${props.theme.separatorBorderThickness} ${props.theme.separatorColor}`};
   opacity: 1;
--- a/lib/src/wizard/Wizard.tsx
+++ b/lib/src/wizard/Wizard.tsx
@@ -1,9 +1,8 @@
-// @ts-nocheck
 import React, { useState } from "react";
 import styled, { ThemeProvider } from "styled-components";
 import { spaces } from "../common/variables.js";
 import useTheme from "../useTheme";
-import WizardPropsType from "./types";
+import WizardPropsType, { Margin, Mode, Space } from "./types";
 
 const icons = {
   validIcon: (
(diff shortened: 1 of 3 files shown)


Lisa159:
What do you think about adding a story checking the step focus in vertical mode?

Lisa159:
Focus story with vertical wizard is being cut in the left side: 

image

Lisa159:
Maybe use useState on variable renderedCurrent? (Line 59)
`const renderedCurrent = useState(currentStep == null ? innerCurrent : currentStep;)`

lib/src/wizard/Wizard.tsx
```diff
@@ -180,17 +191,25 @@ const Step = styled.button`
...
-    outline: ${(props) => props.theme.focusColor} auto 1px;
+    outline: 2px solid ${(props) => props.theme.focusColor};
   }
 `;
 
-const StepHeader = styled.div`
+type StepHeaderProps = {
+  validityIcon: boolean;
+};
+const StepHeader = styled.div<StepHeaderProps>`
```
Lisa159, line 201: What do you think about removing the type and write directly `styled.div<{validityIcon: boolean}>`?

lib/src/wizard/Wizard.tsx
```diff
@@ -329,7 +348,10 @@ const Description = styled.p`
   margin: 0;
 `;
 
-const StepSeparator = styled.div`
+type StepSeparatorProps = {
+  mode: Mode;
+};
+const StepSeparator = styled.div<StepSeparatorProps>`
```
Lisa159, line 354: What do you think about removing the type and write directly `styled.div<{mode: Mode}>`?

lib/src/wizard/Wizard.tsx
```diff
@@ -329,7 +348,10 @@ const Description = styled.p`
...
   width: ${(props) => (props.mode === "horizontal" ? "" : "0")};
   height: ${(props) => (props.mode === "horizontal" ? "0" : "")};
```
Lisa159, line 356: We could refactor this two lines (355 and 356) like:
```
  ${(props) =>
    props.mode === "horizontal"
    ? `height: 0;`
    : `width: 0;`}
```

lib/src/wizard/Wizard.tsx
```diff
@@ -1,9 +1,8 @@
-// @ts-nocheck
 import React, { useState } from "react";
 import styled, { ThemeProvider } from "styled-components";
 import { spaces } from "../common/variables.js";
 import useTheme from "../useTheme";
-import WizardPropsType from "./types";
+import WizardPropsType, { Margin, Mode, Space } from "./types";
 
 const icons = {
```
Lisa159, line 7: Maybe export the icons from a different file?

lib/src/wizard/Wizard.tsx
```diff
@@ -146,20 +147,30 @@ const StepsContainer = styled.div`
...
   flex-grow: ${(props) => (props.lastStep ? "0" : "1")};
   flex-direction: ${(props) => (props.mode === "vertical" ? "column" : "row")};
-  ${(props) => (props.mode === "vertical" ? "width: 100%;" : "")}
+  ${(props) => props.mode === "vertical" && "width: fit-content;"}
 `;
 
-const Step = styled.button`
-  border: none;
-  background: inherit;
+type StepProps = {
```
Lisa159, line 162: Missing `disabled` prop.

lib/src/wizard/Wizard.tsx
GomezIvann: Done!

lib/src/wizard/Wizard.tsx
GomezIvann: Done!

lib/src/wizard/Wizard.tsx
GomezIvann: The step is a button tag so it has a [`disabled`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) prop by default from the HTML implementation, it's not necessary to add it again.

lib/src/wizard/Wizard.tsx
GomezIvann: Since there are only two icons I think it is unnecessary to extract them to another file. I only take them out when they are a lot and increase the component's code considerably.

GomezIvann:
> renderedCurrent

useState should only be used when it is a variable that defines the state, and therefore the component. Derived states must use common variables, as it is the case of `renderedCurrent`.

Lisa159 (approved):

PR #1210, opened 2022-08-01
New Flex component

### LINKED ISSUE #1209: Implement Flex component

marcialfps:
### Changes
- New Flex component. Closes #1209 
- Remove DxcStack from List and QuickNav components.
- Remove DxcStack from stories.


lib/src/flex/types.ts
```diff
@@ -0,0 +1,41 @@
...
+    | "flex-start"
+    | "flex-end"
+    | "start"
+    | "end"
+    | "center"
+    | "space-between"
+    | "space-around"
+    | "space-evenly"
+    | "stretch";
+  gap?: string;
```
GomezIvann, line 36: I think the gap's type should be: `string | Gap`, where the Gap type is `{ rowGap: string, columnGap: string }`. What do you think? I know that with the current type string the user can write "20px 10px" and would work fine but I think the new type could orientate him/her better.

GomezIvann (changes requested):
I think we are missing some key props from Flexbox, here is a summary of them:

1. `flex-grow` as `grow`.
2. `flex-basis` as `basis`.
3. `flex-shrink` as `shrink`.
4. `align-self` stays the same.

marcialfps:
> I think we are missing some key props from Flexbox, here is a summary of them:
> 
> 1. `flex-grow` as `grow`.
> 2. `flex-basis` as `basis`.
> 3. `flex-shrink` as `shrink`.
> 4. `align-self` stays the same.

These props are applied to each child, not to the container.

GomezIvann:
> > I think we are missing some key props from Flexbox, here is a summary of them:
> > 
> > 1. `flex-grow` as `grow`.
> > 2. `flex-basis` as `basis`.
> > 3. `flex-shrink` as `shrink`.
> > 4. `align-self` stays the same.
> 
> These props are applied to each child, not to the container.

Flex can be both container and/or contained item. Here is an example with `grow` prop:

```
      <DxcFlex>
        <DxcFlex grow={1}>
          <DxcTextInput label="First name" size="fillParent" />
        </DxcFlex>
        <DxcFlex grow={1}>
          <DxcTextInput label="Second name" size="fillParent" />
        </DxcFlex>
        <DxcFlex grow={4}>
          <DxcTextInput label="Location" size="fillParent" />
        </DxcFlex>
      </DxcFlex>
```

lib/src/flex/types.ts
marcialfps: Yes, it is a good solution ✌🏼

lib/src/flex/Flex.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+        </DxcFlex>
+        <DxcFlex order={5} grow={1}>
+          <PlaceholderGrowAndShrink>order 5, grow 1</PlaceholderGrowAndShrink>
+        </DxcFlex>
+        <DxcFlex order={2} grow={2}>
+          <PlaceholderGrowAndShrink>order 2. grow 2</PlaceholderGrowAndShrink>
+        </DxcFlex>
+      </DxcFlex>
+    </Container>
+    <Title title="Basis and shrink" level={4} />
```
GomezIvann, line 67: This example doesn't show the effect of `flex-shrink`. I suggest changing this example quite a bit:

```
      <DxcFlex basis="500px" gap="20px">
        <DxcFlex grow={1} shrink={2} basis="400px">
          <PlaceholderGrowAndShrink>This item shrinks</PlaceholderGrowAndShrink>
        </DxcFlex>
        <DxcFlex grow={1} basis="400px">
          <PlaceholderGrowAndShrink>Item 2</PlaceholderGrowAndShrink>
        </DxcFlex>
        <DxcFlex grow={1} basis="400px">
          <PlaceholderGrowAndShrink>Item 3</PlaceholderGrowAndShrink>
        </DxcFlex>
      </DxcFlex>
```

image

GomezIvann (changes requested):

lib/src/flex/Flex.stories.tsx
marcialfps: Done

GomezIvann (approved):

PR #1226, opened 2022-08-05
New Flex component documentation

GomezIvann:
Summary:

1. Added documentation of the Flex component to the new documentation site.
2. Flex component refactor.
3. Updated `as` prop type to avoid future problems with typescript.


lib/src/flex/Flex.tsx
```diff
@@ -2,45 +2,12 @@ import React from "react";
...
-      basis={basis}
-      alignSelf={alignSelf}
-      gap={gap}
-      as={as}
-    >
-      {children}
-    </FlexContainer>
-  );
-}
+const DxcFlex = ({ children, ...props }: FlexPropsType): JSX.Element => <Flex {...props}>{children}</Flex>;
```
marcialfps, line 5: You need the default values.

website/screens/components/flex/code/FlexCodePage.tsx
```diff
@@ -0,0 +1,167 @@
...
+import QuickNavContainer from "@/common/QuickNavContainer";
+import QuickNavContainerLayout from "@/common/QuickNavContainerLayout";
+import Example from "@/common/example/Example";
+import directionAlignment from "./examples/directionAlignment";
+import gapOrderGrow from "./examples/gapOrderGrow";
+import HeaderDescriptionCell from "@/common/HeaderDescriptionCell";
+
+const sections = [
+  {
+    title: "Props",
```
marcialfps, line 13: Maybe we can add a link to Flexbox documentation like [this one ](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Basic_Concepts_of_Flexbox).

website/screens/components/flex/code/examples/gapOrderGrow.ts
```diff
@@ -0,0 +1,34 @@
+import { DxcInset, DxcFlex } from "@dxc-technology/halstack-react";
+import Placeholder from "./Placeholder";
+
+const code = `() => {
+  return (
+    <DxcInset space="2rem">
+      <DxcFlex gap="2rem">
+        <DxcFlex order={3} grow={1}>
+          <Placeholder height="large" />
```
marcialfps, line 9: I think we should add `order` value inside each placeholder to make the example more understandable.

marcialfps (changes requested):

lib/src/flex/Flex.tsx
GomezIvann: Agree! I've just restored them (I though we were using the default value of css anyways but in some cases we aren't).

website/screens/components/flex/code/FlexCodePage.tsx
GomezIvann: Also agreed! I talk this with @jsardid but i completely forgot about it.

website/screens/components/flex/code/examples/gapOrderGrow.ts
GomezIvann: Agree, added!

marcialfps (approved):
💯

PR #1232, opened 2022-08-09
Removed Material UI dependencies from Header

Lisa159:


lib/src/header/Header.tsx
```diff
@@ -153,36 +151,30 @@ const DxcHeader = ({
...
+  background-color: ${(props) => props.theme.backgroundColor};
+  border-bottom: ${(props) =>
+    props.underlined &&
+    `${props.theme.underlinedThickness} ${props.theme.underlinedStyle} ${props.theme.underlinedColor}`};
+  min-height: ${(props) => props.theme.minHeight};
+  box-shadow: none;
+  display: flex;
+  align-items: center;
+  flex-direction: row;
+  align-items: center;
```
GomezIvann, line 172: `align-items: center;` is twice, remove one of them.

lib/src/header/Header.tsx
```diff
@@ -153,36 +151,30 @@ const DxcHeader = ({
...
+  justify-content: space-between;
+  box-sizing: border-box;
 `;
 
-const LogoAnchor = styled.a`
+const LogoAnchor = styled.a<{ interactuable: boolean }>`
   ${(props) => {
```
GomezIvann, line 178: This can be reduced to a smaller piece of code as:
` ${(props) => (props.interactuable ? "cursor: pointer" : "cursor: default; outline:none")};`

GomezIvann (changes requested):
Some code improvements I saw along the way:

1. Maybe we should extract the inner component `ContentContainerComponent` and pass him the props as its own props, just like this:

```
type ContentProps = {
  isResponsive: boolean;
  responsiveContent: (closeHandler: () => void) => React.ReactNode;
  handleMenu: () => void;
  padding: Padding;
  content: React.ReactNode;
}
const Content = ({ isResponsive, responsiveContent, handleMenu, padding, content }: ContentProps) => {
  const backgroundType = useContext(BackgroundColorContext);
  return isResponsive? (
    <MenuContent backgroundType={backgroundType}>{responsiveContent(handleMenu)}</MenuContent>
  ) : (
    <ContentContainer padding={padding} backgroundType={backgroundType}>
      {content}
    </ContentContainer>
  );
};
};
```

This should avoid unnecessary re-renders of the content when the whole header re-renders.

2. These brackets are unnecessary, we should remove them:

image

Also, the `<div>` tag is necessary here? 🤔 

4. `ref` variable and its check inside the useEffect is unnecessary:
```
  useEffect(() => {
    handleResize();
    window.addEventListener("resize", handleResize);

    return () => {
      window.removeEventListener("resize", handleResize);
    };
  }, []);
```
Also, the `handleResize`, is a bit confusing, the `!isResponsive` value is causing falsy values when it shouldn't be false. I would refactor this as:
```
  const handleResize = () => {
    setIsResponsive(window.matchMedia(`(max-width: ${responsiveSizes.medium}rem)`).matches);
  };
```

5. `getLogoElement` function can be refactored to be more understandable.
```
const getLogoElement = (themeInput, logoLabel) => {
  if (!themeInput) return dxcLogo;
  else if (typeof themeInput === "string") return <LogoImg alt={logoLabel} src={themeInput}></LogoImg>;
  else return themeInput;
};
```

6. Responsive mode has an **issue**. When you are in that state and open the responsive menu, if you resize to full screen (not responsive size) and then come back to responsive mode, the contextual menu is opened, which shouldn't be like that, it should be closed when it is no longer in the responsive mode so the next time appears closed by default. You can see how this is fixed in the sidenav of the application layout when this situation occurs:

image

lib/src/header/Header.tsx
```diff
@@ -269,7 +267,7 @@ const MainContainer = styled.div`
   flex-grow: 1;
 `;
 
-const ResponsiveMenu = styled.div`
+const ResponsiveMenu = styled.div<{ hasVisibility: boolean }>`
```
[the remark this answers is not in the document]
Lisa159: Done! Thanks!!

lib/src/header/Header.tsx
Lisa159: Done!

lib/src/header/Header.tsx
Lisa159: Done!

lib/src/header/Header.tsx
```diff
@@ -321,7 +327,7 @@ const CloseContainer = styled.div`
   padding: ${spaces.xxsmall};
 `;
 
-const MenuContent = styled.div`
+const MenuContent = styled.div<{ backgroundType: "dark" | "light" }>`
```
GomezIvann, line 330: This container, and the whole layout inside the responsive menu are a bit odd, look:

image

It has a `margin-top` of `40px` to avoid overlapping with the logo, which is quite confusing considering we control that container and it is not from the custom responsive content. I think we should refactor all the custom responsive content using flex.
Maybe the responsive logo and the close action are a row with a content justification of `space-between` and the custom content another row below.

GomezIvann (changes requested):

lib/src/header/Header.tsx
```diff
@@ -120,30 +124,38 @@ const DxcHeader = ({
...
+              <BackgroundColorProvider color={colorsTheme.header.menuBackgroundColor}>
+                <Content
+                  isResponsive={isResponsive}
+                  responsiveContent={responsiveContent}
+                  handleMenu={handleMenu}
+                  padding={padding}
+                  content={content}
+                />
+              </BackgroundColorProvider>
+              <CloseContainer tabIndex={tabIndex} onClick={handleMenu} className="closeIcon">
```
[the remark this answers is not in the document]
Lisa159: Yes, totally agree

lib/src/header/Header.tsx
```diff
@@ -120,30 +124,38 @@ const DxcHeader = ({
         {isResponsive && responsiveContent && (
           <MainContainer>
             <ChildContainer padding={padding}>
-              <HamburguerItem tabIndex={tabIndex} underlined={underlined} onClick={handleMenu}>
+              <HamburguerItem tabIndex={tabIndex} onClick={handleMenu}>
```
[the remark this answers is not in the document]
Lisa159: Yes, I will change it to button.

lib/src/header/Header.tsx
```diff
@@ -97,19 +100,20 @@ const DxcHeader = ({
...
-      handleResize();
-    }
-
+    handleResize();
+    window.addEventListener("resize", handleResize);
     return () => {
       window.removeEventListener("resize", handleResize);
     };
-  }, []);
+  }, [setIsResponsive]);
```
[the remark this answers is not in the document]
Lisa159: Done!

lib/src/header/Header.tsx
```diff
@@ -97,19 +100,20 @@ const DxcHeader = ({
...
+
+  useEffect(() => {
+    !isResponsive && setIsMenuVisible(false);
+  }, [isResponsive, setIsMenuVisible]);
```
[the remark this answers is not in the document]
Lisa159: Done!

lib/src/header/Header.tsx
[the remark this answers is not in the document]
Lisa159: Done!

lib/src/header/Header.tsx
Lisa159: Done without flex.

lib/src/header/Header.tsx
```diff
@@ -120,30 +124,40 @@ const DxcHeader = ({
...
-                    {closeIcon}
-                  </CloseContainer>
-                </ResponsiveMenu>
-                <Overlay onClick={handleMenu} hasVisibility={isMenuVisible}></Overlay>
-              </div>
-            }
+            <ResponsiveMenu hasVisibility={isMenuVisible}>
+              <ResponsiveIconsContainer>
+                <ResponsiveLogoContainer>{headerResponsiveLogo}</ResponsiveLogoContainer>
+                <CloseContainer tabIndex={tabIndex} onClick={handleMenu} aria-label="closeIcon">
```
GomezIvann, line 135: `aria-label` value needs to follow the same approach as in the other components:

1. Translation process: we need to add the string value of the aria-label to the variables.js file, in the `const defaultTranslatedComponentLabels`.
2. The value should follow the pattern "Verb noun". For example, the clear action of a text input has `aria-label="Clear field"` (maybe for this case a value "Close menu", what do you think?).

lib/src/header/Header.tsx
```diff
@@ -120,30 +124,40 @@ const DxcHeader = ({
         {isResponsive && responsiveContent && (
           <MainContainer>
             <ChildContainer padding={padding}>
-              <HamburguerItem tabIndex={tabIndex} underlined={underlined} onClick={handleMenu}>
+              <HamburguerItem tabIndex={tabIndex} onClick={handleMenu}>
                 {hamburgerIcon}
                 <HamburguerTitle>Menu</HamburguerTitle>
```
GomezIvann, line 129: I think we should remove `HamburguerTitle` since it is not necessary and may be semantically wrong (read [this](https://stackoverflow.com/questions/12982269/is-it-semantically-incorrect-to-put-a-div-or-span-inside-of-a-button)). The font styling can go in the above `HamburguerItem`.

GomezIvann: Also, I think `HamburguerTrigger` fits better than `HamburguerItem`, what do you think?

lib/src/header/Header.tsx
```diff
@@ -303,34 +311,35 @@ const ResponsiveMenu = styled.div`
...
-  left: 20px;
+  display: flex;
+`;
+
+const ResponsiveIconsContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
 `;
 
 const CloseContainer = styled.div`
```
GomezIvann, line 322: In the previous review, we had agreed to change the `div` for a `button`.

image

Also, I'm not completely sure on how we are currently styling this with that absolute positioning. How I would do this:

```
const CloseAction = styled.button`
  display: flex;
  flex-wrap: wrap;
  align-content: center;
  padding: 6px;
  height: 36px;
  width: 36px;
  background-color: transparent;
  border: 1px solid transparent;
  border-radius: 2px;

  :focus,
  :focus-visible {
    outline: ${(props) => props.theme.hamburguerFocusColor} auto 1px;
  }
`;
```

GomezIvann (changes requested):

lib/src/header/Header.tsx
Lisa159: Done!

lib/src/header/Header.tsx
Lisa159: I would not use a transparent border for that button because it doesn't have any, only the focus outline.
I can see it better like this:
```
const CloseAction = styled.button`
  cursor: pointer;
  width: 24px;
  height: 24px;
  border: none;
  background-color: transparent;
  padding: 0px;
  :focus,
  :focus-visible {
    outline: ${(props) => props.theme.hamburguerFocusColor} auto 1px;
  }
  svg {
    width: 24px;
    height: 24px;
  }
`;
```

lib/src/header/Header.tsx
Lisa159: Done!

lib/src/header/Header.tsx
GomezIvann: My thoughts about this are in the review!

lib/src/header/Header.tsx
```diff
@@ -303,34 +308,41 @@ const ResponsiveMenu = styled.div`
...
 `;
 
-const CloseContainer = styled.div`
+const ResponsiveIconsContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+`;
+
+const CloseAction = styled.button`
```
GomezIvann, line 320: Some comments about this.

**New close action**

image

**Previous close action**

image

In the new version, we are changing the default styles of the action, since its size was 36x36px and now is 24x24px. We need to remain as faithful as possible to the original.

lib/src/header/Header.tsx
```diff
@@ -120,30 +124,40 @@ const DxcHeader = ({
         {isResponsive && responsiveContent && (
           <MainContainer>
             <ChildContainer padding={padding}>
-              <HamburguerItem tabIndex={tabIndex} underlined={underlined} onClick={handleMenu}>
+              <HamburguerTrigger tabIndex={tabIndex} onClick={handleMenu}>
                 {hamburgerIcon}
-                <HamburguerTitle>Menu</HamburguerTitle>
-              </HamburguerItem>
+                Menu
```
GomezIvann, line 129: We need to also translate this text since it can't be customizable by the user.

lib/src/header/Header.tsx
GomezIvann:
```
const CloseAction = styled.button`
  display: flex;
  flex-wrap: wrap;
  align-content: center;
  padding: 6px;
  height: 36px;
  width: 36px;
  background-color: transparent;
  border: none;

  :focus,
  :focus-visible {
    outline: ${(props) => props.theme.hamburguerFocusColor} auto 1px;
  }
`;
```
We need the width and height attributes to be set to 36px and padding 6px (as the old close action). Display flex attributes for aligning the icon and the SVG selector is not necessary since the default size of the icon is already 24x24px.

GomezIvann (changes requested):

lib/src/header/Header.tsx
Lisa159: Done!

lib/src/header/Header.tsx
Lisa159: Flex attributes doesn't make any changes, I think the problem is in the svg. I will leave it like this until we get a new icon.

GomezIvann (dismissed):

GomezIvann (approved):

PR #1241, opened 2022-08-12
Remove Material UI dependencies from Tabs

Lisa159:


lib/src/tabs/Tab.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+const LabelTextContainer = styled.div``;
+
+const BadgeContainer = styled.div<IconProps>`
+  position: absolute;
+  right: 0;
+  top: ${(props) => (props.hasLabelAndIcon && props.iconPosition === "top" && "0") || "5px"};
+  width: 23px;
+  height: 17px;
+`;
```
raquelarrojo, line 150: Why these odd values `width` and `height`? Are they needed?

lib/src/tabs/Tabs.tsx
```diff
@@ -1,13 +1,43 @@
...
+  useEffect(() => {
+    window.addEventListener("load", handleWindowSizeChange);
+    window.addEventListener("resize", handleWindowSizeChange);
+
+    return () => {
+      window.removeEventListener("load", handleWindowSizeChange);
+      window.removeEventListener("resize", handleWindowSizeChange);
+    };
+  }, [refTabList, handleWindowSizeChange]);
+
```
raquelarrojo, line 38: Are `refTabList` and `handleWindowSizeChange` really needed as dependencies in useEffect?

raquelarrojo (changes requested):
In scrollable tabs specifications scroll icons are always visible. Also, the selected line goes behind the scroll icon and the scroll icon seems to be same color as the text.
image

Here the icon is not visible and the selected line only occupies the tab. The icon is black instead of the same color as in the tab:
image

There is difference between the line of the tab and the one of the scroll icon. In the specs there is no difference between these lines:
image

Related to tabs documentation, there are only tokens for color. We should add typography, Size, Spacing...

lib/src/tabs/Tab.tsx
```diff
@@ -0,0 +1,182 @@
...
+const TabContainer = styled.button<IconProps>`
+  text-transform: ${(props) => props.theme.fontTextTransform};
+  overflow: hidden;
+  position: relative;
+  text-align: center;
+  flex-shrink: 0;
+  line-height: 1.75;
+  white-space: normal;
+  letter-spacing: 0.02857em;
+  border: 0;
```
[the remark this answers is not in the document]
Lisa159: As there is no info about this values in the design, I kept the ones we where already using.

lib/src/tabs/Tabs.tsx
```diff
@@ -1,13 +1,43 @@
...
+const useResize = (refTabList) => {
+  const [viewWidth, setViewWidth] = useState(0);
+
+  const handleWindowSizeChange = useCallback(() => {
+    setViewWidth(refTabList?.current?.offsetWidth);
+  }, [refTabList]);
+
+  useEffect(() => {
+    window.addEventListener("load", handleWindowSizeChange);
+    window.addEventListener("resize", handleWindowSizeChange);
```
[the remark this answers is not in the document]
Lisa159: It is to check if the window is resizing or not. 
We need to know the visible width where you can see the tabs to calculate the scrolling.
By visible width I mean the blue part, it is going to be always different in different devices:
image
image

So with this function, we are taking this width (offsetWidth) from the reference of that component:

```
  const handleWindowSizeChange = useCallback(() => {
    setViewWidth(refTabList?.current?.offsetWidth);
  }, [refTabList]);
```

lib/src/tabs/Tab.tsx
```diff
@@ -0,0 +1,182 @@
...
+  width: 23px;
+  height: 17px;
+`;
+
+const MainLabelContainer = styled.div<{ hasBadge: number | boolean }>`
+  display: flex;
+  flex-direction: row;
+  margin-left: ${(props) => (props.hasBadge && "35px") || "unset"};
+  margin-right: ${(props) => (props.hasBadge && "35px") || "unset"};
+`;
```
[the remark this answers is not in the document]
Lisa159: That´s true, I will take a look.

lib/src/tabs/Tab.tsx
```diff
@@ -0,0 +1,182 @@
...
+        }}
+        onMouseEnter={() => {
+          onMouseEnter();
+        }}
+        onMouseLeave={() => {
+          onMouseLeave();
+        }}
+      >
+        <ParentLabelSpan>
+          <MainLabelContainer hasBadge={tab.notificationNumber}>
```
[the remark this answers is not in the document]
Lisa159: Done, refactored label!

lib/src/tabs/Tab.tsx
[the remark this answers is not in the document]
Lisa159: Using ref to take the width of the badge didn't work as expected because of the re-renders and the selected underline. So I used the calculation: badge width + the margin left (margin-left=12px and width of the badge is 16px or 23px depending on the notification number).

lib/src/tabs/Tab.tsx
Lisa159: Removed values. Don't know why it was decided to use these values, I agree they are odd. They are even defined in the tokens.
image

lib/src/tabs/Tabs.tsx
Lisa159: Removed refTabList dependency, left the other one just in case.

lib/src/tabs/Tab.tsx
```diff
@@ -0,0 +1,181 @@
...
+          hasLabelAndIcon={hasLabelAndIcon}
+          iconPosition={iconPosition}
+        >
+          {tab.icon && (
+            <TabIconContainer hasLabelAndIcon={hasLabelAndIcon} iconPosition={iconPosition}>
+              {typeof tab.icon === "string" ? <TabIcon src={tab.icon} /> : tab.icon}
+            </TabIconContainer>
+          )}
+          <LabelTextContainer>{tab.label}</LabelTextContainer>
+        </MainLabelContainer>
```
raquelarrojo, line 43: I think here we can use `Typography` component for the label and give there the font styles, color, letterSpacing... instead of in the `TabContainer`. We can also remove the `LabelTextContainer` since it is an empty one and we do not need it.

lib/src/tabs/Tab.tsx
```diff
@@ -0,0 +1,181 @@
...
+  font-weight: ${(props) => props.theme.fontWeight};
+  min-width: 90px;
+  max-width: 360px;
+  padding: ${(props) =>
+    ((!props.hasLabelAndIcon || (props.hasLabelAndIcon && props.iconPosition !== "top")) && "12px 16px") || "8px 16px"};
+  height: ${(props) =>
+    ((!props.hasLabelAndIcon || (props.hasLabelAndIcon && props.iconPosition !== "top")) && "47px") || "71px"};
+  min-height: ${(props) =>
+    ((!props.hasLabelAndIcon || (props.hasLabelAndIcon && props.iconPosition !== "top")) && "47px") || "71px"};
+  background-color: ${(props) => props.theme.unselectedBackgroundColor};
```
raquelarrojo, line 86: Height and min height values are `48px` and `72px`, instead of `47px` and `71px` as here.

lib/src/tabs/Tabs.tsx
```diff
@@ -19,102 +49,232 @@ const DxcTabs = ({
...
+    }, 0);
+    setTotalTabsWidth(sumWidth);
+    setActiveIndicatorWidth(refTabs?.current[activeTabIndex ?? innerActiveTabIndex]?.offsetWidth);
+    setActiveIndicatorLeft(refTabs?.current[activeTabIndex ?? innerActiveTabIndex]?.offsetLeft);
+  }, [refTabs]);
+
+  useEffect(() => {
+    setMinHeightTabs(refTabList?.current?.offsetHeight + 1);
+  }, [refTabList]);
+
```
raquelarrojo, line 84: I think we can put `setMinHeightTabs(refTabList?.current?.offsetHeight + 1);` inside the useEffect above.

raquelarrojo (changes requested):

lib/src/tabs/Tab.tsx
Lisa159: Yes I know, this is on purpose. Because the underline is on top of the tabs so if we higher the height, the tabs will be covering the underline.
With height 47px:
image

With height 48px:
image

I have been trying other solutions like changing the z-index of the underline to 2 and height of the tabs to 48px but this happens:
image
The underline will be covering the 1px that belongs to the outline focus of the tab.

However, because the underline is 1px, in the end the whole tabs with the underline will be 48px.
image

lib/src/tabs/Tabs.tsx
Lisa159: Because refTabList and refTabs are different, they should be in different useEffect to prevent the executing of the rest of the code that should not be executed.

lib/src/tabs/Tab.tsx
Lisa159: Done!

raquelarrojo (approved):

PR #1244, opened 2022-08-22
Removing Material from slider

Jialecl:


marcialfps (changes requested):
### Other changes
- I think we should be consistent in how we implement all the components. In your case you are setting the type next to the styled component, I propose you create types for each one naming them as `XXXXPropsType`.

- The disabled tracker changes its styles when clicked.
image

- Add new tests:
  - Test case for all accessibility attributes.
  - Test cases for all [keyboard interactions](https://www.w3.org/WAI/ARIA/apg/patterns/slider/).
  - Test case we saw in the call: `minValue={5} maxValue={50} step={10}`. Verify that the maximum value is 45.

lib/src/slider/Slider.tsx
```diff
@@ -337,6 +414,41 @@ const MaxLabelContainer = styled.span`
...
+  align-items: center;
+  justify-content: center;
+  margin-right: -2px;
+  padding-top: 1px;
+`;
+
+const MarksContainer = styled.div<{ isFirefox: boolean }>`
+  ${(props) => (props.isFirefox ? getFireFoxStyles() : getChromeStyles())}
+  position: absolute;
+  top: calc(50% - 2px);
```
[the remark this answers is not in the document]
Jialecl: I have decided that the best solution is to change the height to 100% and use displayÑflex to align the marks.

lib/src/slider/Slider.tsx
```diff
@@ -138,7 +194,9 @@ const Container = styled.div`
   width: ${(props) => calculateWidth(props.margin, props.size)};
 `;
 
-const Label = styled.label`
+type LabelProps = { disabled: boolean; backgroundType: "dark" | "light" };
```
marcialfps, line 197: Rename to `LabelPropsType`.

lib/src/slider/Slider.tsx
```diff
@@ -164,146 +222,175 @@ const HelperText = styled.span`
...
-  .MuiSlider-container {
-    padding: 30px 24px;
-  }
-  .MuiSlider-root.Mui-disabled {
-    color: ${(props) =>
-      props.backgroundType === "dark"
-        ? props.theme.disabledThumbBackgroundColorOnDark
-        : props.theme.disabledThumbBackgroundColor};
-    cursor: not-allowed;
+const Slider = styled.input<SliderProps>`
```
marcialfps, line 242: Rename to `SliderInput`.

lib/src/slider/Slider.tsx
```diff
@@ -164,146 +222,175 @@ const HelperText = styled.span`
...
   font-style: ${(props) => props.theme.helperTextFontStyle};
   font-weight: ${(props) => props.theme.helperTextFontWeight};
   line-height: ${(props) => props.theme.helperTextLineHeight};
 `;
 
-const SliderContainer = styled.div`
-  display: flex;
-  height: 48px;
-  align-items: center;
+type SliderProps = {
```
marcialfps, line 233: Rename to `SliderInputPropsType`.

lib/src/slider/Slider.tsx
```diff
@@ -317,31 +404,58 @@ const MinLabelContainer = styled.span`
...
+  width: 100%;
+  height: 24px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin-right: -2px;
+  padding-top: 1px;
+`;
+
+type MarksContainerProps = { isFirefox: boolean };
```
marcialfps, line 428: Rename to `MarksContainerPropsType`.

lib/src/slider/Slider.tsx
```diff
@@ -317,31 +404,58 @@ const MinLabelContainer = styled.span`
...
+const MarksContainer = styled.div<MarksContainerProps>`
+  ${(props) => (props.isFirefox ? getFireFoxStyles() : getChromeStyles())}
+  position: absolute;
+  pointer-events: none;
+  height: 100%;
+  display: flex;
+  align-items: center;
+`;
+
+type TickMarkProps = { stepPosition: number; disabled: boolean; backgroundType: "dark" | "light" };
```
marcialfps, line 439: Rename to `TickMarkPropsType`.

marcialfps (changes requested):
- As I said in the previous comment:
  - Add test case for **ALL** accessibility attributes, including: `aria-labelledby`, `aria-orientation`, `aria-valuemax`, `aria-valuemin`, `aria-valuenow`. This should be added to the first test case, not others. If you need an example, take a look to the first test case from Radio Group component.
  - The disabled tracker changes its styles when clicked.
image

lib/src/slider/Slider.test.js
```diff
@@ -54,17 +70,32 @@ describe("Slider component tests", () => {
     });
     expect(getByRole("textbox").hasAttribute("disabled")).toBeTruthy();
     expect(getByRole("textbox").value).toBe("13");
+    expect(getByRole("slider").hasAttribute("disabled")).toBeTruthy();
   });
   test("Calls correct function onDragEnd", () => {
```
marcialfps, line 75: Add `... when it is uncontrolled`.

marcialfps (approved):

PR #1247, opened 2022-08-29
Fix Sidenav poping in responsive mode

GomezIvann:
Sidenav tends to appear a fraction of a second when the new site is loaded in responsive mode since the beginning. A small code refactor to try to load the page with the sidenav collapsed since the first render and avoid its popping.


marcialfps (approved):
💯

Issue #1242, opened 2022-08-16
Select pop over in firefox


Jialecl:
The select pop over does not apply the correct width in firefox.
But the one inside the paginator seems to be working as intended.

![image](https://user-images.githubusercontent.com/12238686/184868008-dc0733f4-89a2-47f8-bdf2-6dafdbdcefbf.png)

GomezIvann:
Can you explain a bit more about how to reach the bug? After playing with the select on Firefox I can't reproduce the problem issue

![image](https://user-images.githubusercontent.com/44321109/186116764-13f17858-ad87-44b9-9aaa-5144fab205ee.png)

Jialecl:
I launched the old app in which we have all the components.
I was trying to reproduce the issue in different projects and that's the one in which I reproduced the issue consistently.

Jialecl:
To reproduce the issue it seems that the select has to be inside a HalstackProvider
`<HalstackProvider>
        <DxcSelect label="testtttt" options={single_options} />
      </HalstackProvider>`
also it is reproducible only the first time the component is rendered in a new tab.

PR #1249, opened 2022-08-31
[Patch] Select fixes

### LINKED ISSUE #1248: Select event `onBlur` is executing when disabled
**Describe the bug**
The onBlur event is currently executing when the component is disabled. It also gains focus, leading to strange behaviour when tabbing between elements from the screen, since it does not jump between enabled items but also disabled selects.

**To Reproduce**

Code example:
```
const () => {
  const options = [
    { label: "Option 01", value: "1" },
    { label: "Option 02", value: "2" },
    { label: "Option 03", value: "3" },
    { label: "Option 04", value: "4" },
  ];
  const onBlur = ({ value, error }) => {
    setValue(value);
    console.log(value);
  };

  return (
    <DxcSelect
      label="Example"
      options={options}
      value={value}
      onBlur={onBlur}
      disabled
    />
  );
};
```

**Expected behaviour**
The onBlur event must not be executed when the component is disabled.

GomezIvann:
Fixes #1248 and #1242. Also, the deprecated [`keycode`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) has been updated to the `key` property for keyboard events in the Select component.


lib/src/select/Select.tsx
```diff
@@ -289,17 +291,10 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
       [handleSelectChangeValue, closeOptions, multiple]
     );
 
-    const handleListboxResize = () => {
+    const getSelectWidth = useCallback(() => {
       const rect = selectRef?.current?.getBoundingClientRect();
-      setListboxStyles({ width: rect?.width });
-    };
-    useEffect(() => {
```
Jialecl, line 296: remove useEffect import since it is no longer used.

GomezIvann: Done!

lib/src/select/Select.test.js
```diff
@@ -293,26 +278,49 @@ describe("Select component tests", () => {
...
+    const { getByRole } = render(
+      <DxcSelect label="test-select-label" options={single_options} disabled onBlur={onBlur} />
+    );
+    const select = getByRole("combobox");
+
+    userEvent.click(select);
+    fireEvent.keyDown(getByRole("combobox"), { key: "Tab", code: "Tab", keyCode: 9, charCode: 9 });
+    expect(onBlur).not.toHaveBeenCalled();
+  });
+  test("Disabled select - When the component gains the focus, the listbox does not open", () => {
```
Jialecl, line 314: The select does not open when focused even when it's not disabled. We should test that the select can not gain the focus when disabled.
Also, the click test is repeated by the test in line 281.

GomezIvann: You are right Jiale! Already change it 👍🏼.

lib/src/select/Select.test.js
```diff
@@ -293,26 +278,45 @@ describe("Select component tests", () => {
     expect(getByText("Option 02, Option 03, Option 04, Option 06")).toBeTruthy();
     expect(submitInput.value).toBe("4,2,6,3");
   });
-  test("Disabled select - Clear all options action must be shown but not clickable", () => {
-    const { getByRole, getByText, queryByRole } = render(
-      <DxcSelect label="test-select-label" value={["1", "2"]} options={single_options} disabled searchable multiple />
+  test("Disabled select - Listbox cannot be opened by click or keyboard press", () => {
```
Jialecl, line 281: Change the name of the test since there is no longer any keboard press.

GomezIvann: Done!

Jialecl (approved):

PR #1255, opened 2022-09-01
removing material from checkbox

Jialecl:


lib/src/checkbox/Checkbox.tsx
```diff
@@ -67,47 +76,46 @@ const DxcCheckbox = ({
...
-        disabled={disabled}
-        margin={margin}
-        size={size}
-        backgroundType={backgroundType}
-        isLabelHovered={isLabelHovered}
-      >
+      <CheckboxContainer id={name} disabled={disabled} margin={margin} size={size}>
         {label && labelPosition === "before" && labelComponent}
-        <Checkbox
+        <CheckboxInput
```
GomezIvann, line 89: Some comments about this checkbox input. I think we should approach this refactor as the Select, let me explain. The custom checkbox is CheckboxFocus because we do not want to use the native one. So that is what the user sees and it should have the `role="checkbox" name={name} aria-labelledby={labelId} aria-checked={checked ?? innerChecked}`. The input should be only used for the submit event and must be hidden from the user, both visually and in the DOM. 

```
<ValueInput
  name={name}
  value={multiple ? (value ?? innerValue).join(",") : value ?? innerValue}
  readOnly
  aria-hidden="true"
/>
const ValueInput = styled.input`
  display: none;
`;
```

Look the Select's implementation for more information.

GomezIvann (changes requested):
Summary of things I saw along the way

1. Disabled components should not gain the focus:

image

2. The space between label and checkbox should mark as selected the checkbox when clickable. The cursor changes correctly to pointer, but when clicked, doesn't change checkbox's state.
Sin título

3. I think we can take advantage of this refactor and add some more jests tests. For example: 
   - "Checkbox renders with correct text" is no longer necessary since we have visual tests.
   - Check keyboard accessibility.
   - Check disabled state (and avoid bugs such as point 1).
   - Check correct values for aria attributes and role.
   - Check that the value sent in `onSubmit` event is correct.
   - (Any more tests you can think of)

lib/src/checkbox/Checkbox.tsx
```diff
@@ -67,47 +76,46 @@ const DxcCheckbox = ({
...
-          disableRipple
-          className="test"
-          tabIndex={tabIndex}
+          tabIndex={-1}
         />
-        <CheckboxBlackBack
-          labelPosition={labelPosition}
-          disabled={disabled}
-          checked={checked ?? innerChecked}
+        <CheckboxFocus
```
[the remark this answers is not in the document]
Jialecl: This container can not be removed.
If it is removed the checkbox position will not be in the correct position when the size is not fitContent.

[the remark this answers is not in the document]
GomezIvann: This has been resolved in the last commit!

GomezIvann (commented):
More comments regarding the component:

1. The gap between the label and the checkbox doesn't change the state (hover and active) of the checkbox input.
2. Doble click on the label selects the text.

lib/src/checkbox/Checkbox.tsx
```diff
@@ -1,14 +1,18 @@
...
 import { getMargin } from "../common/utils.js";
 import { v4 as uuidv4 } from "uuid";
 import useTheme from "../useTheme";
 import useTranslatedLabels from "../useTranslatedLabels";
 import BackgroundColorContext from "../BackgroundColorContext";
-import CheckboxPropsType from "./types";
+import CheckboxPropsType, { Margin, Space } from "./types";
+
+const checkedIcon = (
+  <svg fill="currentColor" focusable="false" aria-hidden="true" viewBox="0 0 24 24" data-testid="CheckBoxIcon">
```
GomezIvann, line 12: Remove `data-testid="CheckBoxIcon"` from the icon.

lib/src/checkbox/Checkbox.tsx
```diff
@@ -189,106 +218,85 @@ const CheckboxContainer = styled.span`
...
+          : getNotDisabledColor(props, "check")
+        : "transparent"};
+    color: ${(props) => (props.disabled ? getDisabledColor(props, "background") : getNotDisabledColor(props, "hoverBackground"))};
+  }
+  &:focus-visible {
+    outline: 2px solid
+      ${(props) => (props.backgroundType === "dark" ? props.theme.focusColorOnDark : props.theme.focusColor)};
+    outline-offset: 2px;
+  }
+  svg {
```
GomezIvann, line 295: Fix the svg size. Currently, the icon is overflowing due to its size of 18x18px, which, following the specs, should be 16x16px.

MicrosoftTeams-image

image

GomezIvann (changes requested):

lib/src/checkbox/Checkbox.tsx
```diff
@@ -160,26 +163,95 @@ const getNotDisabledColor = (props, element) => {
...
+  background-color: ${(props) =>
+    props.checked
+      ? props.disabled
+        ? getDisabledColor(props, "check")
+        : getNotDisabledColor(props, "check")
+      : "transparent"};
+  color: ${(props) =>
+    props.disabled ? getDisabledColor(props, "background") : getNotDisabledColor(props, "background")};
+
+  &:focus-visible {
```
GomezIvann, line 233: Should we update this to `:focus, :focus-visible` selector? 🤔

GomezIvann (changes requested):
We missed some functionality:

- With the current implementation, our label checks the checkbox input but doesn't give it the focus. We must implement that, which is how it would work if we had to use the native checkbox input with a label and `htmlFor` property.
  - [ ] Add `onClick` event to the label and give the focus to the checkbox (`ref.current.focus();`)
- Then, we must guarantee that the container, which has the "change value" functionality, doesn't steal the focus from the checkbox. This can be achieved by adding an `onMouseDown` to the main container with `event.preventDefault()`.
- And finally, we need to give the focus to the checkbox when he clicks in the gap. This should be as simple as adding `ref.current.focus();` to the change value handler.
- With the previous three points, the `handleLabelDblClick` handler should not be necessary.
- Optionally, if you want to avoid unnecessary calls to the focus() method, you can check if the checkbox already has the focus, like this: `document.activeElement !== ref?.current`. This should be a problem but may be more precise.

lib/src/checkbox/Checkbox.tsx
```diff
@@ -24,92 +27,92 @@ const DxcCheckbox = ({
...
+    if (event.detail > 1) {
+      event.preventDefault();
+    }
+  };
 
   return (
     <ThemeProvider theme={colorsTheme.checkbox}>
-      <CheckboxContainer
+      <MainContainer
         id={name}
```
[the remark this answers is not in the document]
Jialecl: Yes I agree with you this is remaining from the old implementation I will remove it.

Jialecl:
> * Then, we must guarantee that the container, which has the "change value" functionality, doesn't steal the focus from the checkbox. This can be achieved by adding an `onMouseDown` to the main container with `event.preventDefault()`.

In regards to this I find it unnecessary, a div without tabIndex=0 is never focusable.
Also the onClick function already gives the focus to the checkbox.

lib/src/checkbox/Checkbox.tsx
Jialecl: I think we should change it to :focus to make it consistent with our other components.

GomezIvann:
> > * Then, we must guarantee that the container, which has the "change value" functionality, doesn't steal the focus from the checkbox. This can be achieved by adding an `onMouseDown` to the main container with `event.preventDefault()`.
> 
> In regards to this I find it unnecessary, a div without tabIndex=0 is never focusable. Also removes the functionality to select the label.

Totally agree.

lib/src/checkbox/Checkbox.tsx
```diff
@@ -24,92 +27,82 @@ const DxcCheckbox = ({
...
-      if (typeof onChange === "function") {
-        onChange(isChecked);
-      }
-    } else {
-      if (typeof onChange === "function") {
-        onChange(!checked);
-      }
+  const handleCheckboxChange = () => {
+    if (document.activeElement !== checkboxRef?.current) {
+      checkboxRef?.current.focus();
```
GomezIvann, line 41: We need to also check the nullish value `current`: `checkboxRef?.current?.focus();`

lib/src/checkbox/Checkbox.test.js
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import { render, fireEvent } from "@testing-library/react";
 
 import DxcCheckbox from "./Checkbox";
+import userEvent from "@testing-library/user-event";
 
 describe("Checkbox component tests", () => {
   test("Checkbox renders with correct text", () => {
```
GomezIvann, line 8: Remove this test, is unnecessary (it's tested in the stories).

GomezIvann (changes requested):
One final appointment:

- We need to also test `aria-labelledby` and `aria-required` properties.

GomezIvann (approved):

GomezIvann (approved):

PR #1261, opened 2022-09-07
[Patch] Progressbar is determinated or not depending on if it has val…

### LINKED ISSUE #1215: Progress Bar is being indeterminate or determiante depending on prop showValue
**Describe the bug**
Prop `showValue` is making determinate or indeterminate the progress bar instead of the prop `value` as we defined in the documentation.

**To Reproduce**
This should be determinate but it is not:
```
<DxcInset space="2rem">
  <DxcProgressBar label="Loading" value={80} />
</DxcInset>
```

**Screenshots**
image
image
image

jsuarezgonz:
…ue or not (Closes #1215)

### DIFF
--- a/lib/src/progress-bar/ProgressBar.tsx
+++ b/lib/src/progress-bar/ProgressBar.tsx
@@ -47,7 +47,7 @@ const DxcProgressBar = ({
           >
             <LinearProgressBar
               backgroundType={backgroundType}
-              variant={showValue ? "determinate" : "indeterminate"}
+              variant={value === null || value === undefined ? "indeterminate" : "determinate"}
               container="first"
               value={valueProgressBar}
             ></LinearProgressBar>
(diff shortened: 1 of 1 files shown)


lib/src/progress-bar/ProgressBar.tsx
```diff
@@ -47,7 +47,7 @@ const DxcProgressBar = ({
           >
             <LinearProgressBar
               backgroundType={backgroundType}
-              variant={showValue ? "determinate" : "indeterminate"}
+              variant={value === null || value === undefined ? "indeterminate" : "determinate"}
               container="first"
               value={valueProgressBar}
             ></LinearProgressBar>
```
raquelarrojo, line 53: I think with this new change we should remove the following piece of code:
```
{!showValue && (
  <LinearProgressBar
      backgroundType={backgroundType}
      variant="indeterminate"
      container="second"
      value={valueProgressBar}
  ></LinearProgressBar>
)}
```

raquelarrojo (changes requested):

raquelarrojo (approved):

PR #1265, opened 2022-09-14
[Patch] fix Slider to accept zero

Jialecl:
Removing "&& value" from condition to accept value 0 in slider.

Created from #1264

### DIFF
--- a/lib/src/slider/Slider.test.js
+++ b/lib/src/slider/Slider.test.js
@@ -1,7 +1,7 @@
 import React from "react";
 import { render, fireEvent, act } from "@testing-library/react";
-import DxcSlider from "./Slider";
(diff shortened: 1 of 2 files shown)


lib/src/slider/Slider.test.js
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import { render, fireEvent, act } from "@testing-library/react";
 import DxcSlider from "./Slider";
+import userEvent from "@testing-library/user-event";
```
raquelarrojo, line 5: Please put this import before the `DxcSlider` import to avoid the warning:

image

raquelarrojo (changes requested):

raquelarrojo (approved):

PR #1267, opened 2022-09-15
Remove Material UI dependencies from Accordion and Accordion Group

GomezIvann:
Summary of changes:

1. The HTML structure of the component has been completely overwritten to match the current [W3C design pattern for accordions](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/). Now the component has two main parts: the Accordion Header (an `h3` tag with a `button` inside) and the Accordion Panel (regular `div` for the custom content).
2. Accordion Group has also been updated, removing the Material selectors and replacing them for regular CSS.
3. Added accessibility attributes to the component, following the [aria specifications of W3C](https://www.w3.org/WAI/ARIA/apg/example-index/accordion/accordion.html).
4. Updated stories and Jest testing.


lib/src/accordion/Accordion.tsx
```diff
@@ -24,54 +32,73 @@ const DxcAccordion = ({
...
-                </IconContainer>
-              )}
-              <AccordionLabel>{label}</AccordionLabel>
+              <AccordionLabel>
+                {icon && (
+                  <IconContainer disabled={disabled}>
+                    {typeof icon === "string" ? <IconImg src={icon} /> : icon}
+                  </IconContainer>
+                )}
+                {label}
```
Lisa159, line 64: We could use here DxcTypography for the label.

lib/src/accordion/Accordion.tsx
```diff
@@ -1,15 +1,23 @@
...
-import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
 import { getMargin } from "../common/utils.js";
 import { spaces } from "../common/variables.js";
 import useTheme from "../useTheme";
 import { BackgroundColorProvider } from "../BackgroundColorContext";
-import AccordionPropsType from "./types";
+import AccordionPropsType, { Margin, Padding, Space } from "./types";
+import DxcFlex from "../flex/Flex";
+
+const expand_less = (
```
Lisa159, line 10: What do you think about changing this variable name to camel case?

lib/src/accordion/Accordion.tsx
```diff
@@ -83,162 +110,112 @@ const DXCAccordion = styled.div`
...
-    padding-top: ${(props) =>
-      props.padding && typeof props.padding === "object" && props.padding.top ? spaces[props.padding.top] : ""};
-    padding-right: ${(props) =>
-      props.padding && typeof props.padding === "object" && props.padding.right ? spaces[props.padding.right] : ""};
-    padding-bottom: ${(props) =>
-      props.padding && typeof props.padding === "object" && props.padding.bottom ? spaces[props.padding.bottom] : ""};
-    padding-left: ${(props) =>
-      props.padding && typeof props.padding === "object" && props.padding.left ? spaces[props.padding.left] : ""};
+  :active:enabled {
+    background-color: ${(props) => `${props.theme.hoverBackgroundColor}`};
```
Lisa159, line 143: As in the guidelines, active state should also have outline.

Lisa159:
What do you think about adding hover, active and focus stories for accordion group?

lib/src/accordion/Accordion.tsx
```diff
@@ -24,54 +32,73 @@ const DxcAccordion = ({
...
-              <AccordionLabel>{label}</AccordionLabel>
+              <AccordionLabel>
+                {icon && (
+                  <IconContainer disabled={disabled}>
+                    {typeof icon === "string" ? <IconImg src={icon} /> : icon}
+                  </IconContainer>
+                )}
+                {label}
+              </AccordionLabel>
+              {assistiveText && <AccordionAssistiveText disabled={disabled}>{assistiveText}</AccordionAssistiveText>}
```
Lisa159, line 66: We could also use DxcTypography for the assistiveText.

lib/src/accordion-group/AccordionGroup.tsx
```diff
@@ -1,11 +1,10 @@
...
 import React, { createContext, useCallback, useContext, useMemo, useState } from "react";
 import styled, { ThemeProvider } from "styled-components";
 import DxcAccordion from "../accordion/Accordion";
 import { getMargin } from "../common/utils.js";
 import { spaces } from "../common/variables.js";
 import useTheme from "../useTheme";
-import AccordionGroupPropsType, { AccordionPropsType } from "./types";
+import AccordionGroupPropsType, { AccordionPropsType, Margin, Space } from "./types";
 
 type AccordionGroupAccordionContext = {
```
Lisa159, line 9: Maybe adding AccordionGroupAccordionContext type inside types.ts?

lib/src/accordion/Accordion.tsx
GomezIvann: Agree! I don't know why I used "_" to be fair.

lib/src/accordion/Accordion.tsx
GomezIvann: This wouldn't make any difference apart from a storie level. The` :active` selector works from the moment the component has been clicked to the mouse release, which coincides with the focus-gaining moment and, therefore, the `:focus` is activated.

GomezIvann: You can try this by long pressing the component. The design guidelines where trying the mouse-pressed state, which has the focus, as our component.

lib/src/accordion-group/AccordionGroup.tsx
GomezIvann: Agree, done!

lib/src/accordion/Accordion.tsx
GomezIvann: Done!

lib/src/accordion/Accordion.tsx
GomezIvann: Done!

lib/src/accordion-group/AccordionGroup.tsx
```diff
@@ -73,79 +66,47 @@ DxcAccordionGroup.Accordion = AccordionGroupAccordion;
 
 const calculateWidth = (margin) => `calc(100% - ${getMargin(margin, "left")} - ${getMargin(margin, "right")})`;
 
-const AccordionGroupContainer = styled.div`
-  display: inline-block;
+const AccordionGroupContainer = styled.div<{ margin: Space | Margin; disabled: boolean }>`
```
Lisa159, line 69: I think separating the types in a variable would be cleaner.

lib/src/accordion/Accordion.tsx
```diff
@@ -24,54 +32,100 @@ const DxcAccordion = ({
...
+        )}
+      </AccordionContainer>
     </ThemeProvider>
   );
 };
 
 const calculateWidth = (margin) => `calc(100% - ${getMargin(margin, "left")} - ${getMargin(margin, "right")})`;
 
-const DXCAccordion = styled.div`
+const AccordionContainer = styled.div<{ isExpanded: boolean; margin: Space | Margin; padding: Space | Padding }>`
```
Lisa159, line 121: I think separating the types in a variable would be cleaner.

lib/src/accordion-group/AccordionGroup.tsx
GomezIvann: Done 😉

lib/src/accordion/Accordion.tsx
GomezIvann: Done 😉

Lisa159 (approved):

PR #1275, opened 2022-09-19
Working with any minor from React 18

marcialfps:
Before we had React 18.2.0 version, this forces the users to have this exact version as peer-dependency. Now, we allow the users to have any minor version that is part of React 18.

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -13,8 +13,8 @@
   "main": "./main.js",
   "types": "./main.d.ts",
   "peerDependencies": {
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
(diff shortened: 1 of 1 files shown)


marcialfps:
> I would rather enforce they use at least a certain version or a later one.. major or minor. Something like
> 
> ```
> "react": ">=16.0.0",
> ```

I think this is dangerous since major versions can have breaking changes in our library.

kukoc:
If you want to enforce >= version 18, that's fine. However, from now on, I would test every new version of REact and, if breaking changes doesn't affect the DS components, then I would stay enforcing >=18

kukoc (approved):

PR #1295, opened 2022-09-22
Website deployment

marcialfps:
### Changes
- Added assetPrefix and basePath to NextJS config.
- Change images loader when building for production.
- Use of SITE_VERSION variable in scripts.


website/next.config.js
```diff
@@ -31,4 +33,6 @@ module.exports = {
       "/": { page: "/overview/introduction" },
     };
   },
+  assetPrefix: isProd ? `/halstack/${process.env.SITE_VERSION}` : undefined,
```
kukoc, line 36: How these env variables are set? I can¡t find the place where we are assigning values to them

marcialfps: This environment variable is set by default by NextJS.

kukoc (approved):

PR #1300, opened 2022-09-22
Website improvements

marcialfps:
### Changes
- Added version indicator to Sidenav header.
image

- Added _Documentation of other releases_ section to the introduction.


website/screens/overview/introduction/IntroductionPage.tsx
```diff
@@ -185,6 +185,47 @@ const sections = [
       </>
     ),
   },
+  {
+    title: "Documentation of other releases",
+    content: (
```
raquelarrojo, line 190: As it is not so oriented to releases, maybe it would be better a more general title like `Previous documentation` or something like this.

website/screens/overview/introduction/IntroductionPage.tsx
```diff
@@ -185,6 +185,47 @@ const sections = [
...
+          </em>
+        </DxcParagraph>
+        <DxcParagraph>
+          The last development site can be accessed through this{" "}
+          <DxcLink href="https://developer.dxc.com/halstack/next/" newWindow>
+            link
+          </DxcLink>
+          ; and the last release is available{" "}
+          <DxcLink href="https://developer.dxc.com/halstack/6/" newWindow>
+            here
```
raquelarrojo, line 222: There is no way to take the last release programatically instead of change it manually?

raquelarrojo (changes requested):

website/screens/overview/introduction/IntroductionPage.tsx
```diff
@@ -185,6 +185,47 @@ const sections = [
...
+          </DxcBulletedList.Item>
+        </DxcBulletedList>
+        <DxcParagraph>
+          <em>
+            Please note that these sites are discontinued and may not be updated
+            anymore.
+          </em>
+        </DxcParagraph>
+        <DxcParagraph>
+          The last development site can be accessed through this{" "}
```
GomezIvann, line 216: Rephrase both "last" appearances to "latest". _Last_ is used for something that happens at the end, when we are referring to the most recent version of something we should use _latest_.

GomezIvann (changes requested):

website/screens/overview/introduction/IntroductionPage.tsx
marcialfps: We have links to next and current releases... Maybe we can call it `Other documentations`?

website/screens/overview/introduction/IntroductionPage.tsx
marcialfps: Good idea! I think we can take it from the env variable.

marcialfps: Done!

website/screens/overview/introduction/IntroductionPage.tsx
raquelarrojo: Done!

website/screens/overview/introduction/IntroductionPage.tsx
raquelarrojo: Done!

raquelarrojo (approved):

GomezIvann (dismissed):

GomezIvann (approved):

PR #1301, opened 2022-09-22
Reverting useId to uuidv4

Jialecl:


eriesgo:
@Jialecl  for curiosity, why are we reverting to uuidv4?

Jialecl:
> @Jialecl for curiosity, why are we reverting to uuidv4?

Using useId forces the user into upgrading to react 18.
We think that for now it's still not worth forcing the user into upgrading to react 18.

GomezIvann (approved):

PR #1303, opened 2022-09-22
Update component's size fitContent

Lisa159:


Lisa159:
**Update: **
Breaking change in the box component, looks like it wasn't working properly.

image

Jialecl:
Should we consider wrapping or ellipsis for the dropdown, the tag and the slider?

When the content is really large they can affect the width of the container or the window, while other components jump to the next line or have ellipsis.

Lisa159:
Opened discussion about its behaviour: [Behavior for components with fitContent](https://github.com/dxc-technology/halstack-react/discussions/1304)

Jialecl (approved):

PR #1309, opened 2022-09-27
Refactor Switch

Lisa159:
Refactored some details:
- Input has now display: none and fixed aria labels.
- Removed animations.
- Removed variable hasLabel.
- We can change the state by clicking in the gap between label and switch.
- Fixed tests for the changes of the input.


GomezIvann (changes requested):

lib/src/switch/Switch.tsx
```diff
@@ -25,7 +25,7 @@ const DxcSwitch = ({
   const [switchId] = useState(`switch-${uuidv4()}`);
   const labelId = `label-${switchId}`;
   const [innerChecked, setInnerChecked] = useState(defaultChecked ?? false);
-  const [hasLabel] = useState((label !== "" && label !== null && label !== undefined) ?? false);
+  const hasLabel = (label !== "" && label !== null && label !== undefined) ?? false;
```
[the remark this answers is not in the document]
Lisa159: Done

Lisa159:
Because the outline can overlap with other components, I added a margin of 12px to the sides of the switch. Chromatic tests are going to fail because of this.

lib/src/switch/Switch.tsx
```diff
@@ -63,61 +62,64 @@ const DxcSwitch = ({
...
           </LabelContainer>
         )}
-        <SwitchBase
-          labelPosition={labelPosition}
-          hasLabel={hasLabel}
-          htmlFor={labelId}
-          onClick={disabled === true ? () => {} : handlerSwitchChange}
-        >
+        <SwitchBase htmlFor={labelId}>
           <SwitchInput
```
GomezIvann, line 84: Can we rename this styled-component as `ValueInput` to match the other implementations (Checkbox, Select, etc)?

lib/src/switch/Switch.tsx
```diff
@@ -63,61 +62,64 @@ const DxcSwitch = ({
...
             {label} {optional && <>{translatedLabels.formFields.optionalLabel}</>}
           </LabelContainer>
         )}
-        <SwitchBase
-          labelPosition={labelPosition}
-          hasLabel={hasLabel}
-          htmlFor={labelId}
-          onClick={disabled === true ? () => {} : handlerSwitchChange}
-        >
+        <SwitchBase htmlFor={labelId}>
```
GomezIvann, line 83: `htmlFor={labelId}` is unnecessary. When the labelled element is inside the <label> they are already associated, so they are accessible.

GomezIvann (changes requested):
Some comments:

- I see a lot this piece of code: 
```
if (typeof onChange === "function") {
   onChange(isChecked);
}
```
We now do not have the need of checking the type since typescript does that for us. We can refactor this code by simply: `onChange?.(isChecked);`. What do you think?

- In the onChange handler `handlerSwitchChange` we use the event.target.value, which can only be true or false, why do we check it with ?? operator? Also, the expression `checked === undefined` only check `undefined` value, but not `null` (which is very common). My suggest here is:
```
const handlerSwitchChange = (event) => {
   checked ?? setInnerChecked(event.target.checked);
   onChange?.(checked ?? event.target.checked);
};
```

Lisa159:
> 

- Removed type checking for onChange function.
- About the `handlerSwitchChange`, we need to check `innerChecked` for the uncontrolled. With your solution, uncontrolled doesn't work.

lib/src/switch/Switch.tsx
Lisa159: Done

lib/src/switch/Switch.tsx
Lisa159: What do you mean by inside? `ValueInput` and `LabelContainer` are siblings in the DOM.

GomezIvann:
> > 
> 
> * Removed type checking for onChange function.
> * About the `handlerSwitchChange`, we need to check `innerChecked` for the uncontrolled. With your solution, uncontrolled doesn't work. Removed `event.target.checked` because it was always `undefined`.

Yes, you are right. My solution was for a handler in a `<input type="checkbox" />`, and in this case, it is on a `<div>`, which of course doesn't have a checked property.

lib/src/switch/Switch.tsx
GomezIvann: `htmlFor` is used to match an input with its label/output. First, in this case, the `SwitchBase` is a container, so it doesn't need a `for` attribute. The relation is between <LabelContainer> and <DisabledSwitchTrack/SwitchTrack>, which is properly established by using the `aria-labelledBy`.

What I meant with inside is that the SwitchBase is a <label> and is associated automatically with its inner content (look at [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#:~:text=To%20associate%20the%20with,same%20as%20the%20input's%20id%20)).

Lisa159: Removed.

lib/src/switch/Switch.tsx
```diff
@@ -51,73 +48,59 @@ const DxcSwitch = ({
...
         )}
-        <SwitchBase
-          labelPosition={labelPosition}
-          hasLabel={hasLabel}
-          htmlFor={labelId}
-          onClick={disabled === true ? () => {} : handlerSwitchChange}
-        >
-          <SwitchInput
+        <SwitchBase>
+          <ValueInput
```
GomezIvann, line 78: We need to move `ValueInput` outside the `SwitchBase` container to avoid double executing the `onChange` event. It must be a direct child of `SwitchContainer`.

GomezIvann: Also, it misses the `checked` property. Without it, the `value` will always be sent in the submit event.

lib/src/switch/Switch.tsx
```diff
@@ -51,73 +48,59 @@ const DxcSwitch = ({
     if (checked === undefined) {
```
GomezIvann, line 48: This code can be stretched again. Proposal:
```
  const handlerSwitchChange = () => {
    checked ?? setInnerChecked((innerChecked) => !innerChecked);
    onChange?.(checked ? !checked : !innerChecked);
  };
```

lib/src/switch/Switch.tsx
```diff
@@ -51,73 +48,59 @@ const DxcSwitch = ({
...
             name={name}
-            id={labelId}
-            disabled={disabled}
+            aria-hidden={true}
             value={value}
-            aria-labelledby={labelId}
-            aria-label={hasLabel ? label : undefined}
+            disabled={disabled}
+            readOnly
+          ></ValueInput>
```
GomezIvann, line 85: Remove the closing tag `</ValueInput>.` When a component doesn't have children the common code recommendation is to close it within the opening tag (`<Component />`)

GomezIvann (changes requested):

lib/src/switch/Switch.tsx
Lisa159: Done

lib/src/switch/Switch.tsx
Lisa159: Done

lib/src/switch/Switch.tsx
```diff
@@ -40,84 +39,56 @@ const DxcSwitch = ({
...
-            />
-          ) : (
-            <SwitchTrack
-              backgroundType={backgroundType}
-              data-checked={checked ?? (innerChecked ? innerChecked : undefined)}
-              tabIndex={tabIndex}
-              ref={refTrack}
-            />
-          )}
+            tabIndex={tabIndex ? tabIndex : -1}
```
GomezIvann, line 81: I don't understand very well this piece of code: `tabIndex={tabIndex ? tabIndex : -1}`. Wouldn't it be `tabIndex={disabled ? -1 : tabIndex}`? Currently the component is not focusable by keyboard.

GomezIvann (changes requested):

lib/src/switch/Switch.tsx
Lisa159: Done

GomezIvann (approved):

PR #1313, opened 2022-09-28
Website version fixes

marcialfps:
**Checklist**

**Purpose**
The introduction page on the website has a link to the current release documentation, but now it is not working fine in "next". Also, the tag that indicates the version of the documentation is not working as expected.

**Description**
- Now we obtain the current release version from `versions.json` and append it to the link.
- Version tag is obtained from `basePath`.

### DIFF
--- a/website/screens/common/sidenav/SidenavLogo.tsx
+++ b/website/screens/common/sidenav/SidenavLogo.tsx
@@ -6,8 +6,8 @@ import React from "react";
 import { useRouter } from "next/router";
 
 const SidenavLogo = () => {
-  const router = useRouter();
-  const siteVersion = router.basePath.split("/")[2];
+  const { basePath } = useRouter();
+  const siteVersion = basePath.split("/")[2];
 
   return (
     <SidenavLogoContainer>
@@ -24,7 +24,7 @@ const SidenavLogo = () => {
         <Subtitle>Design system</Subtitle>
       </LogoContainer>
       <StatusTag>
-        {router.basePath
+        {basePath
           ? isNaN(parseInt(siteVersion))
             ? siteVersion
             : `v${siteVersion}.0.0`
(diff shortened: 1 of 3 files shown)


website/screens/common/sidenav/SidenavLogo.tsx
```diff
@@ -3,14 +3,12 @@ import Image from "@/common/Image";
...
-  ? typeof process.env.SITE_VERSION === "number"
-    ? `v${process.env.SITE_VERSION}.0.0`
-    : process.env.SITE_VERSION
-  : "dev";
+import { useRouter } from "next/router";
 
 const SidenavLogo = () => {
+  const router = useRouter();
+  const siteVersion = router.basePath.split("/")[2];
+
```
raquelarrojo, line 11: This works with this URL: https://developer.dxc.com/halstack/next/, but does it also work with a URL like this: https://developer.dxc.com/halstack/next/components/application-layout/?

Also, as router is only used for getting the basePath I recommend to obtain the basePath like this, just to follow a good practice:
```
const { basePath } = useRouter();
const siteVersion = basePath.split("/")[2];
```

raquelarrojo (changes requested):

website/screens/common/sidenav/SidenavLogo.tsx
marcialfps: > This works with this URL: https://developer.dxc.com/halstack/next/, but does it also work with a URL like this: https://developer.dxc.com/halstack/next/components/application-layout/?

`basePath` has the value from the config file `next.config.js`. In that path, we always have something like this:  `/halstack/SITE_VERSION`. Take a look at the [documentation](https://nextjs.org/docs/api-reference/next.config.js/basepath).

marcialfps: Done requested change! ✌🏼

raquelarrojo (approved):

PR #1317, opened 2022-09-30
Fix routing introduction page

### LINKED ISSUE #1310: Sidenav's selected link appears duplicated
**Describe the bug**
In the deployed site, when you access to throw the most common URL and go to a page different from _Introduction_, the selected link appears duplicated.

**To Reproduce**

1. Access to our new documentation site: 
2. Jump from the Introduction to another page, for example, the documentation of the Accordion.

**Expected behaviour**
The selected link from the sidenav must be unique.

**Screenshots**
image

Lisa159:
**Checklist**

**Description**
The page Introduction is linked to `/` and `/overview/introduction`, so when entering in the website through https://developer.dxc.com/halstack/next/, it is going to load the page perfectly. However, to check the active page in the sidenav, we are checking the current path of the website:

image

and using this array with types `LinkDetails` which has two strings for the matching (label and path): 

image

So, first option to fix this problem would imply changing the code (this type `LinkDetails` ) to add an array of strings in the path like `["/", "/overview/introduction"]`. I discarded this option because the code is already well written and we should do this with routing.

Second option, using `next.config.js` to use `redirects()` function as we have right now. It is not working after build because we are using `next export` and looks like that `next export` doesn't support rewrites or redirects.

Third option, [middleware](https://nextjs.org/docs/advanced-features/middleware). From Next documentation: `Middleware allows you to run code before a request is completed, then based on the incoming request, you can modify the response by rewriting, redirecting, adding headers, or setting cookies.` For the problem we have, this would be the perfect solution. So I check how to implement it. We are using Next version 12.1.0 and looks like middleware is in beta, however we can use it. To do it so, we need to create a file `_middleware.ts` in pages folder with the following code: 
```
import { NextResponse, NextRequest } from 'next/server'
export async function middleware(req, ev) {
    const { pathname } = req.nextUrl
    if (pathname == '/') {
        return NextResponse.redirect('/overview/introduction')
    }
    return NextResponse.next()
}
```

Besides, middleware is now stable in Next version 12.2.0 but implementation is different. We should create a file `middleware.ts` in the root folder with the code:
```
import { NextResponse } from "next/server";
import type { NextRequest } from "next/server";

export function middleware(request: NextRequest) {
  const url = request.nextUrl.clone();
  url.pathname = "/overview/introduction";
  return NextResponse.redirect(url);
}

export const config = {
  matcher: ["/"],
};
```

**Second and third options are working fine locally. But as happens with `next.config.ts`, `next export` doesn't support middleware: https://nextjs.org/docs/advanced-features/static-html-export#unsupported-features**

Last option, using the hook useRouter. This option is the only one after the others that works with `next export`. It consist on using useRouter and useEffect to redirect when page Introduction has current path as `/`.

image

Info: https://stackoverflow.com/questions/58173809/next-js-redirect-from-to-another-page

**Closes #1310**

### DIFF
--- a/website/pages/overview/introduction.tsx
+++ b/website/pages/overview/introduction.tsx
@@ -1,7 +1,16 @@
 import Head from "next/head";
+import { useRouter } from "next/router";
+import { useEffect } from "react";
 import IntroductionPage from "../../screens/overview/introduction/IntroductionPage";
 
 const Introduction = () => {
+  const router = useRouter();
+
+  useEffect(() => {
+    const { asPath } = router;
+    asPath === "/" && router.push("/overview/introduction");
+  }, [router]);
+
   return (
     <>
       <Head>


marcialfps:
Maybe we can do something like `alternative version 2` from [here](https://github.com/vercel/next.js/discussions/17322).

We can create `index.tsx` file that receives all requests to `/`, then redirects to the introduction page. This solution is related to your last option, but in this case, we avoid the point of adding more complexity to the `Introduction.tsx` file.

Lisa159:
Solution with `index.tsx`:
For this to work, we need to create an `index.tsx` in the folder `pages` with the following code:
```
import { useRouter } from "next/router";
import { useEffect } from "react";

export default function Index() {
  const router = useRouter();

  useEffect(() => {
    const { asPath } = router;
    asPath === "/" && router.push("/overview/introduction");
  }, []);

  return <></>;
}
```

And removed this code from `next.config.js`: 
```
  exportPathMap: async function (
    defaultPathMap,
    { dev, dir, outDir, distDir, buildId }
  ) {
    return {
      ...defaultPathMap,
      "/": { page: "/overview/introduction" },
    };
  },
```

PR #1319, opened 2022-10-03
Radio Group focus behaviour update

GomezIvann:
**Checklist**

**Purpose**
Match the focus behaviour of the Radio Group to the accessibility standards and the acquired knowledge.

**Description**
Summary of changes:

- Update Radio Group's focus to a more logical and accessible behaviour.
- Update `"Space"` key code to `" "` code (read #1324).
- Remove `the "Enter"` key since no other DS implements it and only WAI-ARIA uses it (as an optional extra key, not mandatory).
- Hidden `ValueInput` has been replaced by a more semantic option: `<input type="hidden" />`.
- Minor code improvements.

**Additional context**
- [HTML checkbox with label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio).
- [Focusable areas in HTML Standard](https://html.spec.whatwg.org/multipage/interaction.html#data-model).
- [MDN tabindex docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex#accessibility_concerns) and [HTML Standard definition](https://html.spec.whatwg.org/multipage/interaction.html#the-tabindex-attribute).
- [How to use tabindex properly](https://www.tpgi.com/using-the-tabindex-attribute/).
- [MDN `<input type="hidden" />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio).


Jialecl:
I think we should add a test to chromatic to cover the optional radio group.

Jialecl:
Change the test.js import order to avoid warnings.

Jialecl (approved):

PR #1329, opened 2022-10-07
Fix Select specifications page

GomezIvann:
**Checklist**

**Purpose**
Match the common criteria of how our specification pages should look like.

**Description**
The numbering of the titles of the Select specifications page has been fixed in order to match the rest of the components.

### DIFF
--- a/website/screens/components/select/specs/SelectSpecsPage.tsx
+++ b/website/screens/components/select/specs/SelectSpecsPage.tsx
@@ -1068,9 +1068,6 @@ const sections = [
         title: "Margin",
         content: (
           <>
-            <DxcParagraph>
-              Select <Code>margin</Code> can use the values:
-            </DxcParagraph>
             <DxcTable>
               <thead>
                 <tr>
@@ -1126,8 +1123,7 @@ const sections = [
             <DxcParagraph>
               These values can be applied independently to each side of the
               component: <Code>top</Code>, <Code>bottom</Code>,{" "}
-              <Code>left</Code> and
-              <Code>right</Code>.
+              <Code>left</Code> and <Code>right</Code>.
             </DxcParagraph>
           </>
         ),
--- a/website/screens/components/select/specs/SelectSpecsPage.tsx
+++ b/website/screens/components/select/specs/SelectSpecsPage.tsx
@@ -1021,47 +1021,42 @@ const sections = [
         ),
       },
       {
-        title: "Size",
+        title: "Width",
         content: (
-          <>
-            <DxcParagraph>
-              The component <Code>width</Code> can adopt the following values:
-            </DxcParagraph>
-            <DxcTable>
-              <thead>
-                <tr>
-                  <th>Width</th>
-                  <th>Value</th>
-                </tr>
-              </thead>
-              <tbody>
-                <tr>
-                  <td>
-                    <Code>small</Code>
-                  </td>
-                  <td>240px</td>
-                </tr>
-                <tr>
-                  <td>
-                    <Code>medium</Code>
-                  </td>
-                  <td>360px</td>
-                </tr>
-                <tr>
-                  <td>
-                    <Code>large</Code>
-                  </td>
-                  <td>480px</td>
-                </tr>
-                <tr>
-                  <td>
-                    <Code>fillParent</Code>
-                  </td>
-                  <td>100%</td>
-                </tr>
-              </tbody>
-            </DxcTable>
-          </>
+          <DxcTable>
+            <thead>
+              <tr>
+                <th>Width</th>
+                <th>Value</th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td>
+                  <Code>small</Code>
+                </td>
+                <td>240px</td>
+              </tr>
+              <tr>
+                <td>
+                  <Code>medium</Code>
+                </td>
+                <td>360px</td>
+              </tr>
+              <tr>
+                <td>
+                  <Code>large</Code>
+                </td>
+                <td>480px</td>
+              </tr>
+              <tr>
+                <td>
+                  <Code>fillParent</Code>
+                </td>
+                <td>100%</td>
+              </tr>
+            </tbody>
+          </DxcTable>
         ),
       },
       {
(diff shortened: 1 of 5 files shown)


website/screens/components/select/specs/SelectSpecsPage.tsx
```diff
@@ -143,1056 +350,842 @@ const sections = [
...
+                  </td>
+                  <td>100px</td>
+                </tr>
+              </tbody>
+            </DxcTable>
+            <DxcParagraph>
+              These values can be applied independently to each side of the
+              component: <Code>top</Code>, <Code>bottom</Code>,{" "}
+              <Code>left</Code> and
+              <Code>right</Code>.
```
raquelarrojo, line 1130: Please add a blank space between `and` and `right`.

website/screens/components/select/specs/SelectSpecsPage.tsx
```diff
@@ -143,1056 +350,842 @@ const sections = [
...
+                  <td>
+                    <Code>small</Code>
+                  </td>
+                  <td>240px</td>
+                </tr>
+                <tr>
+                  <td>
+                    <Code>medium</Code> (<em>default</em>)
+                  </td>
+                  <td>360px</td>
```
raquelarrojo, line 1048: The same here for `default`. I would not add it here, to keep it as the rest of components. Also I think the place of this default explanation is the code (in the props table, as it is now).

website/screens/components/select/specs/SelectSpecsPage.tsx
```diff
@@ -143,1056 +350,842 @@ const sections = [
...
+          </>
+        ),
+      },
+      {
+        title: "Margin",
+        content: (
+          <>
+            <DxcParagraph>
+              Select <Code>margin</Code> can use the values:
+            </DxcParagraph>
```
raquelarrojo, line 1073: Same here for the `margin` explanation.

website/screens/components/select/specs/SelectSpecsPage.tsx
```diff
@@ -143,1056 +350,842 @@ const sections = [
...
+          </>
+        ),
+      },
+      {
+        title: "Size",
+        content: (
+          <>
+            <DxcParagraph>
+              The component <Code>width</Code> can adopt the following values:
+            </DxcParagraph>
```
raquelarrojo, line 1029: I would not add this explanation to follow the rest of components documentation. I think it is clear without it.

raquelarrojo (changes requested):

website/screens/components/select/specs/SelectSpecsPage.tsx
GomezIvann: Done!

website/screens/components/select/specs/SelectSpecsPage.tsx
GomezIvann: Maybe this was done by Alex to let the user know that the Specifications examples have a `"medium"` size.

GomezIvann: Regarding the blank space, there is already one, I don't fully understand this comment 🤔.

website/screens/components/select/specs/SelectSpecsPage.tsx
GomezIvann: I'm not so sure here. I'm going to remove it but the Margin section seems to not be very consistent through the components...

website/screens/components/select/specs/SelectSpecsPage.tsx
GomezIvann: I think is interesting what Alex did. Since the section is called `Size` the user cannot know if this affects the width or the height. In other components, this becomes trivial since the section is called `Width`.

website/screens/components/select/specs/SelectSpecsPage.tsx
raquelarrojo: I've only seen this `default` explanation in `Select` and `Text Input`. In the rest of components,  it is not specified the default. I think we should keep it like this also in Select and Text input, since in any other specifications prop the default one is explained, and it makes sense to me. The default values should only be explained in the props table in code tab.

Which blank space do you refer here? I've only commented about the blank space in the comment above:
image

website/screens/components/select/specs/SelectSpecsPage.tsx
raquelarrojo: As you can see in the specs of all the components, in any of them there is this explanation, only the explanation related to how the margin is applied. I'm fine with its removal.

Box:
image

Card:
image

Checkbox:
image

Dropdown:
image

Header:
image

Link:
image

Progress Bar:
image

Radio Froup:
image

Switch:
image
Text Input:
image

Toggle Group:
image

Wizard:
image

website/screens/components/select/specs/SelectSpecsPage.tsx
raquelarrojo: After reviewing the specs of ll the components, the guidelines related to the size are explained like this:

If the section is for the `width` and `height`, the title is `Size`. If it is only the `width`, the title is called `Width`. So, for the Select component I would change the title to `Width`, remove the with explanation and the `(default)`, to keep it as the rest of the components specs.

Could you please remove also the `(default)` from Text Input? It is the only one apart from the one you added in Select.

website/screens/components/select/specs/SelectSpecsPage.tsx
GomezIvann: It is not only present in the Select and Text Input, take a look:

**Dialog**:
image

**Dropdown**:
image

Shall we also remove it in these places?

raquelarrojo: Yes, I would remove the `(default)` from `Dialog` and `Dropdown` too, since it is not present in `Margin`, `Padding`, and `Width` sections of any of the rest of the components.

website/screens/components/select/specs/SelectSpecsPage.tsx
GomezIvann: Agree with the Width section decision. About the `(default)`, I have already removed every appearance, so I think this is an outdated comment!

raquelarrojo (approved):

PR #1338, opened 2022-10-19
Fixed navigation using repeated sublinks in QuickNav

### LINKED ISSUE #1323: Navigation using repeated sublinks not working in QuickNav
**Describe the bug**
Navigation using sublinks is not working fine if the sublink text is the same as other sublink of other section in the QuickNav component.

**To Reproduce**
Repeate sublinks like here:
image

**Expected behavior**
Sublinks should navigate to the correct section although there are several sublinks with same texts.

Lisa159:
**Checklist**

**Purpose**
Navigation using sublinks wasn't working fine if the sublink text is the same as other sublink of other section in the QuickNav component.

**Description**
Added parent label to sublinks `href`. So for example, if parent label is `overview` and sublink is `introduction`, navigation for sublink will be `overview-introduction`.

Besides, updated stories with working links and sublinks. Removed unused `id` prop from array.

**Screenshots**
image

image

**Closes #1323**

### DIFF
--- a/lib/src/quick-nav/QuickNav.stories.tsx
+++ b/lib/src/quick-nav/QuickNav.stories.tsx
@@ -11,6 +11,36 @@ export default {
   component: DxcQuickNav,
 };
 
+const defaultLinks = [
+  {
+    label: "Overview",
+    id: "overview",
+  },
+  {
+    label: "Principles",
+    id: "principles",
+    links: [
+      { label: "Color", id: "color" },
+      { label: "Spacing", id: "spacing" },
+      { label: "Typography", id: "typography" },
+    ],
+  },
+  {
+    label: "Components",
+    id: "components",
+    links: [
+      {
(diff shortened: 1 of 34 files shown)


lib/src/quick-nav/QuickNav.stories.tsx
```diff
@@ -44,30 +44,36 @@ const defaultLinks = [
 const links = [
```
raquelarrojo, line 44: Please remove `id` from `defaultLinks` also (line 14).

raquelarrojo (changes requested):

lib/src/quick-nav/QuickNav.stories.tsx
Lisa159: Done!

raquelarrojo (approved):

PR #1344, opened 2022-10-21
Using transient props in Flex component

### LINKED ISSUE #1343: Wrong attributes appearing in Flex HTML outcome
**Describe the bug**
While launching the [W3C HTML validator](https://validator.w3.org/) on our site I saw a consistent error throw our pages regarding our Flex component. Its styled component seems to be adding some of the props to the outcome DOM as if they were HTML attributes and that should be happening since they are CSS properties.

**Expected behaviour**
`direction="column"` should only be visible in our internal code and not appear in the output DOM tree.

**Screenshots**

image

Error

GomezIvann:
**Checklist**
- [ ] Build process is done without errors and all tests pass in the `/lib` directory.

**Description**
To avoid passing some specific props that could be potentially HTML attributes, styled-components offer what they called **transient props**. These are regular props marked with the symbol `$` so the library understands that they shouldn't be passed to the underlying component.

**Screenshots**
image

**Additional context**
[What transient props are](https://styled-components.com/docs/api#transient-props)
https://medium.com/@probablyup/introducing-transient-props-f35fd5203e0c
https://dev.to/sarahscode/props-are-not-forever-preventing-props-from-being-passed-to-the-dom-with-styled-components-v5-1-l47

**Closes #1343 **

### DIFF
--- a/lib/src/flex/Flex.tsx
+++ b/lib/src/flex/Flex.tsx
@@ -1,35 +1,52 @@
 import React from "react";
 import styled from "styled-components";
-import FlexPropsType from "./types";
+import FlexPropsType, { StyledProps } from "./types";
 
-const DxcFlex = ({ children, ...props }: FlexPropsType): JSX.Element => <Flex {...props}>{children}</Flex>;
+const DxcFlex = ({
+  direction = "row",
+  wrap = "nowrap",
+  gap = "0",
+  order = 0,
+  grow = 0,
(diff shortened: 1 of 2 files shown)


GomezIvann:
Also, I fixed an small error catched during validation in the Text Input styles which was rendering `false;` when hovered:

image

image

Lisa159 (approved):

PR #1366, opened 2022-11-03
Fix Checkbox tick with sticky positioning

### LINKED ISSUE #1355: Checkbox position in scrollable multiselect
When options are checked in a select multiple dialog with overflowing content, scrolling down causes the checkbox "check" to behave as sticky.

Image for reference:

image

GomezIvann:
**Checklist**

**Description**
To solve this issue, I just had to add `position: relative` to the checkbox parent container, so the `position: absolute` of the tick is calculated based on the actual positioning of the component as a whole. When an element is has position absolute, it is positioned relative to its closest positioned ancestor, if any; otherwise, it is placed relative to the initial [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block). 

**Screenshots**

image

image

**Closes #1355**

### DIFF
--- a/lib/src/checkbox/Checkbox.stories.tsx
+++ b/lib/src/checkbox/Checkbox.stories.tsx
@@ -4,7 +4,7 @@ import { BackgroundColorProvider } from "../BackgroundColorContext";
 import Title from "../../.storybook/components/Title";
 import ExampleContainer from "../../.storybook/components/ExampleContainer";
 import DarkContainer from "../../.storybook/components/DarkSection";
-import { userEvent } from "@storybook/testing-library";
+import styled from "styled-components";
 
 export default {
   title: "Checkbox",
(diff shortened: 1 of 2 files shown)


lib/src/checkbox/Checkbox.stories.tsx
```diff
@@ -84,44 +68,16 @@ const Checkbox = () => (
...
-        <ExampleContainer>
-          <Title title="Checked with label after" theme="dark" level={4} />
-          <DxcCheckbox label="Checkbox" defaultChecked labelPosition="after" />
-        </ExampleContainer>
-        <ExampleContainer>
-          <Title title="Optional with label after" theme="dark" level={4} />
-          <DxcCheckbox label="Checkbox" optional labelPosition="after" />
-        </ExampleContainer>
-        <ExampleContainer>
-          <Title title="Disabled and checked with label after" theme="dark" level={4} />
```
Jialecl, line 115: I would not remove some scenarios like the disabled optional.

GomezIvann: Why should we test that? `(Optional)` text is just been added as plain text inside the `<Label>` component. It would be the same as testing the disabled with a different string each time.

Jialecl: It's not about testing the string but making sure that even disabled, the `(Optional)` string is still added with the correct styles.

GomezIvann: The presence of the optional label doesn't depend on whether the component is disabled or not. And, as I said, is just a string inside the same component as the `label` prop, tested one, tested all. I can't see the value here Jiale.

GomezIvann: I still do not agree, we should not program for the future (in this kind of situation) since we do not know if the component will evolve and if a dependency will appear here. Anyways, I add the test to unlock this PR.

Jialecl (approved):

PR #1367, opened 2022-11-03
Removing Material-ui datepicker

### LINKED ISSUE #1368: Remove Material UI dependencies from Date Input

Jialecl:
**Checklist**

**Purpose**
Material-ui datepicker removed and added month picker to the new calendar

**Additional context**
Dayjs has been used to format and help to calculate the correct date within the calendar while navigating.

**Closes #1368 #855**


lib/src/date-input/DateInput.tsx
```diff
@@ -1,18 +1,17 @@
 // @ts-nocheck
```
Lisa159, line 1: I think you can remove this @ts-nocheck

lib/src/date-input/DateInput.tsx
```diff
@@ -1,18 +1,17 @@
...
 import React, { useState, useRef } from "react";
-import { DatePicker, MuiPickersUtilsProvider } from "@material-ui/pickers";
-import { createMuiTheme, MuiThemeProvider, Paper } from "@material-ui/core";
-import ClickAwayListener from "@material-ui/core/ClickAwayListener";
-import Popover from "@material-ui/core/Popover";
 import dayjs from "dayjs";
-import DayjsUtils from "@date-io/dayjs";
 import styled, { ThemeProvider } from "styled-components";
```
Lisa159, line 4: `styled` is not used. You can remove it.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,488 @@
...
+    </DatePicker>
+  );
+};
+
+function getYearsArray() {
+  const yearList = [];
+  for (let i = firstYear; i <= lastYear; i++) {
+    yearList.push(i);
+  }
+  return yearList;
```
marcialfps, line 236: Are you creating an array of years **each time** you open the year picker? I think this is not the best approach. 

Maybe you can create a stateful variable that depends on `date` prop and create the array from that value. Also, I think we can remove `firstYear` and `lastYear` variables, and calculate the years by a maximum and a minimum number of years to show.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,488 @@
...
+      }
+    }
+    function handleKeyboardEvent(event) {
+      switch (event.key) {
+        case "Escape":
+          event.preventDefault();
+          onCloseCalendar();
+          break;
+      }
+    }
```
marcialfps, line 42: I think we should be consistent with the style of our code:

```
const handleClickOutside = (event) => { ... }
const handleKeyboardEvent = (event) => { ... }
```

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,488 @@
...
+
+  const ref = useRef(null);
+
+  useEffect(() => {
+    function handleClickOutside(event) {
+      if (ref.current && !ref.current.contains(event.target)) {
+        onCloseCalendar();
+      }
+    }
+    function handleKeyboardEvent(event) {
```
marcialfps, line 35: We can move this function outside the `useEffect`.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,488 @@
...
+            </PickerHeaderButton>
+            <HeaderMonthYear aria-live="polite">{innerDate.format("MMMM YYYY")}</HeaderMonthYear>
+            <PickerHeaderButton
+              aria-label="next month"
+              onClick={() => setInnerDate(innerDate.set("month", innerDate.get("month") + 1))}
+            >
+              <svg fill="currentColor" focusable="false" viewBox="0 0 24 24" aria-hidden="true">
+                <path d="M8.59 16.59L13.17 12 8.59 7.41 10 6l6 6-6 6-1.41-1.41z"></path>
+                <path fill="none" d="M0 0h24v24H0V0z"></path>
+              </svg>
```
marcialfps, line 163: We should move outside this `svg` code to make our code more "readable".

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,488 @@
...
+              <DayHeaderCell key={weekDay}>{weekDay}</DayHeaderCell>
+            ))}
+          </WeekHeader>
+
+          <DayCellsContainer>
+            {dayCells.map((day) =>
+              day !== 0 ? (
+                <DayCell
+                  onKeyDown={(event) => handleDayKeyboardEvent(event, day)}
+                  id={"day_" + day}
```
marcialfps, line 177: Change to `day_${day}`

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,488 @@
...
+            <YearPickerButton
+              aria-label={year}
+              selected={selectedDate.get("year") === year}
+              autoFocus={selectedDate.get("year") === year}
+              onClick={() => {
+                handleDateSelect(year, "year");
+                setContent("calendar");
+              }}
+            >
+              <span>{year}</span>
```
marcialfps, line 205: Maybe we can move this to a styled component.

lib/src/date-input/Calendar.tsx
Jialecl: I agree that calculating the years by a maximum and a minimum number of years to show is a better approach, but we would need to discuss more this behaviour, right now the year selector works the same way as the one before.

In regards of creating the array each time I agree with you I will rethink this to avoid it.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,495 @@
+import dayjs, { UnitType, Dayjs } from "dayjs";
+import React, { useState, useMemo, useEffect, useRef } from "react";
+import styled from "styled-components";
+import { CalendarPropsType } from "./types";
+
+const weekDays = ["Mo", "Tu", "We", "Th", "Fr", "Sa", "Su"];
+const months = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
+
+const firstYear = 1899;
```
GomezIvann, line 9: Why are you creating variables when both `firstYear` and `lastYear` are only used once inside the loop? 🤔

GomezIvann: If you do it for a better understanding of the code, it makes more sense to have them inside the function that uses them.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,495 @@
...
+          onClick={() => setContent((content) => (content === "monthPicker" ? "calendar" : "monthPicker"))}
+        >
+          {selectedDate.format("MMMM")}
+        </DatePickerToolbarSubtitleButton>
+      </DatePickerToolbar>
+      {content === "calendar" && (
+        <Calendar>
+          <PickerHeader>
+            <PickerHeaderButton
+              aria-label="previous month"
```
GomezIvann, line 168: Icon buttons (buttons without text) need to a have `title` defined for accessibility standards. It should have the same string as the `aria-label`. Also, these inner texts need to be part of our translation strategy, look at the Clear action of the Select for a better understanding.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,495 @@
...
+              onClick={() => setInnerDate(innerDate.set("month", innerDate.get("month") - 1))}
+            >
+              {leftCaret}
+            </PickerHeaderButton>
+            <HeaderMonthYear aria-live="polite">{innerDate.format("MMMM YYYY")}</HeaderMonthYear>
+            <PickerHeaderButton
```
GomezIvann, line 174: Same as for the previous month button.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,495 @@
...
+            >
+              {rightCaret}
+            </PickerHeaderButton>
+          </PickerHeader>
+          <WeekHeader>
+            {weekDays.map((weekDay) => (
+              <DayHeaderCell key={weekDay}>{weekDay}</DayHeaderCell>
+            ))}
+          </WeekHeader>
+
```
GomezIvann, line 186: Remove this empty space.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,495 @@
...
+            </MonthPickerButton>
+          ))}
+        </MonthPicker>
+      )}
+    </DatePicker>
+  );
+};
+
+const DatePicker = styled.div`
+  box-shadow: 0px 5px 5px -3px rgb(0 0 0 / 20%), 0px 8px 10px 1px rgb(0 0 0 / 14%), 0px 3px 14px 2px rgb(0 0 0 / 12%);
```
GomezIvann, line 247: I think we should use the `box-shadow` of the other components with a dialog associated, such as the Dropdown or the Select, for consistency. Maybe Raffy can give as and advice.

lib/src/date-input/DateInput.tsx
```diff
@@ -115,209 +101,42 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
...
+              tabIndex={tabIndex}
+              ref={refDate}
+            />
+          </Popover.Trigger>
+          <Popover.Content align="end" aria-modal={true}>
+            <DxcDateInputCalendar
+              onCloseCalendar={closeCalendar}
+              onDateSelect={handleCalendarOnClick}
+              date={value || innerValue ? getValueForPicker(value ?? innerValue, format.toUpperCase()) : dayjs()}
+            ></DxcDateInputCalendar>
```
GomezIvann, line 134: Components without children don't need a closing tag: `<Component {...props} />`.

GomezIvann (changes requested):
More comments:

- [ ] Date picker seems to overlap the text input. 

image

- [ ] Each day's font-size is smaller than in the previous date picker.
- [ ] Focus behaviour is not working as expected. The `Tab` should move the focus throw the focusable elements inside the calendar dialog. These components are:
   - Year button.
   - Month button.
   - Previous month button.
   - Next month button.
   - **Only** the selected day. Right now, when you tab, the focus goes throw every day of the month, which is a bit tedious and not correct. Only the selected (active) day should be focusable. See [this](https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-datepicker.html) or any other calendar.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,495 @@
...
+        break;
+      case " ":
+        event.preventDefault();
+        handleDateSelect(day, "date");
+        break;
+    }
+  };
+
+  return (
+    <DatePicker id="date-picker" ref={ref}>
```
Lisa159, line 151: What is this `id` for?

lib/src/date-input/Calendar.tsx
GomezIvann: I think what Jiale did here is right. Since the function of `getYearsArray` is defined and executed outside of the main component, it is only calculated once, which is the correct approach!

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,495 @@
...
+  color: ${(props) => props.theme.dateInput.pickerActualDateFontColor};
+  font-family: ${(props) => props.theme.dateInput.pickerFontFamily};
+  background: ${(props) => props.theme.dateInput.pickerBackgroundColor};
+  font-size: 2rem;
+  border: none;
+  cursor: pointer;
+  border-radius: 4px;
+  font-weight: 400;
+  line-height: 1.75;
+  letter-spacing: 0.00938em;
```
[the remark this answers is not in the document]
Jialecl: These are the values been used already and they would still be strange values converted to px.

lib/src/date-input/Calendar.tsx
Jialecl: After consulting with Raffy he agrees that for consistency we should have the same `box-shadow`

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,495 @@
...
+    document.addEventListener("mousedown", handleClickOutside);
+    document.addEventListener("keydown", handleKeyboardEvent);
+    return () => {
+      document.removeEventListener("mousedown", handleClickOutside);
+      document.removeEventListener("keydown", handleKeyboardEvent);
+    };
+  }, [onCloseCalendar]);
+
+  useEffect(() => {
+    document.getElementById(`day_` + dateToFocus.get("date"))?.focus();
```
marcialfps, line 71: Change to ``day_${dateToFocus.get("date")}``.

marcialfps (changes requested):

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,495 @@
+import dayjs, { UnitType, Dayjs } from "dayjs";
+import React, { useState, useMemo, useEffect, useRef } from "react";
+import styled from "styled-components";
+import { CalendarPropsType } from "./types";
+import useTranslatedLabels from "../useTranslatedLabels";
+
+const getYearsArray = () => {
+  const yearList = [];
+  for (let i = 1899; i <= 2100; i++) {
```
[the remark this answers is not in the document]
Jialecl: I agree that I prefer that behavior but that changes how the component works. Do we want to change it or add props to this component in this PR?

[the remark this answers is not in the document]
GomezIvann: I agree that we may need to calculate the limits based on the current year ([1899, currentYear+100]), but I don't the that calc should be repeated each time the user changes the selected year.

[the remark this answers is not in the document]
marcialfps: We can cover this functionality in the next iteration 👍🏼

lib/src/date-input/DateInput.stories.tsx
```diff
@@ -118,8 +130,9 @@ const YearPicker = () => (
 );
 
 export const ShowYearPicker = YearPicker.bind({});
```
Lisa159, line 132: Change the title of the story to show the behaviour that the component should have.

lib/src/date-input/DateInput.tsx
```diff
@@ -115,209 +101,42 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
       ),
```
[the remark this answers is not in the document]
Jialecl: I will move it outside the component for now. But yes I agree that we should consider having all the icons in the same place.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,518 @@
...
+        break;
+      case " ":
+        event.preventDefault();
+        handleDateSelect(day, "date");
+        break;
+    }
+  };
+
+  return (
+    <DatePicker ref={ref}>
```
GomezIvann, line 169: I think the JSX code of this component is huge. What about extracting some parts of it into separate components just to make the code cleaner? After reviewing the implementation I see three different modules based on `content`'s value:

1. Month Picker
2. Year Picker
3. Calendar
4. Date Picker

Date Picker would be the main parent component which would have all the logic, and the others would be representational components containing the different possible views of the picker and receive as props all the necessary values.

GomezIvann: When a component increases its side, it's recommended to separate it into smaller and more [pure components](https://beta.reactjs.org/learn/describing-the-ui#keeping-components-pure).

GomezIvann (changes requested):
Summary:

- [x] The date picker is been pushed by the error message.

image

- [x] Depending on the month and its string length, the right arrow (next month) changes its positioning.

The focus is not always working as expected:
- [x] If you open the calendar and then click the date input, the date picker does not close.
- [x] If you close the date picker with the `Esc` key, the focus doesn't go back to the date input, as it was in the previous one or the one from W3C.
- [x] If you open the calendar, then click on a space (for example, the blank spaces on the right side of the selected month/year) and press any arrow button, the dialogue gains the focus:

image

lib/src/common/variables.js
```diff
@@ -1519,6 +1524,8 @@ export const defaultTranslatedComponentLabels = {
       "December",
     ],
     monthsShort: ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"],
+    previousMonthTitle: "previous month",
```
GomezIvann, line 1527: Please, all our titles start with capital letters, we must continue doing that.

lib/src/date-input/DatePicker.tsx
```diff
@@ -0,0 +1,207 @@
...
+      </PickerHeader>
+      {content === "calendar" && (
+        <Calendar
+          innerDate={innerDate}
+          selectedDate={selectedDate}
+          onInnerDateChange={setInnerDate}
+          onDaySelect={handleDateSelect}
+        />
+      )}
+
```
GomezIvann, line 107: Remove this empty space.

Lisa159:
Width of the calendar is different comparing it with the designs:

image

image

There is more space in the sides than in the design:
image

Adding `box-sizing: border-box;` in the `DatePicker` div fixes the problem.

lib/src/date-input/DateInput.tsx
```diff
@@ -36,28 +41,35 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
       tabIndex,
     },
     ref
```
GomezIvann, line 43: The `ref` prop is being declared but never connected with any component at all. If the user tries to use it it will be `undefined`!

lib/src/date-input/DateInput.tsx
```diff
@@ -36,28 +41,35 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
...
     const [innerValue, setInnerValue] = useState(defaultValue);
     const [isOpen, setIsOpen] = useState(false);
-    const [anchorEl, setAnchorEl] = useState(null);
+    const [calendarId] = useState(`date-picker-${uuidv4()}`);
+
+    useLayoutEffect(() => {
+      if (!disabled) {
+        let actionButtonRef;
+        if (clearable) {
+          actionButtonRef = refDate?.current.getElementsByTagName("button")[1];
```
GomezIvann, line 53: The useLayoutEffect is placed before the `refDate` declaration, although it hasn't been declared. I know is inside a useLayoutEffect which may be executed later on but it's a bit confusing from a code-understanding perspective.

GomezIvann (changes requested):
### Review

**Date Picker**
1. If the focus leaves the window, the date picker does not close.
2. Why does the outline from the previous/next month arrows have an offset of -2px? It doesn't make sense since it has enough space to appear. The same question goes for the year dropdown button.
3. The box-shadow was design to be complemented with a border (like in the Select) because in certain screens it can seem not visible:

image

4. Sometimes,  the focus leaves a kind of visual trace:

image

I think this is caused by the component dimensions plus thestyling itself, maybe the outline offset and the border setted to `none` are provoking this.

**Year selector**
1. Inconsistent scroll: Quick arrow presses can make the scroll work a bit strange and not always center the visually focused year.
2. The focus has some sort of memory, which I think doesn't quite fit the intended functionality. I mean, if the focus goes out from the years' list and comes back, it goes to the last focused item.

**Calendar**
1. Focus proposal for the Calendar (currently, is not working as expected):
   - If there's a day selected at the current month, the focus goes to this one.
   - If there's no day selected but **the actual date and the actual month** are being displayed, the focus goes to the actual day.
   - If both previous conditions are not true, the focus goes to the first day of the month.

2. If the day selected is from the previous/next month and it is also display in the current view (the light gray color days), it must be displayed as selected. See for example the [calendar from Cloudscape](https://cloudscape.design/components/calendar/?tabId=playground)

image

3. The days are too close to each other horizontally, but they have an small amount of space vertically. I think both rows and columns should have the same gap.

Date Picker (1)
Date Picker (2)

6. Same issue as in the year selector: The focus has some sort of memory, which I don't fully agree. It makes a bit unpredictable for the user to know where the focus is going to be placed, since sometimes it's memorized and other times not.

lib/src/date-input/types.ts
```diff
@@ -1,3 +1,5 @@
+import { Dayjs, UnitType } from "dayjs";
```
Lisa159, line 1: Removed unused `UnitType`

GomezIvann (changes requested):

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,263 @@
...
+  font-style: normal;
+  font-weight: 400;
+  font-size: 14px;
+  line-height: 19px;
+  letter-spacing: 0.03333em;
+`;
+
+const DayCellsContainer = styled.div`
+  background: ${(props) => props.theme.dateInput.pickerBackgroundColor};
+  height: 224px;
```
GomezIvann, line 205: This `height: 224px` extracted from the designs is not working as expected: 

Sin título

To fix this, you just need to remove that line and add a padding-top of `8px` (the calc would be 36px * 6 rows + 8px).

Sin título

GomezIvann (changes requested):

lib/src/date-input/YearPicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+    line-height: 33px;
+    color: ${(props) => props.theme.dateInput.pickerHoverDateFontColor};
+    outline: ${(props) => props.theme.dateInput.pickerFocusColor + " solid 2px"};
+  }
+  &:active {
+    font-weight: 400;
+    font-size: 1.5rem;
+    line-height: 33px;
+    color: ${(props) => props.theme.dateInput.pickerActiveDateFontColor};
+    background-color: ${(props) => props.theme.dateInput.pickerActiveDateBackgroundColor} !important;
```
GomezIvann, line 131: Same as above.

GomezIvann (changes requested):
Some comments:

- If you change the year, the current day appears focused instead of the first one.
- The scroll does not behave well yet. Same thoughts as in the last review and adding to that, we should remove the `smooth` behaviour of it, as in other scrollable components is being used.
- If you click on a day and then press tab button, the calendar closes instead of going to the previous month arrow.
- Storybooks:
  - We need to check every state and from the buttons of the Year Picker and the Calendar (focus, hover, active, etc.)
  - We need to check that the date picker appears over other components with a z-index higher than 0 (check the dropdown example to get an idea).

lib/src/date-input/DatePicker.tsx
```diff
@@ -0,0 +1,205 @@
...
+  };
+
+  useEffect(() => {
+    document.addEventListener("mousedown", handleClickOutside);
+    document.addEventListener("keydown", handleKeyboardEvent);
+    return () => {
+      document.removeEventListener("mousedown", handleClickOutside);
+      document.removeEventListener("keydown", handleKeyboardEvent);
+    };
+  }, [onCloseCalendar]);
```
GomezIvann, line 61: `onEscCalendar` dependecy is missing.

lib/src/date-input/DatePicker.tsx
```diff
@@ -0,0 +1,205 @@
...
+  border-radius: 4px;
+`;
+
+const PickerHeader = styled.div`
+  display: flex;
+  gap: 8px;
+  align-items: center;
+  justify-content: space-between;
+  box-sizing: border-box;
+  height: 40px;
```
GomezIvann, line 135: This height is unnecessary since it is already being defined by the height of the `HeaderYearTrigger`

lib/src/date-input/DatePicker.tsx
```diff
@@ -0,0 +1,205 @@
...
+  }
+`;
+
+const HeaderYearTriggerLabel = styled.span`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-family: ${(props) => props.theme.dateInput.pickerFontFamily};
+  font-size: 0.875rem;
+  font-weight: 400;
```
GomezIvann, line 201: Both `font-weight` and `line-height` I think they are unnecessary here (default value and don't do anything, respectively).

GomezIvann (changes requested):

lib/src/date-input/DateInput.tsx
```diff
@@ -278,46 +151,37 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
...
-              </Popover>
-            </StyledDPicker>
-          </MuiPickersUtilsProvider>
-        </MuiThemeProvider>
+            </Popover.Trigger>
+            <StyledContent
+              sideOffset={error ? -18 : 2}
+              align="end"
+              aria-modal={true}
+              onFocusOutside={handleFocusOutside}
```
GomezIvann, line 161: For some reason, the Radix Popover method onFocusOutside is never being called (you can check it by simply putting a console.log). We can fix this by replacing it with the `onBlur` event like this:

```
    const handleDatePickerOnBlur = (event) => {
      if (!event?.currentTarget.contains(event.relatedTarget)) {
        console.log("blur")
        closeCalendar();
      }
    };
```

This will allow us to remove also the window listener `window.addEventListener("blur", closeCalendar);`. This is a more specific and correct answer.

lib/src/date-input/DateInput.tsx
```diff
@@ -278,46 +151,37 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
...
 );
 
-const StyledDPicker = styled.div``;
+const StyledContent = styled(Popover.Content)`
+  &:focus-visible {
+    outline: none;
+  }
+`;
+
+const DxcDateInputContainer = styled.div``;
```
GomezIvann, line 185: The excess of styled components can affect performance, we should remove this and use directly `div`.

GomezIvann (changes requested):

lib/src/date-input/DatePicker.tsx
```diff
@@ -0,0 +1,205 @@
...
+      </PickerHeader>
+      {content === "calendar" && (
+        <Calendar
+          innerDate={innerDate}
+          selectedDate={selectedDate}
+          onInnerDateChange={setInnerDate}
+          onDaySelect={handleDateSelect}
+        />
+      )}
+
```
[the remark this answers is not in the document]
GomezIvann: Fixed.

lib/src/date-input/YearPicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+    !props.selected &&
+    `border: 1px solid ${props.theme.dateInput.pickerCurrentYearBorderColor}; color: ${props.theme.dateInput.pickerCurrentYearFontColor};`}
+  ${(props) =>
+    props.selected
+      ? `font-weight: 400;
+      font-size: 1.5rem;
+      line-height: 33px;
+      color: ${props.theme.dateInput.pickerSelectedDateColor} !important;
+      background-color: ${props.theme.dateInput.pickerSelectedDateBackgroundColor} !important;`
+      : `transparent;`}
```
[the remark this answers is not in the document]
GomezIvann: Fixed.

lib/src/date-input/YearPicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+  flex-direction: column;
+  gap: 4px;
+  align-items: center;
+  overflow-y: scroll;
+  box-sizing: border-box;
+  width: ${(props) => props.theme.dateInput.pickerWidth};
+  height: 312px;
+  padding: 0px 8px 8px 8px;
+  background-color: ${(props) => props.theme.dateInput.pickerBackgroundColor};
+  border-radius: 4px;
```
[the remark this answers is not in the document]
GomezIvann: Removed.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+const CalendarContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  box-sizing: border-box;
+  padding: 0px 8px 8px 8px;
+  width: ${(props) => props.theme.dateInput.pickerWidth};
+  background-color: ${(props) => props.theme.dateInput.pickerBackgroundColor};
+  border-radius: 4px;
```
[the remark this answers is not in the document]
GomezIvann: Removed.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,285 @@
...
+  }, [innerDate]);
+  const translatedLabels = useTranslatedLabels();
+  const weekDays = translatedLabels.calendar.daysShort;
+
+  const onDateClickHandler = (date: { day: number; month: number; year: number }) => {
+    const newDate = innerDate.set("month", date.month).set("date", date.day);
+    onDaySelect(newDate);
+  };
+
+  const calendarOnBlur = (event?) => {
```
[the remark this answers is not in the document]
GomezIvann: Removed.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+const Calendar = ({ selectedDate, innerDate, onInnerDateChange, onDaySelect }: CalendarPropsType): JSX.Element => {
+  const [dateToFocus, setDateToFocus] = useState(
+    selectedDate?.get("month") === innerDate.get("month") && selectedDate?.get("year") === innerDate.get("year")
+      ? selectedDate
+      : dayjs()
+  );
+  const [toFocus, setToFocus] = useState(false);
+  const today = dayjs();
+  const dayCells = useMemo(() => {
```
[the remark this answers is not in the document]
GomezIvann: Fixed.

lib/src/date-input/DatePicker.tsx
GomezIvann: Added.

lib/src/date-input/DatePicker.tsx
GomezIvann: Removed.

GomezIvann (changes requested):

lib/src/date-input/DatePicker.tsx
```diff
@@ -0,0 +1,205 @@
...
+    outline: ${(props) => props.theme.dateInput.pickerFocusColor + " solid 2px"};
+  }
+  &:hover {
+    background-color: ${(props) => props.theme.dateInput.pickerHoverDateBackgroundColor};
+  }
+  &:active {
+    background-color: ${(props) => props.theme.dateInput.pickerActiveDateBackgroundColor};
+    color: ${(props) => props.theme.dateInput.pickerActiveDateFontColor};
+  }
+  svg {
```
GomezIvann, line 161: Not pretty sure if all this SVG rule is doing too much stuff...

GomezIvann (changes requested):

lib/src/date-input/DateInput.tsx
GomezIvann: Done!

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,285 @@
...
+  };
+
+  useEffect(() => {
+    if (toFocus) {
+      document.getElementById(`day_${dateToFocus.get("date")}_month${dateToFocus.get("month")}`)?.focus();
+      setToFocus(false);
+    }
+  }, [dateToFocus]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
GomezIvann: The first one is already done.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,285 @@
...
+    setDateToFocus(date);
+    setToFocus(true);
+  };
+
+  useEffect(() => {
+    if (toFocus) {
+      document.getElementById(`day_${dateToFocus.get("date")}_month${dateToFocus.get("month")}`)?.focus();
+      setToFocus(false);
+    }
+  }, [dateToFocus]);
```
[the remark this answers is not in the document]
GomezIvann: Done.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,285 @@
...
+};
+
+const CalendarContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  box-sizing: border-box;
+  padding: 0px 8px 8px 8px;
+  width: ${(props) => props.theme.dateInput.pickerWidth};
+  background-color: ${(props) => props.theme.dateInput.pickerBackgroundColor};
```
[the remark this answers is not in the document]
GomezIvann: Removed.

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,285 @@
...
+  color: ${(props) => props.theme.dateInput.pickerWeekFontColor};
+`;
+
+const DayCellsContainer = styled.div`
+  display: flex;
+  gap: 4px;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  box-sizing: border-box;
+  background-color: ${(props) => props.theme.dateInput.pickerBackgroundColor};
```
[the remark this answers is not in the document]
GomezIvann: Removed.

lib/src/date-input/YearPicker.tsx
```diff
@@ -0,0 +1,135 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
+  align-items: center;
+  overflow-y: scroll;
+  box-sizing: border-box;
+  width: ${(props) => props.theme.dateInput.pickerWidth};
+  height: 312px;
+  padding: 0px 8px 8px 8px;
+  background-color: ${(props) => props.theme.dateInput.pickerBackgroundColor};
```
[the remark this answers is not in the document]
GomezIvann: Removed.

lib/src/date-input/DatePicker.tsx
```diff
@@ -0,0 +1,205 @@
...
+      onEscCalendar();
+    }
+  };
+  const handleClickOutside = (event) => {
+    if (ref?.current && !ref?.current.contains(event.target)) {
+      onCloseCalendar();
+    }
+  };
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
GomezIvann: Removed.

GomezIvann (approved):

marcialfps (approved):
💯

lib/src/date-input/Calendar.tsx
```diff
@@ -0,0 +1,285 @@
...
+        event.preventDefault();
+        dateToFocusTemp.get("day") !== 0 && (dateToFocusTemp = dateToFocusTemp.day(7));
+        focusDate(dateToFocusTemp);
+        break;
+      case " ":
+        event.preventDefault();
+        onDaySelect(dateToFocusTemp);
+        break;
+    }
+  };
```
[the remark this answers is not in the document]
GomezIvann: Done.

GomezIvann:
This branch is gonna be merged but there is still some work to be done.

PR #1377, opened 2022-11-10
[File Input] File item styles update

### LINKED ISSUE #1379: [File Input] File Item is not following the latest designs
**Describe the bug**
Both the component and its specifications are not following the latest designs Alex did in the past (https://xd.adobe.com/view/31372830-345b-408b-aa2c-19ccab7f6923-6464/grid). Some parts of the design are already implemented but not reflected in the design specifications. Also, the implementation has some visual bugs, so the tasks here are:

- [ ] Preview size is inconsistent (according to the designs, should always be 24x24 px). Clear action does not have the same distance to the right border in every variant. The preview icon should be 15x15 px and not 24x24 as it currently is.

image

- [ ] Update specification images to match the latest designs.

image

Current design

image

GomezIvann:
**Checklist**

**Description**
Using the grid layout to solve the issues the component had in regards to following the [latests specifications](https://xd.adobe.com/view/31372830-345b-408b-aa2c-19ccab7f6923-6464/grid/).

**Closes #1379**


GomezIvann:
> * In `filedrop` mode, please change the padding of the `ButtonContainer` to `4px 12px 4px 4px` instead of `2px 12px 2px 3px` as specified in the guidelines.
> 
> Current implementation: image
> 
> Guidelines: image
> 
> * In `filedrop` and `dropzone` modes, if it is not multiple and the file has an error, the error is not shown.
> 
> image
> 
> image
> 
> I think the error should be shown like this: image

Are you sure the padding is `4px 12px 4px 4px`? I mean, the styles of Alex say that the distance is `4px` taking into account the border (`4px` - `1px` of border = `3px` of padding), am I right?

raquelarrojo:
> > * In `filedrop` mode, please change the padding of the `ButtonContainer` to `4px 12px 4px 4px` instead of `2px 12px 2px 3px` as specified in the guidelines.
> > 
> > Current implementation: image
> > Guidelines: image
> > 
> > * In `filedrop` and `dropzone` modes, if it is not multiple and the file has an error, the error is not shown.
> > 
> > image
> > image
> > I think the error should be shown like this: image
> 
> Are you sure the padding is `4px 12px 4px 4px`? I mean, the styles of Alex say that the distance is `4px` taking into account the border (`4px` - `1px` of border = `3px` of padding, so the value would be `3px 12px 3px 3px`), am I right?

Yes, I totally agree. Taking a look to the specs, `4px` is the distance with the border. So we could calculate the padding as you say, calc(`4px - dropBorderThickness`), since `dropBorderThickness` is the token for the border width.

raquelarrojo (changes requested):
When it is `multiple={false}` and it has several files in value, it should show the files like this:

Code:
```
<DxcFileInput
    label="Single file with files in value"
    helperText="Please select files"
    callbackFile={callbackFile}
    value={allFiles}
    multiple={false}
/>
```
image

Now they are shown:
image

raquelarrojo (approved):

PR #1380, opened 2022-11-15
Fix dropdown menu resize when label changes

### LINKED ISSUE #1374: DxcApplicationLayout.Header.Dropdown wrong optionsList size
**Describe the bug**
After upgrading from v6.0.0 to v7.0.0, the component DxcApplicationLayout.Header.Dropdown displays the option list with the wrong size, the list is shrunk to minimum width instead of showing the options' values completely as it did.

**To Reproduce**
I attached our component code:
```

<DxcApplicationLayout.Header
      underlined={true}
      onClick={() => {}}
      padding={{ left: "medium", right: "large" }}
      content={
        <HeadingContainer>
          <DxcHeading
            level={3}
            text="DXC Assure Solution Composer"
            weight="normal"
            margin={{ left: "medium", right: "medium" }}
          />
          <DxcFlex>
            <DxcApplicationLayout.Header.Dropdown
              label={userName}
              options={optionsUser}
              iconSrc={accountIcon}
              iconPosition="before"
              onSelectOption={signOut}
            ></DxcApplicationLayout.Header.Dropdown>
          </DxcFlex>
        </HeadingContainer>
      }
    ></DxcApplicationLayout.Header>
```

**Expected behaviour**
Header dropdown options to be displayed as they did on v6.

**Screenshots**
- HDS v6.0.0
  - **Chrome**
imagen
  - **Firefox**
imagen

- HDS v7.0.0
  - **Chrome**
imagen
  - **Firefox**
imagen

marcialfps:
**Checklist**

**Description**
`label` prop is included in the `useEffect` that executes the resize of the dropdown menu.

**Screenshots**
Screenshots from this code:
```
const [label, setLabel] = useState("1");

  useEffect(() => {
    new Promise((resolve, reject) => {
      setTimeout(() => {
        resolve("foo");
      }, 1000);
    }).then(() => {
      setLabel("vicrodriguezgu");
    });
  }, []);

  return (
    <>
      <ExampleContainer>
        <Title title="Default" theme="light" level={4} />
        <DxcDropdown label={label} options={options} onSelectOption={(value) => {}} />
      </ExampleContainer>
    </>
  );
```

Before:
image

After:
image

**Closes #1374**

### DIFF
--- a/lib/src/dropdown/Dropdown.tsx
+++ b/lib/src/dropdown/Dropdown.tsx
@@ -157,7 +157,7 @@ const DxcDropdown = ({
     return () => {
       window.removeEventListener("resize", handleMenuResize);
     };
-  }, []);
+  }, [label]);
 
   return (
     <ThemeProvider theme={colorsTheme.dropdown}>


lib/src/dropdown/Dropdown.tsx
```diff
@@ -157,7 +157,7 @@ const DxcDropdown = ({
     return () => {
       window.removeEventListener("resize", handleMenuResize);
     };
-  }, []);
+  }, [label]);
```
GomezIvann, line 160: I don't think this is the right approach. Inside the array of dependencies should only go the `scope variables that are used by the side effect and change over time. From [React docs](https://reactjs.org/docs/hooks-reference.html#useeffect):

image

The `label` variable is not been used by the `useEffect` so it cannot go inside the dependency array, even if its change indirectly affects the effect. Also, that was not the intended purpose of this side effect (it has to be executed only once to add the event to the listeners of the window, so when the viewport changes the list also changes).

GomezIvann: The problem here (if I understood Viti's issue correctly) is that first the dropdown renders and then changes its label with the user's username, so the list needs to adapt to that dynamic change. This is a different side effect which needs to be approached in a specific way. Maybe we should use the [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) for this purpose.

GomezIvann (changes requested):

marcialfps:
Closing this PR in order to merge the solution purposed by @GomezIvann .

PR #1396, opened 2022-11-29
Upgraded jest to v29.2.3

Lisa159:
**Checklist**

**Purpose**
Dependabot is failing when upgrading Jest to v29.x

**Description**
- Fixed some warnings in the tests.
- Fixed tests that were failing after upgrading one by one the versions of Jest.
- Upgraded User Event to v13.
- Added ```--env=jsdom``` in the test script because it ensures that a mock browser environment is provided to our tests and thus allows them to pass.
- Installed ```jest-environment-jsdom``` because is no longer shipped by default.
- Added ```uuid``` in ```moduleNameMapper``` to avoid error with Jest. It maps the ```uuid``` module to it's commonJS export

### DIFF
--- a/lib/src/text-input/TextInput.test.js
+++ b/lib/src/text-input/TextInput.test.js
@@ -518,7 +518,7 @@ describe("TextInput component synchronous autosuggest tests", () => {
       <DxcTextInput label="Autocomplete Countries" value="Andor" suggestions={countries} onChange={onChange} />
     );
     const input = getByRole("combobox");
-    fireEvent.focus(input);
+    await userEvent.click(getByText("Autocomplete Countries"));
     expect(input.value).toBe("Andor");
     expect(getByText("Andor")).toBeTruthy();
     expect(getByText("ra")).toBeTruthy();
--- a/lib/src/text-input/TextInput.test.js
+++ b/lib/src/text-input/TextInput.test.js
@@ -70,7 +70,6 @@ describe("TextInput component tests", () => {
       <DxcTextInput label="Input label" placeholder="Placeholder" onChange={onChange} onBlur={onBlur} clearable />
     );
     const input = getByRole("textbox");
-
     fireEvent.focus(input);
     fireEvent.blur(input);
     expect(onBlur).toHaveBeenCalled();
@@ -86,7 +85,6 @@ describe("TextInput component tests", () => {
       <DxcTextInput label="Input label" placeholder="Placeholder" onChange={onChange} clearable />
     );
     const input = getByRole("textbox");
-
     fireEvent.change(input, { target: { value: "Test" } });
     expect(onChange).toHaveBeenCalled();
     expect(onChange).toHaveBeenCalledWith({ value: "Test" });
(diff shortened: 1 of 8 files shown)


lib/src/text-input/TextInput.test.js
```diff
@@ -394,7 +389,6 @@ describe("TextInput component tests", () => {
     expect(options[0].getAttribute("aria-selected")).toBeNull();
   });
 });
-
 describe("TextInput component synchronous autosuggest tests", () => {
   test("Autosuggest is displayed when the input gains focus", async () => {
```
GomezIvann, line 393: Remove the `async` word from this test and the two next to it, since it makes no sense to have it (there isn't any `await` inside of them).

lib/src/text-input/TextInput.test.js
```diff
@@ -394,7 +389,6 @@ describe("TextInput component tests", () => {
     expect(options[0].getAttribute("aria-selected")).toBeNull();
   });
 });
-
```
GomezIvann, line 397: Don't remove this space, have a line jump between describes to make the code more readable.

lib/src/text-input/TextInput.test.js
```diff
@@ -524,7 +518,7 @@ describe("TextInput component synchronous autosuggest tests", () => {
       <DxcTextInput label="Autocomplete Countries" value="Andor" suggestions={countries} onChange={onChange} />
     );
     const input = getByRole("combobox");
-    fireEvent.focus(input);
+    await userEvent.click(getByText("Autocomplete Countries"));
```
GomezIvann, line 521: There's no need for an `await` here, it works fine without it. In that case, we can also remove the async.

lib/src/text-input/TextInput.test.js
```diff
@@ -777,11 +771,12 @@ describe("TextInput component asynchronous autosuggest tests", () => {
       <DxcTextInput label="Autosuggest Countries" value="Denm" onChange={onChange} suggestions={callbackFunc} />
     );
     const input = getByRole("combobox");
-    fireEvent.focus(input);
-    await waitForElementToBeRemoved(() => getByText("Searching..."));
     expect(input.value).toBe("Denm");
+    await userEvent.click(getByText("Autosuggest Countries"));
```
GomezIvann, line 775: Same here, there's no need for an `await`.

lib/src/text-input/TextInput.test.js
```diff
@@ -777,11 +771,12 @@ describe("TextInput component asynchronous autosuggest tests", () => {
...
+    await waitForElementToBeRemoved(() => getByText("Searching..."));
     expect(getByText("Denmark")).toBeTruthy();
-    userEvent.click(getByRole("option"));
+    fireEvent.focus(getByRole("option"));
+    await userEvent.click(getByText("Denmark"));
```
GomezIvann, line 779: Same here, there's no need for an `await`.

GomezIvann (changes requested):

lib/src/text-input/TextInput.test.js
Lisa159: Done

lib/src/text-input/TextInput.test.js
Lisa159: Done

lib/src/text-input/TextInput.test.js
Lisa159: Done

GomezIvann (approved):

PR #1439, opened 2023-01-16
Date year focus & two digit year format fix

### LINKED ISSUE #1435: Date input not selecting correct the year with specific format
**Describe the bug**
Date input is not selecting the correct year with a specific format.

**To Reproduce**
In the following example, execute the steps:
```
<DxcDateInput label="Date input" format="dd/mm/yy" />
```

1. Select a year between 1899 and 1968, (for example 1910).
2. Select a day.
3. Go back to select a new year.
4. The selected year is in the 20xx, (it would be 2010).

**Expected behaviour**
It should have the correct year selected.

**Screenshots**
image

**Add labels**
Date input.

Jialecl:
**Checklist**

**Purpose**
Fixed some bugs:
- Fixed bug related to the focus when a year is selected
- Improved functionality when it is two-digit year format

And added some tests.

**Description**
Added a third party library to mock the "today value" in the tests.
Added visual tests and a new test for the new behavior related to the two-digit year format.

The  two-digit year format behavior works like this:
The first time a valid year is entered it will remember the value.
If we change the value of the year in the input it will add the first two digits of the year based on the last valid date year.

**Closes #1435**


lib/src/date-input/YearPicker.tsx
```diff
@@ -97,23 +100,20 @@ const YearPickerButton = styled.button<YearPickerButtonPropsType>`
       : ``}
 
   &:hover {
```
Lisa159, line 102: The font size and line height are being repeated in three different states, causing duplication of code. We can put it all together to avoid this:
```
  &:active, &:hover, &:focus{
    font-size: 1.5rem;
    line-height: 36px;
  }
```

lib/src/date-input/YearPicker.tsx
```diff
@@ -21,6 +22,8 @@ const YearPicker = ({ onYearSelect, selectedDate }: YearPickerPropsType): JSX.El
     yearToFocusEl?.focus();
```
GomezIvann, line 22: What do you think about changing the `scrollIntoView` parameters to:

`yearToFocusEl?.scrollIntoView?.({ block: "center", inline: "start" });`

so when the user opens the year picker the current focused year appears in the middle of the container and not at the bottom as it currently does.

GomezIvann (changes requested):

lib/src/date-input/Calendar.tsx
```diff
@@ -35,19 +35,23 @@ const getDays = (innerDate: Dayjs) => {
...
   selectedDate?.get("year") === date.year &&
   selectedDate?.get("date") === date.day;
 
 const Calendar = ({ selectedDate, innerDate, onInnerDateChange, onDaySelect }: CalendarPropsType): JSX.Element => {
-  const [dateToFocus, setDateToFocus] = useState(
-    selectedDate?.get("month") === innerDate.get("month") && selectedDate?.get("year") === innerDate.get("year")
-      ? selectedDate
-      : dayjs()
-  );
+  const [today] = useState(dayjs());
```
GomezIvann, line 52: Why `today` variable is a stateful variable? It never modifies its value so it may be a better approach to create a global variable outside the component.

Not everything needs to be a stateful variable, try always to keep the component's state as light as possible.

lib/src/date-input/Calendar.tsx
```diff
@@ -35,19 +35,23 @@ const getDays = (innerDate: Dayjs) => {
...
+  const [dateToFocus, setDateToFocus] = useState(getDateToFocus(selectedDate, innerDate, today));
   const [toFocus, setToFocus] = useState(false);
```
GomezIvann, line 54: As the stateful variable `toFocus` is used to determine whether the `dateToFocus` should be focused or not, I think `isFocusable` or `isFocused` is a better naming strategy (it took me a while to understand what it was doing by just simply reading its name).

GomezIvann (changes requested):

lib/src/date-input/Calendar.tsx
Jialecl: I agree, but in this case not being a stateful variable will create a new dayjs object for each render, and having it as a global variable was causing problems in the test, I will try to check it again in case I find a better solution.

lib/src/date-input/DateInput.stories.tsx
```diff
@@ -110,42 +113,96 @@ export const Chromatic = () => (
...
-  await userEvent.tab();
+const YearPickerIsolated = () => (
+  <>
+    <ExampleContainer expanded>
+      <Title title="Isolated year picker" theme="light" level={4} />
+      <YearPicker selectedDate={dayjs("06-04-1905", "DD-MM-YYYY")} onYearSelect={() => {}} />
+    </ExampleContainer>
+    <ExampleContainer expanded pseudoState="pseudo-focus">
+      <Title title="Isolated year picker focused" theme="light" level={4} />
+      <YearPicker selectedDate={dayjs("06-04-1905", "DD-MM-YYYY")} onYearSelect={() => {}} />
```
Lisa159, line 192: The focus is a bit cut.
image

lib/src/date-input/DateInput.stories.tsx
```diff
@@ -110,42 +113,96 @@ export const Chromatic = () => (
...
 
-export const ShowYearPickerFocus = YearPickerFocus.bind({});
-ShowYearPickerFocus.play = async ({ canvasElement }) => {
-  const canvas = within(canvasElement);
-  await userEvent.click(canvas.getByRole("combobox"));
-  await fireEvent.click(screen.getByText("April 2027"));
-  await userEvent.tab();
+const YearPickerIsolated = () => (
+  <>
+    <ExampleContainer expanded>
```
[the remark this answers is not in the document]
Lisa159: Without this `expanded`, the examples are getting closer which I think looks better.
image

lib/src/date-input/YearPicker.tsx
Jialecl: While this improves the scroll in the year picker this also makes the whole window scroll. I am not sure that I prefer this to how it was before.

lib/src/date-input/Calendar.tsx
GomezIvann: What I meant by global variable was outside of the component. It is rare that you have problems in the tests 🤔

lib/src/date-input/YearPicker.tsx
GomezIvann: But it also happens with `block: "nearest"`. It's not related to the parameters, that's how `scrollIntoViewWorks`, takes the whole screen.

Jialecl: Yes, the window scroll still happens with "nearest" but not every time the year picker is opened and the window is not exactly centered on the year to focus.

lib/src/date-input/Calendar.tsx
Jialecl: The issue lies in mocking the value of today in the tests. It seems unable to change today's value if it is instantiated outside the component.

lib/src/date-input/YearPicker.tsx
GomezIvann: Agree! For the moment, we stay like this.

lib/src/date-input/DateInput.tsx
```diff
@@ -46,10 +46,41 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
...
+        } else {
+          newDate = newDate.set("year", (lastValidYear <= 1999 ? 1900 : 2000) + +newDate.format("YY"));
+        }
+        return newDate;
+      }
+    };
+
+    useEffect(() => {
+      if (value || value === "") setDayjsDate(getDate(value));
+    }, [value]);
```
GomezIvann, line 82: `getDate` parameter is missing. If you want to avoid adding it, you can move outside the function `getDate` and pass the properties as references.

lib/src/date-input/DateInput.stories.tsx
```diff
@@ -108,9 +114,9 @@ export const Chromatic = () => (
 );
 
 const DatePicker = () => (
```
GomezIvann, line 116: I think the example 'Show Date Picker' can be avoided. If we add the regular Date Picker to the story 'Date Picker Button States' and apply the same play function to this first example we would be testing the same.

image

To keep testing the distance between the date picker and the text input you can add a click example to the first story 'Chromatic'.

lib/src/date-input/DateInput.stories.tsx
```diff
@@ -119,33 +125,100 @@ ShowDatePicker.play = async ({ canvasElement }) => {
...
+const YearPickerCalendar = () => (
+  <ExampleContainer>
     <Title title="Show year picker" theme="light" level={4} />
-    <DxcDateInput label="Date input" defaultValue="06-04-2027" />
+    <DxcDateInput label="Date input" defaultValue="06-04-1950" />
   </ExampleContainer>
 );
 
-export const ShowYearPicker = YearPicker.bind({});
+export const ShowYearPicker = YearPickerCalendar.bind({});
```
GomezIvann, line 143: Same as above, this can be done in the 'Year Picker Button States'.

GomezIvann (changes requested):
We are missing a test to check z-index in regards to the Date Picker popover.

image

lib/src/date-input/DatePicker.tsx
```diff
@@ -129,9 +135,8 @@ const HeaderButton = styled.button`
   svg {
     display: inline-block;
```
GomezIvann, line 136: You don't need to change the display of the SVG. Indeed, the only necessary rules are the width and height, user-select and flex-shrink are not doing anything at all...

GomezIvann:
```
svg {
   width: 24px;
   height: 24px;
}
```

GomezIvann (changes requested):

lib/src/date-input/DateInput.stories.tsx
```diff
@@ -107,45 +113,107 @@ export const Chromatic = () => (
...
-    <Title title="Show date picker" theme="light" level={4} />
-    <DxcDateInput label="Date input" defaultValue="06-04-2027" />
-  </ExampleContainer>
-);
+const DatePickerButtonStates = () => {
+  const colorsTheme: any = useTheme();
+  return (
+    <>
+      <ExampleContainer expanded>
+        <Title title="Show date picker over another element with z-index 0" theme="light" level={4} />
```
GomezIvann, line 121: I think you didn't understand me Jiale... What I meant is that we need to test that the date picker appears over other web elements with a `z-index` > 0. It's good to check that it appears over our components but the real case is with items like Dialogs, Headers, etc, whose `z-index` is higher than expected. For example, in the Autosuggest:

```
      <ExampleContainer>
        <Title title="Autosuggest listbox" theme="light" level={2} />
        <ExampleContainer>
          <Title title="List dialog uses a Radix Popover to appear over elements with a certain z-index" theme="light" level={3} />
          <div
            style={{
              display: "flex",
              flexDirection: "column",
              gap: "20px",
              height: "150px",
              width: "500px",
              marginBottom: "250px",
              padding: "20px",
              border: "1px solid black",
              borderRadius: "4px",
              overflow: "auto",
              zIndex: "1300",
            }}
          >
            <DxcTextInput
              label="Label"
              suggestions={countries}
              optional
              size="fillParent"
            />
            <button style={{ zIndex: "1", width: "100px" }}>Submit</button>
          </div>
        </ExampleContainer>
```

GomezIvann (changes requested):

lib/src/date-input/DateInput.stories.tsx
GomezIvann: These changes were not done as I mentioned so I reopen it.

lib/src/date-input/DateInput.stories.tsx
GomezIvann: These changes were not done as I mentioned so I reopen it.

lib/src/date-input/DateInput.stories.tsx
```diff
@@ -104,48 +110,128 @@ export const Chromatic = () => (
...
+  await userEvent.click(dateBtn);
 };
 
-const YearPickerFocus = () => (
-  <ExampleContainer expanded>
-    <Title title="Show year picker and focus" theme="light" level={4} />
-    <DxcDateInput label="Date input" defaultValue="06-04-2027" />
-  </ExampleContainer>
-);
+export const YearpickerStates = () => {
```
GomezIvann, line 179: Rename this to `YearPickerStates`

GomezIvann (commented):
Some thoughts:

- Summing up every component of the Date Input, we currently have 5 icons. Maybe we can create an Icons.tsx as in other components and export them from there.
- Why do both `leftCaret` and `rightCaret` have `aria-hidden` set to `true`?
- Remove the unused import from Calendar.tsx.

image

GomezIvann (approved):

PR #1443, opened 2023-01-17
Fixed autosuggest input with special characters

### LINKED ISSUE #1416: Autosuggest option values not working properly
**Describe the bug**
Writing certain strings in the autosuggest will show error because we are converting the value into a regular expression inside the component.

If we write for example `*`, the value of the autosuggest will be converted to regex. Because `*` is a wrong regular expression it will show this error.

**To Reproduce**
1. Create a new React App using `create-react-app`.
2. Install the `@dxc-technology\halstack-react` dependencies and run the app.
3. Write the following example:
```
  const countries = ["*", "Albania"];
  const onChange = ({ value }) => {
    setValue(value);
  };
  const onBlur = ({ value }) => {
    setValue(value);
  };

  const callbackFunc = (newValue) => {
    const result = new Promise((resolve) =>
      setTimeout(() => {
        resolve(
          newValue
            ? countries.filter((option) =>
                option.toUpperCase().includes(newValue.toUpperCase())
              )
            : countries
        );
      }, 1500)
    );
    return result;
  };
  return (
    <DxcInset space="2rem">
      <DxcTextInput
        label="Enter your country"
        value={value}
        onChange={onChange}
        suggestions={callbackFunc}
        onBlur={onBlur}
      />
    </DxcInset>
  );
```
4. Click the component and select the option `*`.
5. See the error.

**Expected behaviour**
It should select the option.

**Screenshots**
image

image

image

image

**Add labels**
Text input, autosuggest.

Lisa159:
**Checklist**

**Description**
Special characters were transformed to regular expression so it was giving an error if they were invalid. 
Added `\` to escape the special characters. For example, if we write `*` in the input, internally the component will transform the value received to `\*` so it will be valid.
Added new test.

About the implementation, I have created a regular expression with the special characters that are giving error. Then I have converted it to string and created an array with each character. It is needed to remove the duplicate characters from the array because the function `.toString()` is literally converting the regular expression in a string without removing the characters `/`, `[` and `]` that are used to create de regular expresion (example, `const a = /[a-z]/`).

**Screenshots**
Created a new function to transform the value:
image

**Closes #1416**

### DIFF
--- a/lib/src/text-input/Suggestion.tsx
+++ b/lib/src/text-input/Suggestion.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useMemo } from "react";
 import styled from "styled-components";
 import { SuggestionProps } from "./types";
 
@@ -24,10 +24,10 @@ const Suggestion = ({
   visuallyFocused,
   highlighted,
 }: SuggestionProps): JSX.Element => {
-  const specialChar = transformSpecialChars(value);
-  const regEx = new RegExp(specialChar, "i");
-  const matchedWords = suggestion.match(regEx);
-  const noMatchedWords = suggestion.replace(regEx, "");
+  const matchedSuggestion = useMemo(() => {
+    const regEx = new RegExp(transformSpecialChars(value), "i");
+    return { matchedWords: suggestion.match(regEx), noMatchedWords: suggestion.replace(regEx, "") };
+  }, [value, suggestion]);
 
   return (
     <SuggestionContainer
@@ -42,8 +42,8 @@ const Suggestion = ({
       <StyledSuggestion last={isLast} visuallyFocused={visuallyFocused}>
         {highlighted ? (
           <>
-            <strong>{matchedWords}</strong>
-            {noMatchedWords}
+            <strong>{matchedSuggestion.matchedWords}</strong>
+            {matchedSuggestion.noMatchedWords}
           </>
         ) : (
           suggestion
--- a/lib/src/text-input/Suggestion.tsx
+++ b/lib/src/text-input/Suggestion.tsx
@@ -2,6 +2,15 @@ import React from "react";
 import styled from "styled-components";
 import { SuggestionProps } from "./types";
 
+const transformSpecialChars = (str: string) => {
+  const specialChars = "[\\`!@#$%^&*()_+-=[]{};':\"|,.<>/?~]/";
+  let value = str;
+  specialChars.split("").forEach((specialChar) => {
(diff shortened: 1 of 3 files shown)


GomezIvann:
This is great stuff, Aida! 🚀 🚀 🚀

lib/src/text-input/TextInput.test.js
```diff
@@ -643,6 +645,54 @@ describe("TextInput component synchronous autosuggest tests", () => {
     expect(input.value).toBe("");
     expect(queryByRole("listbox")).toBeFalsy();
   });
+
+  test("Autosuggest escapes special characters", () => {
+    const onChange = jest.fn();
+    const { getAllByText, getByText, getByRole } = render(
+      <DxcTextInput label="Autocomplete Countries" suggestions={special_characters} onChange={onChange} />
```
Jialecl, line 652: Change "special_characters" to camel case.

lib/src/text-input/TextInput.test.js
```diff
@@ -643,6 +645,54 @@ describe("TextInput component synchronous autosuggest tests", () => {
...
+    );
+    const input = getByRole("combobox");
+    fireEvent.focus(input);
+    const list = getByRole("listbox");
+    fireEvent.change(input, { target: { value: "/" } });
+    expect(list).toBeTruthy();
+    expect(getAllByText("/").length).toBe(1);
+    expect(getByText("/")).toBeTruthy();
```
Jialecl, line 660: Isn't this redundant to the previous expect?

Lisa159: It is just double check, but yes, it is a bit redundant.

Lisa159: Done

lib/src/text-input/TextInput.test.js
Lisa159: Done

lib/src/text-input/Suggestion.tsx
```diff
@@ -11,7 +24,8 @@ const Suggestion = ({
   visuallyFocused,
   highlighted,
 }: SuggestionProps): JSX.Element => {
-  const regEx = new RegExp(value, "i");
+  const specialChar = transformSpecialChars(value);
```
Jialecl, line 27: This can be in a useMemo

Lisa159: Done

Jialecl (approved):

PR #1445, opened 2023-01-18
Fix custom icons sizing

### LINKED ISSUE #1418: `icon` prop wrong sizing
**Describe the bug**
The component is not styling the incoming SVG icons, only their container. In case the user gives the SVG a size bigger than expected, the interactive area of the component increases massively.

**To Reproduce**
Give the `icon` prop an SVG icon bigger than expected (this is: > `24x24px`). Then inspect it and see its area, if you hover in any part of that area you will see the bug.

**Expected behaviour**
Fit the container size.

**Screenshots**

image

GomezIvann:
**Checklist**

**Description**
Trying to solve custom icons issue when the size was bigger or smaller than expected. I standardized the implementation: Icons will always fulfil their parent container to perfectly fit the component's anatomy.

**Closes #1418**


lib/src/button/Button.stories.tsx
```diff
@@ -12,11 +12,42 @@ export default {
...
 );
+
+const facebookIcon = (
+  <svg
+    version="1.1"
+    id="Capa_1"
+    x="0px"
+    y="0px"
+    width="438.536px"
+    height="438.536px"
```
Jialecl, line 28: Even though the sizes are different the path seems to be that of a 24px icon.

Jialecl:
Should consider centering the icons if they are smaller than the space that they are trying to fill?

Jialecl:
As the SVGs were changed for the paginator to have the correct size I think that the same can be done for the sidenav for example.

lib/src/select/Option.tsx
```diff
@@ -106,12 +106,15 @@ type OptionIconProps = {
 };
 const OptionIcon = styled.span<OptionIconProps>`
   display: flex;
-  flex-wrap: wrap;
-  align-content: center;
-  height: 24px;
-  width: 24px;
-  ${(props) => (props.grouped && !props.multiple ? "padding-left: 16px;" : "padding-left: 8px;")}
+  padding: 0.125rem;
```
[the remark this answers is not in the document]
GomezIvann: The icon specifications say that the icon is 24x24, but after talking to Raffy, we changed that to the size of the container and the image now is 20x20 with a padding of 2px (0.125rem). 

image

So, 20 + 2 (above) + 2 (below) = 24px

GomezIvann:
> Should we consider centering in some way the icons if they are smaller than the space that they are trying to fill?
> 
> image

As we have already discussed, this is a problem of the SVG itself and not of our component.

lib/src/chip/Chip.stories.tsx
```diff
@@ -23,20 +46,29 @@ export const Chromatic = () => (
...
+      />
     </ExampleContainer>
     <ExampleContainer>
       <Title title="Disabled chip" theme="light" level={4} />
-      <DxcChip label="Disabled" disabled prefixIcon={iconSVG} suffixIcon={iconSVG} />
+      <DxcChip
+        label="Disabled"
+        disabled
+        prefixIcon={iconSVG}
+        suffixIcon="https://upload.wikimedia.org/wikipedia/commons/e/e0/Check_green_icon.svg"
```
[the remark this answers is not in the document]
GomezIvann: Yes... But that was a premeditated decision. Look for example the action of a Text Input when it receives an image:

image

This is because color styling cannot be properly applied to images and we don't usually style with `opacity` instead we use opacity in the colors. As you can read in the docs:

image

lib/src/sidenav/Sidenav.stories.tsx
```diff
@@ -41,11 +54,17 @@ export const Chromatic = () => (
...
+          <DxcSidenav.Group
+            collapsable={true}
+            title="Section Group"
+            icon="https://cdn-icons-png.flaticon.com/512/5039/5039041.png"
+          >
             <DxcSidenav.Link selected>Group Link</DxcSidenav.Link>
             <DxcSidenav.Link icon={iconSVG}>Group Link</DxcSidenav.Link>
           </DxcSidenav.Group>
-          <DxcSidenav.Link icon={iconSVG}>Single Link</DxcSidenav.Link>
+          <DxcSidenav.Link icon="https://upload.wikimedia.org/wikipedia/commons/7/73/Flat_tick_icon.svg">
```
Jialecl, line 65: The newWindow prop should be added to one of the links to have a test showing the external link.

GomezIvann: Done!

Jialecl (approved):

PR #1501, opened 2023-02-28
File input inside a form improvements

### LINKED ISSUE #1492: File Input bugs
- The description of the `callbackFile` prop does not specify at any point that it is going to be called whenever a file is added and also **deleted** (any time the inner value is modified, basically). We need to be more precise with this one.

image

- The internal hidden input value (value in forms) of the File Input does not work as expected. It only stores the first value(s) uploaded. If the user adds more files or deletes some of added ones, the internal input does not take it into account and submits the first stored values.

[screen-capture.webm](https://user-images.githubusercontent.com/44321109/220626269-393a4980-48d2-4cc6-910b-460821237c45.webm)

raquelarrojo:
**Checklist**

**Purpose**
Improve file input's documentation and behavior.

**Description**
- Improved `callbackFile` description in props table.
- Removed `name` from the input file to avoid that inconsistent data is received in submit inside a form.
- Added `type="button"` to file item's delete button, since by default button type is `submit`. This would cause a problem when the file input is inside a form.
- Changes `basicUsage` example to print `files` stateful variable instead of input file' `files`. Input file is uncontrolled and files are read only, so they do not contain files updated.
- Added file input inside a form example.
- Remove file input inside a form test, since we are not controlling what is arriving in submit event.

**Closes #1492**


lib/src/file-input/FileInput.tsx
```diff
@@ -179,7 +179,6 @@ const DxcFileInput = React.forwardRef<RefType, FileInputPropsType>(
                 accept={accept}
                 multiple={multiple}
                 onChange={selectFiles}
-                name={name}
                 disabled={disabled}
                 readOnly
                 aria-hidden="true"
```
GomezIvann, line 184: One comment here. I recently discovered that aria-hidden is not necessary when an element is hidden using `display: none;`

image


https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

Can you please remove it from this hidden input?

lib/src/file-input/FileInput.tsx
```diff
@@ -223,7 +222,6 @@ const DxcFileInput = React.forwardRef<RefType, FileInputPropsType>(
...
```
GomezIvann, line 227: Same here.

website/screens/components/file-input/code/examples/formFileInput.ts
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <form onSubmit={handleSubmit}>
+      <DxcInset space="2rem">
+        <DxcFlex direction="column" gap="2rem">
+          <DxcFileInput
+            label="Select your files"
+            value={files}
+            callbackFile={callbackFile}
+          />
+          <input type="submit" value="Submit" />
```
GomezIvann, line 30: I think we should use our implementation of a submit button instead of the browser's one: `<DxcButton type="submit" label="Submit" />`.

GomezIvann: I know it looks very similar to the first example, but I think they will understand the slight but important difference.

lib/src/file-input/FileInput.tsx
raquelarrojo: Totally agree!
Done.

lib/src/file-input/FileInput.tsx
raquelarrojo: Done!

website/screens/components/file-input/code/examples/formFileInput.ts
raquelarrojo: Agree! Done.

GomezIvann (approved):

PR #1504, opened 2023-03-01
Dialog review

GomezIvann:
**Checklist**
- [ ] Build process is done without errors and all tests pass in `/lib` directory.
- [ ] Self-reviewed the code prior to submitting.

**Purpose**
A clear description of the purpose of the change. Applicable in case this change is not linked to an issue.

**Description**
A clear and concise description of the change.

**Screenshots**
If applicable, add screenshots to help explain your change.

**Additional context**
Add any other context about the change here.

**Closes #`issue-number`**


Jialecl (commented):
These are some of the strange behaviours I have found out for now:

If there is an element with a `tabIndex > 0` the focus can leave the dialog.
When using the shift tab from the first element it seems to need an extra step to focus on the last element.

lib/src/common/variables.js
```diff
@@ -455,19 +455,16 @@ export const componentTokens = {
     closeIconHeight: "24px",
     closeIconTopPosition: "20px",
     closeIconRightPosition: "20px",
-    closeIconBackgroundColor: "none",
+    closeIconBackgroundColor: globalTokens.hal_white,
```
Jialecl, line 458: The value of this token does not match the values updated in the documentation.

GomezIvann: Fixed!

GomezIvann:
> These are some of the strange behaviours I have found out for now:
> 
> If there is an element with a `tabIndex > 0` the focus can leave the dialogue. When using the shift tab from the first element it seems to need an extra step to focus on the last element.

The second case is already fixed. In regards to the first case, I could reproduce it since in my laptop is working fine, could you please paste your code in a comment?

image

Jialecl:
I just simply added a `DxcButton` with `tabIndex={1}` to the dialog.
Related to this issue using the shift tab I was nerver able to focus on the button with positive tabIndex.

GomezIvann:
> I just simply added a `DxcButton` with `tabIndex={1}` to the dialog. Related to this issue using the shift tab I was nerver able to focus on the button with positive tabIndex.

I saw the problem. When we change the `tabIndex` of the inner content to a different value than the Dialog's one, it seems to lose the lock (understandable). I'm currently checking this since I'm not sure how it should work in this cases.

Jialecl:
> > I just simply added a `DxcButton` with `tabIndex={1}` to the dialog. Related to this issue using the shift tab I was nerver able to focus on the button with positive tabIndex.
> 
> I saw the problem. When we change the `tabIndex` of the inner content to a different value than the Dialog's one, it seems to lose the lock (understandable). I'm currently checking this since I'm not sure how it should work in this case.

One more thing I noticed is that if the first element has  `tabIndex > 0` it is still autofocused even if there are elements with higher priority.

GomezIvann:
> > > I just simply added a `DxcButton` with `tabIndex={1}` to the dialog. Related to this issue using the shift tab I was nerver able to focus on the button with positive tabIndex.
> > 
> > 
> > I saw the problem. When we change the `tabIndex` of the inner content to a different value than the Dialog's one, it seems to lose the lock (understandable). I'm currently checking this since I'm not sure how it should work in this case.
> 
> One more thing I noticed is that if the first element has `tabIndex > 0` it is still autofocused even if there are elements with higher priority.

That's because that's not how `tabIndex` works. Here, the main goal is to `autofocus the first available focusable item of the Dialog`. By changing the value of its `tabIndex` you are only establishing a different order in the normal sequence of tabbable elements. `tabIndex` is not used for autofocusing, for that kind of purpose is the `autofocus` property (which is what we are trying to reproduce).

website/screens/components/dialog/usage/DialogUsagePage.tsx
```diff
@@ -36,6 +36,11 @@ const sections = [
           Modal dialog boxes should overlay only a portion of the underlying
           page to keep the user oriented within the workflow.
         </DxcBulletedList.Item>
+        <DxcBulletedList.Item>
+          In a dialog, the focus should remain within the modal box until the
+          user completes the required action (if any) or closes it. In addition,
+          the focus should appear on the first focusable child when it opens.
```
Jialecl, line 42: This is not always the case since there are situations in which the interactions are limited and it may be advisable to set focus to the element that is likely to be most frequently used.

I think this can be reworded to generally the focus should appear on the first focusable child.

lib/src/dialog/Dialog.stories.tsx
```diff
@@ -170,112 +185,108 @@ export const DialogWithXsmallPadding = () => (
...
+          eros aliquam erat, ut efficitur ante neque id odio. Nam orci leo, dignissim sit amet dolor ut, congue gravida
+          enim. Donec rhoncus aliquam nisl, ac cursus enim bibendum vitae. Nunc sit amet elit ornare, malesuada urna eu,
+          fringilla mauris. Vivamus bibendum turpis est, id elementum purus euismod sit amet. Etiam sit amet maximus
+          augue. Vivamus erat sapien, ultricies fringilla tellus id, condimentum blandit justo. Praesent quis nunc
+          dignissim, pharetra neque molestie, molestie lectus.
+        </DxcParagraph>
+      </DxcFlex>
     </DxcDialog>
   </ExampleContainer>
 );
```
Jialecl, line 292: Maybe we can choose a better sized viewport to test the limits/changes to the responsiveness.

GomezIvann: I updated the stories. I added a new test to check the limits of the media queries (`45rem`/`720px`).

Jialecl (approved):

PR #1552, opened 2023-04-05
Re-writing variables with Typescript

GomezIvann:
**Checklist**

**Purpose**
While moving our repo to a mono repo with Nx I realized some tokens were pointing to none existing global tokens. That wouldn't have happened if we had Typescript in that file, so I decided to reimplement it with Typescript. This took a few code refactors, component inner updates and other challenges along the way.

**Description**
Summary of changes:

- New and more complete types for every `HalstackProvider` prop:
```
  theme?: DeepPartial<OpinionatedTheme>;
  advancedTheme?: DeepPartial<AdvancedTheme>;
  labels?: DeepPartial<TranslatedLabels>;
```
- `BackgroundColorContext`  types updated.
- New `BaseTypography` component added. It is an inner primitive component used for most typography-related components. It can be used by our components as well.
- Added library `@types/color`. This revealed some outdated usage which has already been fixed.
- Removed unused library `rgb-hex`.
- In /website, file renamed from `variables.js` to `variables.ts`.

**Screenshots**

### Prop-safety
image

### Token suggestions
Sin título


lib/src/HalstackContext.tsx
```diff
@@ -343,23 +337,24 @@ const parseTheme = (theme) => {
...
-  const parsedLabels = defaultTranslatedComponentLabels;
+const parseLabels = (labels: DeepPartial<TranslatedLabels>): TranslatedLabels => {
+  const parsedLabels = JSON.parse(JSON.stringify(defaultTranslatedComponentLabels));
+
   Object.keys(labels).map((component) => {
     if (parsedLabels[component]) {
       Object.keys(parsedLabels[component]).map((label) => {
-        if (labels[component][label]) {
+        if (labels[component][label] && labels[component]![label]) {
           parsedLabels[component][label] = labels[component][label];
```
raquelarrojo, line 358: Why is this condition added to the if statement? `labels[component]![label]`

GomezIvann: It's the typescript operator `!`. Checks if the value before it is null or not, which in this case doesn't make any sense since it is already checked in the previous condition. I don't know why I add that, I'm removing it.

raquelarrojo (approved):

Issue #1561, opened 2023-04-05
Remove ts-nocheck from Typograhpy


raquelarrojo:
Remove `ts-nocheck` from Typography.

Issue #1560, opened 2023-04-05
Remove ts-nocheck from Toggle group


raquelarrojo:
Remove `ts-nocheck` from Toggle group.

Issue #1559, opened 2023-04-05
Remove ts-nocheck from Textarea


raquelarrojo:
Remove `ts-nocheck` from Textarea.

Issue #1558, opened 2023-04-05
Remove ts-nocheck from Tag


raquelarrojo:
Remove `ts-nocheck` from Tag.

PR #1563, opened 2023-04-11
Remove ts-nocheck and change in typing styled components

### LINKED ISSUE #1544: Remove ts-nocheck from Alert.
Remove `ts-nocheck` from Alert.

raquelarrojo:
**Checklist**

**Purpose**
Remove `ts-nocheck` from components.
Unify how we are typing the styled components of all components.

**Description**
`ts-nocheck` removed from all components except from Select. Even so, some type fixing has been done in select component.
Unified all styled components typing in all components.

**Closes #1544 #1545 #1546 #1547 #1548 #1549 #1550 #1551 #1553 #1554 #1556 #1557 #1558 #1559 #1560 #1561 #1569**


lib/src/alert/Alert.tsx
```diff
@@ -122,7 +121,7 @@ const calculateWidth = (margin, size) =>
     ? `calc(${sizes[size]} - ${getMargin(margin, "left")} - ${getMargin(margin, "right")})`
     : sizes[size];
 
-const AlertModal = styled.div`
+const AlertModal = styled.div<{ mode: "inline" | "modal" }>`
```
GomezIvann, line 124: As we discuss, extracting the types from the `ComponentPropsType` may be a better approach for us:

image

GomezIvann: This would reduce the difficulty and cost of updating the type of a prop.

raquelarrojo: Totally agree! I will change it in all our components.

lib/src/spinner/Spinner.tsx
```diff
@@ -36,7 +34,7 @@ const DxcSpinner = ({
           </BackgroundSpinner>
```
GomezIvann, line 35: Can you remove this empty space?

GomezIvann (changes requested):
Some general comments:

- In the Card component, the stories are throwing me an error:
image

We have an image inside the component folder which is being exported alongside the component files after the build. Shouldn't we remove it and instead, in the stories, use an image from a simple URL?

lib/src/file-input/FileInput.tsx
```diff
@@ -285,7 +284,7 @@ const DxcFileInput = React.forwardRef<RefType, FileInputPropsType>(
   }
 );
 
-const FileInputContainer = styled.div`
+const FileInputContainer = styled.div<{ margin: FileInputPropsType["margin"]; name: FileInputPropsType["name"] }>`
```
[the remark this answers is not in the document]
raquelarrojo: Yes, it it not used in the styled component but necessary for the case of a file input inside a form. I will remove it from the styled component typing!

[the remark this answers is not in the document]
GomezIvann: But that's an error. `name` attribute does not exist in a `<div>` element so it won't do anything inside a form. Since we remove it from the input to avoid the component from sending wrong values when submitting, I think we should remove the prop. It doesn't do anything at this point.

If you pass a HTML markup validator it will trigger this as an error:
image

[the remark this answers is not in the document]
GomezIvann: Code from the example above:

image

lib/src/spinner/Spinner.tsx
```diff
@@ -110,14 +109,14 @@ const DxcSpinner = ({
   );
 };
 
-const determinatedValue = (props, strokeDashArray) => {
+const determinatedValue = (value, strokeDashArray) => {
```
GomezIvann, line 112: Sorry not to tell you this before, but can you add the `SpinnerPropsType["value"]` and `number` types to both parameters?

lib/src/file-input/FileInput.tsx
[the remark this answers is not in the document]
raquelarrojo: Totally agree!!

GomezIvann (approved):

PR #1589, opened 2023-05-05
Updates to Sidenav component

### LINKED ISSUE #1587: Change Sidenav.Link children prop type
We have found a valid use case for which the API of the `DxcSidenav.Link` does not fit properly. This use case was to add a status tag, inline with the text of the Sidenav link, to let the user know the current state of each documentation page. Since the children of the Link are typed as `string`, that is currently not allowed.

The feature here is to replace this type with `React.ReactNode`.

GomezIvann:
**Checklist**

**Description**
Summary of changes in the Sidenav component:

- sidenav.Link prop `children` has changed its type from `string` to `React.ReactNode`.
- Specs updated:
  - Removed line heights to be able to properly align text inside a Sidenav link.
  - Removed unnecessary margin-right for svgs inside the compound component Title.
  - Removed React.Children from the implementation and using instead React Context API. In the new React site of documentation they set Children as a legacy API and recommend not to use it whenever possible (https://react.dev/reference/react/Children).
- Documentation updated:
  - More logical hierarchy and improved descriptions.
  - Updated children's prop type.
  - Updated description for some props.

**Closes #1587**


website/screens/components/sidenav/code/SidenavCodePage.tsx
```diff
@@ -22,216 +22,211 @@ const sections = [
...
+                Element or path used as the icon that will be placed to the left
+                of the link text.
+              </td>
+            </tr>
+            <tr>
+              <td>selected: boolean</td>
+              <td>
+                <Code>false</Code>
+              </td>
+              <td>If true, the link will be marked as selected.</td>
```
Jialecl, line 206: Maybe we can add a recommendation regarding this prop and how the group if collapsable should have this prop accessible in a direct child.

Jialecl:
The hover collapsed group title does not look correct.

Jialecl:
Also I noticed that we don't have a visual test that covers the active state.

website/screens/components/sidenav/code/SidenavCodePage.tsx
GomezIvann: Is hard to make the prop accessible if they are using a custom or external link (forcing them to add the `selected` prop just for our implementation to use it and not theirs is a bit strange and difficult to understand). But I agree that we should notify them that if they don't use our Sidenav link this can lead to some functionality loss.

GomezIvann:
We are gonna close this pull request and continue the work in a new one, since we have made too many changes just to fix a chromatic issue.

PR #1595, opened 2023-05-09
Fix Focus Lock issue

### LINKED ISSUE #1586: Jest Test - issue "Focus"
**Describe the bug**
I'm facing issue - Jest test case - after updating to Halstack 9
Dialog-Box, Any idea ?

**To Reproduce**
``` cmd
npm run test:ci
```
Example:

1. Install the `@dxc-technology\halstack-react@9.0.0` dependencies and run the app.

**Screenshots**
image

GomezIvann:
**Checklist**

**Description**
As described in the linked issue, an error was spotted at the Dialog when using Jest and having to render this component. After some research, this problem was related to a [nwsapi](https://github.com/dperini/nwsapi) problem, which is a library used by Jest `jsdom`.

So, to fix this, the aria-disabled selector was removed from the tabbable elements query at the `FocusLock` component. Apart from solving this problem, the change is motivated by the fact that `aria-disabled` is a semantic attribute, i.e. it has no functionality on the component. Therefore, excluding elements with this attribute from the list of tabbable elements of the focus lock is unnecessary.

**Additional context**
Issue: https://github.com/dperini/nwsapi/issues/81

**Closes #1586**

### DIFF
--- a/lib/src/textarea/Textarea.test.js
+++ b/lib/src/textarea/Textarea.test.js
@@ -64,7 +64,7 @@ describe("Textarea component tests", () => {
     expect(textarea.value).toBe("Example text");
   });
 
-  test("Disabled textarea renders with correct aria and can not be modified", () => {
+  test("Disabled textarea can not be modified", () => {
     const onChange = jest.fn();
     const { getByLabelText } = render(<DxcTextarea label="Example label" onChange={onChange} disabled />);
     const textarea = getByLabelText("Example label");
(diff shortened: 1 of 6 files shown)


lib/src/textarea/Textarea.test.js
```diff
@@ -69,7 +68,6 @@ describe("Textarea component tests", () => {
     const onChange = jest.fn();
```
aidamag, line 68: I would change the name of this test as now we are not checking any aria.

aidamag:
Also, I did `npm install` in the lib folder and I can see a new dependency added into the package-lock.json. I think it would be a good idea to update it.

image

aidamag (approved):

PR #1618, opened 2023-06-13
Radio Group update

GomezIvann:
**Checklist**

**Description**
Code updates based on some issues spotted during UXPin integration.

### DIFF
--- a/lib/src/radio-group/RadioGroup.tsx
+++ b/lib/src/radio-group/RadioGroup.tsx
@@ -73,7 +73,6 @@ const DxcRadioGroup = React.forwardRef<RefType, RadioGroupPropsType>(
       // If the radio group loses the focus to an element not contained inside it...
       if (!event.currentTarget.contains(event.relatedTarget as Node)) {
         setFirstTimeFocus(true);
-
         const currentValue = value ?? innerValue;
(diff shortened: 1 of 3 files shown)


lib/src/radio-group/RadioGroup.tsx
```diff
@@ -42,12 +42,21 @@ const DxcRadioGroup = React.forwardRef<RefType, RadioGroupPropsType>(
...
-      disabled,
-    };
-    const innerOptions = useMemo(() => (optional ? [...options, optionalItem] : options), [optional, options]);
+    const innerOptions = useMemo(
+      () =>
+        optional
+          ? [
+              ...options,
+              {
+                label: optionalItemLabel ?? translatedLabels.radioGroup.optionalItemLabelDefault,
```
Jialecl, line 51: Changing || to ?? will change the accepted values for optionalItemLabel.
For example the optionalItemLabel="" will be accepted instead of defaulting to N/A.

GomezIvann: I thought about that. I see no impediment to why we cannot accept it as a valid value (the empty string value). Only `undefined` would add the default value to the prop.

Jialecl (approved):

PR #1632, opened 2023-06-20
Remove deprecated `contentPadding` prop from the Card component

### LINKED ISSUE #1626: Remove deprecated `contentPadding` prop from Card

GomezIvann:
**Checklist**

**Closes #1626**


lib/src/card/Card.tsx
```diff
@@ -14,48 +14,46 @@ const DxcCard = ({
...
-  const imageComponent = (
-    <ImageContainer imageBgColor={imageBgColor}>
-      <TagImage imagePadding={imagePadding} imageCover={imageCover} src={imageSrc}></TagImage>
-    </ImageContainer>
-  );
-
   return (
-    <StyledDxcCard
+    <Card
+      hasAction={onClick || linkHref ? true : false}
```
Jialecl, line 26: We can have this condition in a variable to avoid repeating it.

GomezIvann: Remember that the variables defined inside a React component are always recalculated every time the component re-renders. Since the code repetition is not in enough places, I think it is not worth it (it would require a `useMemo`).

Jialecl (approved):

PR #1647, opened 2023-07-07
Adding `title` prop to Button 

### LINKED ISSUE #1483: Add `title` prop to the Button component
**Is your feature request related to a problem? Please describe.**
After some feedback regarding the necessity of some sort of [Tooltip](https://github.com/dxc-technology/halstack-react/issues/1440) component, a temporary solution and interesting feature for the `DxcButton` is to add/open the `title` prop to the user. This would allow them to create accessible action buttons (buttons with only an icon as the label) and display a small label when they hover the button, so the final user can know which action the button does.

This feature should not be too much of an issue since we already added a title prop to the actions of the `DxcTextInput`.

Add `title` prop to `DxcButton`.

GomezIvann:
**Checklist**

**Description**
Added a new prop `title` to the Button component. It's meant to fulfil the requirement of an accessible icon-only button.
Also, some padding and margin specs from the component have been updated to make it more intuitive and consistent.

**Closes #1483**

### DIFF
--- a/website/screens/components/button/usage/ButtonUsagePage.tsx
+++ b/website/screens/components/button/usage/ButtonUsagePage.tsx
@@ -114,6 +114,12 @@ const sections = [
             is mandatory.
           </DxcBulletedList.Item>
           <DxcBulletedList.Item>
+            The <Code>title</Code> prop offers a contextual bubble with missing
+            information necessary to clarify the action performed by the button.
+            It also provides an accessible gateway when no regular label can be
+            specified.
+          </DxcBulletedList.Item>
+          <DxcBulletedList.Item>
             Try to limit the use of icon-only buttons. Whenever possible, the
             icon should be accompanied by a label.
           </DxcBulletedList.Item>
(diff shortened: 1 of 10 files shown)


website/screens/components/button/usage/ButtonUsagePage.tsx
```diff
@@ -91,15 +92,15 @@ const sections = [
...
+          user to choose which one suits its use case better. But, it is
+          recommended to use the same library of icons throughout the
+          application to keep consistency.
         </DxcParagraph>
         <DxcBulletedList>
           <DxcBulletedList.Item>
-            Icon can go before or after the text with a separation of 8 pixels
-            respecting the main button&#39;s text.
+            The icon can go before or after the text with a separation of 8
+            pixels respecting the main button's text.
```
Jialecl, line 103: We should maintain &#39; instead of "'".

GomezIvann: For the moment, we are going to keep this until we know more about this issue.

Jialecl (approved):

PR #1688, opened 2023-09-26
Add `readOnly` prop to the Text Input and Date Input

### LINKED ISSUE #1687: Add the `readOnly` prop to the Text Input component
Design, implement and document a new read-only variant for the Text Input component.

GomezIvann:
**Checklist**

**Description**
The new read-only variant of the Text Input and the Date Input.

**Closes #1687 and #1689**

### DIFF
--- a/lib/src/text-input/types.ts
+++ b/lib/src/text-input/types.ts
@@ -61,6 +61,10 @@ type Props = {
    */
   disabled?: boolean;
   /**
+   * If true, the component will not be mutable, meaning the user can not edit the control.
+   */
+  readOnly?: boolean;
+  /**
    * If true, the input will be optional, showing '(Optional)'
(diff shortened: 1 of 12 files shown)


lib/src/text-input/types.ts
```diff
@@ -60,6 +60,10 @@ type Props = {
    * If true, the component will be disabled.
    */
   disabled?: boolean;
+  /**
+   * If true, the component will not be mutable, meaning the user can not edit the control.
```
Jialecl, line 64: What do you think about mentioning that it will affect the events' behavior?

GomezIvann: Agree on that!

Jialecl (approved):

PR #1696, opened 2023-10-02
Add `readOnly` prop to Number Input component

### LINKED ISSUE #1697: Add the `readOnly` prop to the Number Input component
Design, implement and document a new read-only variant for the Number Input component.

GomezIvann:
**Checklist**

**Description**
The new read-only variant of the Number Input.

**Closes #1697**

### DIFF
--- a/lib/src/text-input/TextInput.stories.tsx
+++ b/lib/src/text-input/TextInput.stories.tsx
@@ -457,7 +457,6 @@ const AutosuggestListbox = () => {
               optional
               placeholder="Choose an option"
               size="fillParent"
-              readOnly
             />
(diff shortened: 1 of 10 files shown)


lib/src/text-input/TextInput.stories.tsx
```diff
@@ -457,6 +457,7 @@ const AutosuggestListbox = () => {
               optional
               placeholder="Choose an option"
               size="fillParent"
+              readOnly
```
Jialecl, line 460: This test case should not have readOnly.

GomezIvann: Oops! My fault, I removed it.

PR #1766, opened 2024-01-19
Fixing typing for the whole Select component

### LINKED ISSUE #1555: Remove ts-nocheck from Select
Remove `ts-nocheck` from Select.

Mil4n0r:
**Checklist**

**Description**
Removed ts-nocheck and fixed the problems related to it. Also typed all the methods and variables.

**Additional context**
A few small changes to logic were made in order to allow a better readibility and, overall, cleaner code. Also removed redundant types (SingleSelect and MultipleSelect could be merged into one type).

**Closes #1555**


lib/src/select/types.ts
```diff
@@ -103,43 +103,14 @@ type CommonProps = {
...
-type MultipleSelect = CommonProps & {
+
+type Props = CommonProps & {
   /**
    * If true, the select component will support multiple selected options.
    * In that case, value will be an array of strings with each selected
    * option value.
    */
-  multiple: true;
+  multiple?: boolean;
```
GomezIvann, line 113: We should stick to the previous implementation with `SingleSelect` and `MultipleSelect` types. The point of how it was done was to guide the user through the construction of a Select using the types. For example, if `multiple` is true, the `defaultValue` should be an array of strings.

The solution should only be in the Select.tsx files, not in the current types.

GomezIvann (changes requested):

lib/src/select/types.ts
Mil4n0r: The problem is that, using those types, for some reason both the onChange and the onBlur callback seem to be inferred as the following: 

```
(val: {
    value: string;
    error?: string;
} & {
    value: string[];
    error?: string;
}) => void
```
Instead of the expected:
```
((val: {
    value: string;
    error?: string;
}) => void) | ((val: {
    value: string[];
    error?: string;
}) => void)
```

I have tried the approach recommended here: [Conditional Types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) among others recommended in forums with no success for this case.

Mil4n0r: We have decided that the best and easiest approach would be to use type assertion. [More info here](https://stackoverflow.com/questions/76188330/typescript-union-type-of-functions-weirdly-becomes-an-intersection-type-of-the-a)

lib/src/select/Select.tsx
```diff
@@ -151,8 +147,8 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
     const [visualFocusIndex, changeVisualFocusIndex] = useState(-1);
     const [isOpen, changeIsOpen] = useState(false);
 
-    const selectRef = useRef(null);
-    const selectSearchInputRef = useRef(null);
+    const selectRef = useRef<HTMLInputElement | null>(null);
```
GomezIvann, line 150: The select ref is on a div element (<Select> styled-component). We should type this one as an `HTMLDivElement`.

GomezIvann: This affects the `useWidth` parameter, it should also be an `HTMLDivElement` element.

lib/src/select/Select.tsx
```diff
@@ -292,27 +288,27 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
...
 
-    const handleSearchIOnChange = (event) => {
+    const handleSearchIOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
       setSearchValue(event.target.value);
       changeVisualFocusIndex(-1);
       openOptions();
     };
 
-    const handleClearOptionsActionOnClick = (event) => {
+    const handleClearOptionsActionOnClick = (event: React.MouseEvent) => {
```
GomezIvann, line 297: I think we should type the MouseEvent like this: `React.MouseEvent<HTMLButtonElement>`.

lib/src/select/Select.tsx
```diff
@@ -292,27 +288,27 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
...
       value ?? setInnerValue([]);
       !optional
-        ? onChange?.({ value: [], error: translatedLabels.formFields.requiredValueErrorMessage })
-        : onChange?.({ value: [] });
+        ? onChange?.({ value: [] as string & string[], error: translatedLabels.formFields.requiredValueErrorMessage })
+        : onChange?.({ value: [] as string & string[]});
     };
 
-    const handleClearSearchActionOnClick = (event) => {
+    const handleClearSearchActionOnClick = (event: React.MouseEvent) => {
```
GomezIvann, line 305: Same as above.

GomezIvann (changes requested):

GomezIvann:
One last comment, the code is unformatted. Can you please install the code formatter we use in Halstack in your Visual Studio Code IDE?

Name: Prettier - Code formatter
Id: esbenp.prettier-vscode
Description: Code formatter using prettier
Version: 10.1.0
Publisher: Prettier
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

lib/src/select/Select.tsx
```diff
@@ -1,31 +1,27 @@
...
 import Listbox from "./Listbox";
 import * as Popover from "@radix-ui/react-popover";
 
-const groupsHaveOptions = (innerOptions) =>
-  innerOptions[0].hasOwnProperty("options")
-    ? innerOptions[0].options
-      ? innerOptions.some((groupOption) => groupOption.options.length > 0)
-      : false
-    : true;
+const isOption = (option: Option | OptionGroup): option is OptionGroup => 'options' in option;
```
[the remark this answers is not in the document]
Mil4n0r: Your proposal looks good to me!

Mil4n0r:
Thanks for the detailed review Iván! I just added the requested changes 😄

GomezIvann (approved):

GomezIvann:
Great job Enrique, this was a tough one!

PR #1773, opened 2024-01-23
Fixed bug with checkboxes in resultset table

### LINKED ISSUE #1752: Checkbox not working properly with Resultset Table
**Describe the bug**
When you add a DxcCheckbox component for each item to rows array, checkboxes not being checked separately, instead of, checkboxes are being checked based on index in the pagination. I can checked each checkbox separately if I use checked prop but every time I update the state of the checkboxes, the table is re-rendered and it returns to the first page.

**To Reproduce**
1. Create a simple component using the example in the documentation https://developer.dxc.com/halstack/10/components/resultset-table/

2. Add a DxcCheckbox for each element like this:

`
const rows = [
   [
      { displayValue: <DxcCheckbox /> },
      { displayValue: "001"},
      { displayValue: "Cris"},
      { displayValue: "Paris"}
    ]
 ]`

3. Add more than 5 elements in the rows array, in order to display pagination

4. Check the first DxcCheckbox in the first page and navigate to the others pages, you will see all DxcCheckbox in the first position will be checked

**Expected behaviour**
We want to check each element into the array separately to handle the checkboxes are been checked

**Screenshots**
checkboxes-resultset-table

Mil4n0r:
**Checklist**

**Description**
Fixes the bug that occurred when trying to change pages after modifying a `Checkbox` within a `Resultset Table` component.

**Additional context**
It not also applied to `Checkbox` but to any updatable component (`TextArea`, etc.)

The problem happened due to the key value not being calculated properly when iterating through the `filteredResultSet` (as it was not taking into account the current page index)

**Closes #1752**

### DIFF
--- a/lib/src/resultset-table/ResultsetTable.test.js
+++ b/lib/src/resultset-table/ResultsetTable.test.js
@@ -2,6 +2,7 @@ import React from "react";
 import { render, fireEvent } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import DxcResultsetTable from "./ResultsetTable.tsx";
+import DxcCheckbox from "../checkbox/Checkbox.tsx";
 
 // Mocking DOMRect for Radix Primitive Popover
 global.globalThis = global;
@@ -218,6 +219,15 @@ const rows2 = [
   ],
 ];
 
+const checkboxRows = [
+  [ { displayValue: <DxcCheckbox checked /> }, { displayValue: "Peter" }, { displayValue: "Miami" }],
+  [ { displayValue: <DxcCheckbox checked /> }, { displayValue: "Louis" }, { displayValue: "London" }],
+  [ { displayValue: <DxcCheckbox checked /> }, { displayValue: "Lana" }, { displayValue: "Amsterdam" }],
+  [ { displayValue: <DxcCheckbox /> }, { displayValue: "Rick" }, { displayValue: "London" }],
+  [ { displayValue: <DxcCheckbox /> }, { displayValue: "Mark" }, { displayValue: "Miami" }],
+  [ { displayValue: <DxcCheckbox /> }, { displayValue: "Cris" }, { displayValue: "Paris" }],
+];
+
 describe("Resultset table component tests", () => {
   test("Resultset table rendered correctly", () => {
     const { getByText } = render(<DxcResultsetTable columns={columns} rows={rows} itemsPerPage={3} />);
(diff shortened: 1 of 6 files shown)


lib/src/resultset-table/ResultsetTable.test.js
```diff
@@ -218,6 +219,15 @@ const rows2 = [
...
 
+const checkboxRows = [
+  [ { displayValue: <DxcCheckbox checked /> }, { displayValue: "Peter" }, { displayValue: "Miami" }],
+  [ { displayValue: <DxcCheckbox checked /> }, { displayValue: "Louis" }, { displayValue: "London" }],
+  [ { displayValue: <DxcCheckbox checked /> }, { displayValue: "Lana" }, { displayValue: "Amsterdam" }],
+  [ { displayValue: <DxcCheckbox /> }, { displayValue: "Rick" }, { displayValue: "London" }],
+  [ { displayValue: <DxcCheckbox /> }, { displayValue: "Mark" }, { displayValue: "Miami" }],
+  [ { displayValue: <DxcCheckbox /> }, { displayValue: "Cris" }, { displayValue: "Paris" }],
+];
+
```
Jialecl, line 230: i am not sure this visual test is necessary since it does not test the scenario that this pr is trying to fix.

lib/src/resultset-table/ResultsetTable.test.js
```diff
@@ -244,6 +254,20 @@ describe("Resultset table component tests", () => {
     expect(getAllByRole("row").length - 1).toEqual(3);
   });
 
+  test("Resultset table handles components properly after swapping page", () => {
+    const { getByText, getAllByRole } = render(<DxcResultsetTable columns={columns} rows={checkboxRows} itemsPerPage={3} />);
```
Jialecl, line 258: getByText can be removed since it is not used.

lib/src/resultset-table/ResultsetTable.tsx
```diff
@@ -119,7 +119,7 @@ const DxcResultsetTable = ({
           </thead>
           <tbody>
             {filteredResultset.map((cells, index) => (
-              <tr key={`resultSetTableCell_${index}`}>
+              <tr key={`resultSetTableCell_${page * itemsPerPage + index ?? index}`}>
```
Jialecl, line 122: I think this can be simplified to resultSetTableCell_page_index.

Mil4n0r: Yup, looks much neater.

lib/src/resultset-table/ResultsetTable.test.js
Mil4n0r: You are right, the visual part should simply be checked in Cromatic. I could add a test that checks the value of the key in the <tr> using JSDOM but it seems a bit overkill and unneeded.

Jialecl (approved):

PR #1796, opened 2024-02-01
Added 100% width underline to NavTab

### LINKED ISSUE #1525: Nav Tabs divider not occupying 100% width available
**Describe the bug**
Each tab of the Nav Tabs component has its divider instead of the main container. This is a bit strange since, normally, in most of the Tabs implemented by other DSs the divider fits 100% of its parent container. Even on our site, we extend this line manually to try to fill the available space. This use case would be widespread in our users when using our component and it currently doesn't have a solution from our side.

This issue is related to #1524. Fixing this one would also fix that one.

**Expected behaviour**
The divider should occupy the whole width of the parent container.

**Screenshots**
image
image

**Additional context**
Other Design Systems:

- https://spectrum.adobe.com/page/tabs/ and https://react-spectrum.adobe.com/react-spectrum/Tabs.html
- https://cloudscape.design/components/tabs/?tabId=playground
- https://www.lightningdesignsystem.com/components/tabs/
- https://ant.design/components/tabs

Mil4n0r:
**Checklist**

**Description**
It has been decided that it is more appropiate for NavTabs component to always fill all the available space, that prevents ourselves from using "tricks" in order to achieve that.

**Screenshots**
image

**Closes #1525**

### DIFF
--- a/lib/src/common/variables.ts
+++ b/lib/src/common/variables.ts
@@ -506,6 +506,7 @@ export const componentTokens = {
     selectedIconColor: CoreTokens.color_grey_700,
     unselectedIconColor: CoreTokens.color_grey_700,
     disabledIconColor: CoreTokens.color_grey_500,
+    dividerThickness: "2px",
   },
   paginator: {
     backgroundColor: CoreTokens.color_grey_100,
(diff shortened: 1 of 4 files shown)


lib/src/common/variables.ts
```diff
@@ -506,6 +506,7 @@ export const componentTokens = {
     selectedIconColor: CoreTokens.color_grey_700,
     unselectedIconColor: CoreTokens.color_grey_700,
     disabledIconColor: CoreTokens.color_grey_500,
+    dividerThickness: "2px",
```
GomezIvann, line 509: Has this token been agreed with a designer?

Mil4n0r: It has been decided that this token is not needed, so I am removing it.

GomezIvann (approved):

PR #1798, opened 2024-02-02
Tooltip support in the Select component

### LINKED ISSUE #1495: Add built-in tooltips to the Select options
**Is your feature request related to a problem?**
Recently, we have been requested to add support to longer options labels in the Select component. Since implementing a line break would lead to ugly scenarios (visually) and the list should have the same size as the input, we have decided to use this HTML solution to display the full label without the need to address some unexpected scenarios inside the list.

Add the `title` prop to the Select options.

Sin título

**Additional context**
The full discussion can be founded [here](https://github.com/dxc-technology/halstack-react/discussions/901)

GomezIvann:
**Checklist**

**Description**
A tooltip has been added to the Select component. When the label of an option, both when it is selected and in the listbox, is too long, a built-in tooltip will appear to help the user see the full text.

This functionality should be tested manually since there's no way of testing refs and document object calls with React Testing Library or snapshot testing (chromatic).

**Closes #1495**

### DIFF
--- a/lib/src/select/Option.tsx
+++ b/lib/src/select/Option.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from "react";
+import React from "react";
 import styled from "styled-components";
 import { OptionProps } from "../select/types";
 import DxcCheckbox from "../checkbox/Checkbox";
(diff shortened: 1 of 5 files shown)


lib/src/select/Option.tsx
```diff
@@ -13,41 +13,49 @@ const Option = ({
...
-    selected={isSelected}
-    role="option"
-    aria-selected={isSelected}
-  >
-    <StyledOption
+}: OptionProps): JSX.Element => {
+  const labelRef = useRef<HTMLSpanElement>(null);
+
+  useEffect(() => {
+    if (labelRef.current.scrollWidth > labelRef.current.clientWidth) document.getElementById(id).title = option.label;
```
Jialecl, line 20: The first time opening a select labelRef.current.clientWidth can be 0.

Jialecl:
It seems that the tooltip is not updated correctly whenever options are deselected.
image

Jialecl:
The font color of the selected values seems to change if the select has a placeholder and the list is opened.
image

lib/src/select/Option.tsx
GomezIvann: The `useEffect` executes after the component is rendered, how could that be possible?

Jialecl: image
image
These are screenshot from the first opening.

image
And this one after closing and reopening the list.

Jialecl (approved):

PR #1800, opened 2024-02-05
Handle scroll in Number Inputs

### LINKED ISSUE #460: [Number Input] Mouse wheel modifies the value of the component

Mil4n0r:
**Checklist**

**Description**
It is desired to stop the page scroll when the focus is on the number input and the cursor is hovering over the component. If these conditions are met, the value of the number input should be increased or decreased with the mouse wheel.

**Closes #460**


lib/src/text-input/TextInput.tsx
```diff
@@ -246,6 +246,12 @@ const DxcTextInput = React.forwardRef<RefType, TextInputPropsType>(
           break;
       }
     };
+    const handleWheel = useCallback((event) => {
```
GomezIvann, line 249: Can you type the event parameter, please?

GomezIvann (changes requested):

GomezIvann (approved):

PR #1815, opened 2024-02-09
New Badge component

### LINKED ISSUE #1812: New component: Badge
Create new Badge component taking into account the design.

https://www.figma.com/file/ld2CyC6umeONXpRCK6LWm1/Halstack-Design-Library-(WIP%E2%9A%A0%EF%B8%8F)?type=design&node-id=3153-4518&mode=design&t=cCv1Vh3swV32YuWM-4

raquelarrojo:
**Checklist**

**Description**
New `Badge` component implementation based on design: https://www.figma.com/file/ld2CyC6umeONXpRCK6LWm1/Halstack-Design-Library-(WIP%E2%9A%A0%EF%B8%8F)?type=design&node-id=3153-4518&mode=design&t=cCv1Vh3swV32YuWM-4.

Old badge has been replaced by the new one on `Nav tabs` and `Tabs` component. This has caused some changes in these components:
- When a tab is disabled, the notification is not shown (even it has one), since, from design view, it makes no sense that a tab that is disabled, has a notification.
- In tabs and nav tabs, badge notification limit is `99` (the badge's default one).

**Closes #1812**


lib/src/badge/Badge.tsx
```diff
@@ -1,47 +1,156 @@
...
+  size = "medium",
+}: BadgePropsType): JSX.Element => {
   return (
-    <ThemeProvider theme={colorsTheme.tabs}>
-      <StyledDxcBadge notificationText={notificationText} disabled={disabled}>
-        <span>{notificationText}</span>
-      </StyledDxcBadge>
-    </ThemeProvider>
+    <BadgeContainer label={label} mode={mode} color={color} size={size} title={title} aria-label={title}>
+      {(mode === "contextual" && (
```
jsuarezgonz, line 96: As we check this in several places, I think we could create a flag with the verification and use it in the component. Something like:
`const showContextual = mode ==="contextual"`

lib/src/badge/Badge.tsx
```diff
@@ -1,47 +1,156 @@
...
-  disabled: BadgePropsType["disabled"];
+const getColor = (mode, color) => {
+  return mode === "contextual" ? contextualColorMap[color].text : CoreTokens.color_white;
+};
+
+const getBackgroundColor = (mode, color) => {
+  return mode === "contextual" ? contextualColorMap[color].background : CoreTokens.color_red_700;
+};
+
+const getPadding = (mode, size) => {
```
jsuarezgonz, line 118: In these three functions(getColor, getBackgroundColor, and getPadding) we can remove the braces and 'return' and leave them in a single line.

lib/src/badge/Badge.tsx
```diff
@@ -1,47 +1,156 @@
...
+      {typeof label === "number" ? (label > notificationLimit ? "+" + notificationLimit : label) : label}
+    </LabelContainer>
+  );
+};
+
+const DxcBadge = ({
+  label,
+  title,
+  mode = "contextual",
+  color = "grey",
```
jsuarezgonz, line 89: The default value of the prop color should be calculated according to the mode (`color = mode === "contextual" ? "grey" : "red"`), if not, when mode ="notification" color would be 'grey' which is strange since they are red.

image

jsuarezgonz (changes requested):

lib/src/badge/Badge.tsx
raquelarrojo: Agree!

lib/src/badge/Badge.tsx
raquelarrojo: I limited the color prop to only contextual mode. Notification mode will not have color prop.

jsuarezgonz (approved):

Issue #1809, opened 2024-02-07
Some components not working inside `Dialog`


Mil4n0r:
**Describe the bug**

This bug seems to be related to #1515. The components that don't work as expected are `Dropdown` and `Calendar`.
- `Dropdown`: When pressing up and down keys, the focus gets stuck in the last item, no matter what.
- `Calendar`: When pressing it, it won't open.

It seems that the bug is related to the way focus events are handled, as both of those components are calling `element.focus()` at the same time as the `FocusLock`

**To Reproduce**

Example:

1. Create a new React app using Vite.
2. Install `@dxc-technology\halstack-react` package.
3. Put `DxcDropdown` or `DxcCalendar` inside a `DxcDialog`
5. Interact with the component inside the `DxcDialog`

**Expected behaviour**
Both components must work the same way they do when they are outside of a `Dialog`

PR #1817, opened 2024-02-12
Dialog popups interaction fix

### LINKED ISSUE #1515: Pop up components keyboard interactions inside a dialog
Right now closing a dropdown or a calendar using the `ESC` key also closes the dialog itself.

The popup component should be able to be closed without closing the dialog.

One of the possible solutions is implementing `event.stopPropagation()`, but this needs to be further investigated to see if it is the best solution and if there is any other keys that may have similar issues.

Mil4n0r:
**Checklist**

**Description**
Popups inside dialog were having unexpected behaviors when interacting with them. This has been fixed properly handling the keyboard events when wanting to close an inner popup and ensuring that Popups are being properly identified inside the DOM.

**Closes #1515, #1809**

### DIFF
--- a/lib/src/date-input/DateInput.tsx
+++ b/lib/src/date-input/DateInput.tsx
@@ -197,7 +197,6 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
                 align="end"
                 aria-modal={true}
                 onBlur={handleDatePickerOnBlur}
-                // onEscapeKeyDown={handleDatePickerEscKeydown}
                 onKeyDown={handleDatePickerEscKeydown}
                 avoidCollisions={false}
               >
--- a/lib/src/date-input/DateInput.tsx
+++ b/lib/src/date-input/DateInput.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useLayoutEffect, useRef, useEffect } from "react";
+import React, { useState, useLayoutEffect, useRef, useEffect, useContext } from "react";
 import dayjs from "dayjs";
 import styled, { ThemeProvider } from "styled-components";
 import useTheme from "../useTheme";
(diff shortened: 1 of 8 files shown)


lib/src/date-input/DateInput.tsx
```diff
@@ -191,7 +197,7 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
                 align="end"
                 aria-modal={true}
                 onBlur={handleDatePickerOnBlur}
-                onEscapeKeyDown={handleDatePickerEscKeydown}
```
Mil4n0r, line 194: This might seem redundant, however it is important because, from my testing, it seems that `onEscapeKeyDown` has a 'lower priority' than `onKeyDown`, making it being triggered *after* the parent's `onKeyDown`, which in our case closes the whole modal.

Jialecl:
I am not sure this is a good fix.
Is this better than using the stopPropagation?
This behavior needs to be added to the textInput and all future components that use the 'Escape' key.

lib/src/dialog/Dialog.stories.tsx
```diff
@@ -262,6 +265,42 @@ export const DialogWithInputs = () => (
   </ExampleContainer>
 );
 
+export const DialogWithPopups = () => {
```
Jialecl, line 268: This test does not accomplish anything.

Mil4n0r:
> I am not sure this is a good fix. Is this better than using the stopPropagation? This behavior needs to be added to the textInput and all future components that use the 'Escape' key.

As Iván pointed out [here](https://github.com/dxc-technology/halstack-react/issues/1515#issuecomment-1462302334), it is advisable to avoid using `stopPropagation` due to the potential risks associated with disabling event bubbling. In our case, while we are altering the way popup elements interact, we are careful not to prevent bubbling. Thus, we should not encounter dead zones within the UI, even with multiple nested menus.

I acknowledge that both this workaround and the use of  `stopPropagation` come with the drawback that these lines must be added to all components that require the usage of events with the Escape key (both existing and new ones). I will continue exploring alternative approaches to address this issue. However, right now, I am unable to identify a more suitable solution.

Mil4n0r:
It has been decided to split this PR in 2, one that fixes popups not working inside Dialog and the other one (which is still pending for approval from the rest of the members) will fix the interaction with the Escape key.

PR #1820, opened 2024-02-12
Fixing problems with NavTab inside website

Mil4n0r:
**Checklist**

**Description**
There was a problem where the NavTab with the style modifications was not covering the full width in the website.

It seems that it was due to using negative `z-index`, the Underbar was not showing, changing the values and removing the unneeded grid inside the website layout should work.

It is still needed to update the library used inside the website


lib/src/nav-tabs/NavTabs.tsx
```diff
@@ -76,8 +76,8 @@ const DxcNavTabs = ({ iconPosition = "top", tabIndex = 0, children }: NavTabsPro
   return (
     <ThemeProvider theme={colorsTheme.navTabs}>
       <NavTabsContainer onKeyDown={handleOnKeyDown} role="tablist" aria-label="Navigation tabs">
-        <NavTabsContext.Provider value={contextValue}>{children}</NavTabsContext.Provider>
         <Underline />
+        <NavTabsContext.Provider value={contextValue}>{children}</NavTabsContext.Provider>
```
GomezIvann, line 80: There's no need for this change.

lib/src/nav-tabs/Tab.tsx
```diff
@@ -93,7 +93,7 @@ const TabContainer = styled.div<{ active: TabProps["active"] }>`
   align-items: stretch;
   border-bottom: 2px solid ${(props) => (props.active ? props.theme.selectedUnderlineColor : 'transparent')};
   padding: 0.5rem;
-
+  z-index: 1;
```
GomezIvann, line 96: There's no need for this change.

GomezIvann (changes requested):
I've reviewed the PR and I think there's no need for so many changes. Removing the grid container from website (since there's no divider in there anymore) and also removing the `z-index: -1` from the NavTabs do the job.

lib/src/nav-tabs/NavTabs.tsx
Mil4n0r: This was done for the sake of homogeneity, as I realised that inside the normal `Tags`, the `Underline` was placed before the Context Provider, I know it makes no difference but I would like to follow the same structure

lib/src/nav-tabs/Tab.tsx
Mil4n0r: If we don't give the Tabs a higher z-index, the color of the `Underline` and the `border-bottom` of the Tab will be overlapped right? (Showing only the gray color)

lib/src/nav-tabs/NavTabs.tsx
GomezIvann: Yes, I understood the change but, from my point of view, I don't know why we should show the underline before the tabs in the DOM tree. I know it doesn't affect since it has an absolute position but, for me, it makes more sense to have it after the tabs.

lib/src/nav-tabs/Tab.tsx
GomezIvann: You are right!

GomezIvann (approved):

PR #1827, opened 2024-02-15
Changing the footer padding according to Design

### LINKED ISSUE #1814: Footer Componet - content padding
The documentation states that the `padding-right` and `padding-left` are both 32px, however, in the actual implementation, it is set to 36px.

image

kukoc:
**Checklist**

**Purpose**
Padding left and right was 36 instead of 32 as per designs
**Description**
hanging the padding value

**Closes #1814 **


lib/src/footer/Footer.tsx
```diff
@@ -90,7 +90,7 @@ const FooterContainer = styled.footer<{ margin: FooterPropsType["margin"]; mode?
   width: 100%;
   gap: ${(props) => (props?.mode === "default" ? "0px" : "32px")};
   @media (min-width: ${responsiveSizes.small}rem) {
-    padding: ${(props) => (props?.mode === "default" ? "24px 36px 24px 36px" : "12px 32px 12px 32px")};
+    padding: ${(props) => (props?.mode === "default" ? "24px 32px 24px 32px" : "12px 32px 12px 32px")};
```
Mil4n0r, line 93: We should use two values for the padding prop instead to make it more readable:
`padding: ${(props) => (props?.mode === "default" ? "24px 32px" : "12px 32px")};`

Mil4n0r (changes requested):

lib/src/footer/Footer.tsx
kukoc: Changed!

Mil4n0r (approved):

PR #1835, opened 2024-02-19
Added new `Divider` component

### LINKED ISSUE #1834: New component: Divider
Create new Badge component taking into account the design.

https://preview.uxpin.com/c59c7c54afe8009ba5082153d70cd86fb91be7e7#/pages/165565398/simulate/no-panels

Mil4n0r:
**Checklist**

**Description**
A new `Divider` component is required based on the following design and props: https://preview.uxpin.com/c59c7c54afe8009ba5082153d70cd86fb91be7e7#/pages/165565398/simulate/no-panels

**Closes #1834**


lib/src/divider/Divider.tsx
```diff
@@ -0,0 +1,34 @@
...
+);
+
+const StyledDivider = styled.div<DividerPropsType>`
+  ${({ orientation, weight, width, color }) => `
+    background-color: ${color === "default" ? CoreTokens.color_grey_400 : CoreTokens.color_grey_700};
+    ${
+      orientation === "horizontal" && `width: ${width === "fixed" ? "480px" : width === "Full-Width" ? "100%" : width}`
+    };
+    ${orientation === "horizontal" && `height: ${weight === "regular" ? "1px" : "2px"}`};
+    ${orientation === "vertical" && `height: ${width === "fixed" ? "240px" : width === "Full-Width" ? "100%" : width}`};
```
kukoc, line 29: I know this is API config discussion. If we accept vertical and horizontal, then maybe the prop width should be renamed as length.

lib/src/divider/types.ts
```diff
@@ -0,0 +1,22 @@
+export type Width = "fixed" | "30%" | "50%" | "80%" | "Full-Width";
```
kukoc, line 1: We should use fullWidth or fullLength

lib/src/divider/types.ts
```diff
@@ -0,0 +1,22 @@
...
+   */
+  weight?: "regular" | "strong";
+  /**
+   * Modifies the length of the divider
+   */
+  width?: Width;
+  /**
+   * Modifies the color of the divider
+   */
+  color?: "default" | "dark-grey";
```
kukoc, line 19: Following our conventions, replace by darkGrey

lib/src/divider/Divider.tsx
```diff
@@ -0,0 +1,34 @@
...
+    aria-orientation={orientation}
+    role="separator"
+  />
+);
+
+const StyledDivider = styled.div<DividerPropsType>`
+  ${({ orientation, weight, width, color }) => `
+    background-color: ${color === "default" ? CoreTokens.color_grey_400 : CoreTokens.color_grey_700};
+    ${
+      orientation === "horizontal" && `width: ${width === "fixed" ? "480px" : width === "Full-Width" ? "100%" : width}`
```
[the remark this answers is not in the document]
kukoc: Furthermore, if we want to be consistent with other Hasltack components, we should use fillParent instead of full-length

kukoc (changes requested):
I suggest to quickly prepare the component api using your template and share it to have a quick review from the team devs.

lib/src/divider/Divider.tsx
Mil4n0r: It has been decided that length/width is no longer part of the component

lib/src/divider/types.ts
Mil4n0r: It has been decided that length/width is no longer part of the component.

lib/src/divider/Divider.tsx
```diff
@@ -0,0 +1,29 @@
...
+  />
+);
+
+const StyledDivider = styled.hr<DividerPropsType>`
+  ${({ orientation, weight, color }) => `
+    border-color: ${color === "default" ? CoreTokens.color_grey_400 : CoreTokens.color_grey_700};
+    ${orientation === "horizontal" ? "width" : "min-height"}: 100%;
+    ${orientation === "horizontal" ? "height" : "width"}: 0px;
+    border-width: ${weight === "regular" ? "1px" : "2px"};
+    margin: 0px;
```
Mil4n0r, line 25: By default, HR includes a 0.5em margin at the beginning and at the end.

lib/src/divider/types.ts
```diff
@@ -0,0 +1,16 @@
...
+   */
+  orientation?: "horizontal" | "vertical";
+  /**
+   * Modifies the thickness of the divider.
+   */
+  weight?: "regular" | "strong";
+  /**
+   * Modifies the color of the divider
+   */
+  color?: "default" | "darkGrey";
```
GomezIvann, line 13: I don't think that the word `default` is appropriate here. Both values are inconsistent with each other. Do we consider something like: `"mediumGrey"` and `"darkGrey"`? Also, aren't we missing a light grey, like the one in the Select component (`color-grey-200`)?

Mil4n0r: This is currently under investigation by the designers.

rconcepcion3: Hello... my suggestion for this if we're following this line of thought is we use lightGrey, Grey, and darkGrey

Mil4n0r:
I also added a change at SideNav to match the colors used in the rest of the dividers.

lib/src/divider/types.ts
```diff
@@ -0,0 +1,16 @@
+type Props = {
+  /**
+   * The divider can be showed in horizontal or vertical
```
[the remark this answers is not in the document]
Mil4n0r: I have realised that there are other places in the code where showed is used for the past participle so I will change those as well.

lib/src/divider/types.ts
```diff
@@ -0,0 +1,16 @@
...
+   */
+  orientation?: "horizontal" | "vertical";
+  /**
+   * Modifies the thickness of the divider.
+   */
+  weight?: "regular" | "strong";
+  /**
+   * Modifies the color of the divider
```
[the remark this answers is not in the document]
Mil4n0r: Done!

lib/src/divider/types.ts
Mil4n0r: In the end it was decided that the default option will be called mediumGrey instead of just Grey for homogeneization purposes.

Mil4n0r (commented):
There is still a pending discussion regarding whether or not we will add a field to address accessibility scenarios where screen readers would want to detect the `Divider` if it is not only decorative but serves as a semantic separator.

It was proposed by @GomezIvann to add a new `decorative` field that sets aria-hidden and is set as true by default.

Mil4n0r:
It has been decided to add the decorative prop to the component API for accessibility purposes (regarding screen readers and divider visibility).

The component is now ready to review. We have to take into account that, after the component is added, it should be integrated into the existing parts of the code that use dividers (SideNav, Contextual Menu...)

GomezIvann (approved):

kukoc (approved):

PR #1838, opened 2024-02-19
Fixed bug in TextInputs when scrolling

Mil4n0r:
**Checklist**

**Description**
Fixes bug related to #1800 where, on scroll, text inputs where displaying NaN.
The solution has been adding the pre-condition of the field being a number field before applying the `handleWheel` callback.


GomezIvann (changes requested):
A few comments:

- Since you are conditionally adding an event listener, you should also check the same condition when cleaning the effect. If not, you are always removing a non-existing event.
- `handleWheel` is missing the dependencies `incrementNumber` and `decrementNumber` in the `useCallback`.
- Please, format the code before committing it.

lib/src/text-input/TextInput.tsx
```diff
@@ -248,12 +285,15 @@ const DxcTextInput = React.forwardRef<RefType, TextInputPropsType>(
           break;
       }
     };
-    const handleWheel = useCallback((event: WheelEvent) => {
-      if (document.activeElement === inputRef.current) {
-        event.preventDefault();
-        event.deltaY < 0 ? incrementNumber(inputRef.current.value) : decrementNumber(inputRef.current.value);
-      }
-    }, []);
+    const handleWheel = useCallback(
```
GomezIvann, line 288: Even though you are wrapping `handleWheel` in a `useCallback` hook, it will re-render anyway because increment and decrement functions are reassigned in each re-render.

GomezIvann (commented):
We can solve the `useEffect` dependencies issue splitting the `onWheel` functionality:

1. Use the `onWheel` event of the Input component to do the increment/decrement functionality (remember to remove the event.preventDefault, is a passive event):
```
    const handleNumberInputWheel = (event: React.WheelEvent<HTMLInputElement>) => {
      if (document.activeElement === inputRef.current)
        event.deltaY < 0 ? incrementNumber(inputRef.current.value) : decrementNumber(inputRef.current.value);
    };  
```
`onWheel={numberInputContext?.typeNumber === "number" ? handleNumberInputWheel : undefined}`

2. Remove the `useEffect` from the `TextInput` and move it to the `NumberInput`. This makes sense since this functionality should only be available in that component:
```
    const numberInputRef = React.useRef<HTMLInputElement>(null);

    useEffect(() => {
      const input = numberInputRef.current?.getElementsByTagName("input")[0] as HTMLInputElement;
      const preventDefault = (event: WheelEvent) => {
        event.preventDefault();
      };

      input?.addEventListener("wheel", preventDefault, { passive: false });
      return () => {
        input?.removeEventListener("wheel", preventDefault);
      };
    }, []);

    return (
      <NumberInputContext.Provider value={{ typeNumber: "number", minNumber: min, maxNumber: max, stepNumber: step }}>
        <NumberInputContainer ref={numberInputRef}>
```

What do you think?

lib/src/text-input/TextInput.tsx
Mil4n0r: You are right.

Mil4n0r:
> We can solve the `useEffect` dependencies issue splitting the `onWheel` functionality:
> 
> 1. Use the `onWheel` event of the Input component to do the increment/decrement functionality (remember to remove the event.preventDefault, is a passive event):
> 
> ```
>     const handleNumberInputWheel = (event: React.WheelEvent<HTMLInputElement>) => {
>       if (document.activeElement === inputRef.current)
>         event.deltaY < 0 ? incrementNumber(inputRef.current.value) : decrementNumber(inputRef.current.value);
>     };  
> ```
> 
> `onWheel={numberInputContext?.typeNumber === "number" ? handleNumberInputWheel : undefined}`
> 
> 2. Remove the `useEffect` from the `TextInput` and move it to the `NumberInput`. This makes sense since this functionality should only be available in that component:
> 
> ```
>     const numberInputRef = React.useRef<HTMLInputElement>(null);
> 
>     useEffect(() => {
>       const input = numberInputRef.current?.getElementsByTagName("input")[0] as HTMLInputElement;
>       const preventDefault = (event: WheelEvent) => {
>         event.preventDefault();
>       };
> 
>       input?.addEventListener("wheel", preventDefault, { passive: false });
>       return () => {
>         input?.removeEventListener("wheel", preventDefault);
>       };
>     }, []);
> 
>     return (
>       <NumberInputContext.Provider value={{ typeNumber: "number", minNumber: min, maxNumber: max, stepNumber: step }}>
>         <NumberInputContainer ref={numberInputRef}>
> ```
> 
> What do you think?

This looks like a much more logical approach, encapsulating the NumberInput functionality encapsulated within the appropiate component and reducing all the unnecessary dependencies inside the TextInput component.

Thanks for your time 😄

GomezIvann (approved):

PR #1847, opened 2024-02-23
Dropdown theme overwrite in actionsCell

Jialecl:
**Checklist**

**Purpose**
This change allows to overwrite the dropdown theme with the actionIcon theme from the table.
image


raquelarrojo (changes requested):
I would add a visual test to check that the theme is being applied correctly (another one different from the dxc theme)

lib/src/table/DropdownTheme.jsx
```diff
@@ -1,14 +1,16 @@
...
     buttonPaddingTop: "4px",
     buttonPaddingBottom: "4px",
     buttonPaddingLeft: "4px",
     buttonPaddingRight: "4px",
     buttonHeight: "24px",
     buttonBorderRadius: "2px",
+    buttonBorderStyle: "none",
+    buttonBorderThickness: "0px",
     buttonBorderColor: "transparent",
     optionFontSize: "14px",
```
raquelarrojo, line 15: `optionFontSize` is repeated with different values.

raquelarrojo (changes requested):

raquelarrojo (changes requested):
This error is thrown in the table stories. 
image

lib/src/table/DropdownTheme.jsx
```diff
@@ -1,14 +1,16 @@
 import { componentTokens } from "../common/variables";
 export default {
   dropdown: {
-    // Breadcrumbs tokens
+    // ActionIcon size tokens
     buttonIconSize: "16px",
```
raquelarrojo, line 5: Some of this tokens are different from the dropdown component tokens. Which is the reason of this difference? Do we have a design for this?
In comments, we have the dropdown componentTokens values:
image

raquelarrojo (changes requested):

raquelarrojo (changes requested):
`ActionsCell` is used both in table and resultset table but is only inside `Table` folder. I think we should find a better common place.

raquelarrojo (approved):

PR #1848, opened 2024-02-26
Added specs and usage doc to StatusLight component

Mil4n0r:
**Checklist**

**Description**
Completed documenting StatusLight component


website/screens/components/status-light/usage/StatusLightUsagePage.tsx
```diff
@@ -0,0 +1,118 @@
...
+        ),
+      },
+    ],
+  },
+  {
+    title: "Variants",
+    content: (
+      <>
+        <DxcParagraph>
+          The Status light component has five different modes, each one
```
GomezIvann, line 74: For consistency purposes, I think we should keep the same approach as with the [Button variants](https://developer.dxc.com/halstack/next/components/button/usage/#variants) or the [Alert variants](https://developer.dxc.com/halstack/next/components/alert/usage/).

1. Use a table with a row for each variant.
2. Instead of using an image, I think it's better to use coded examples as in the rest of the Usage pages.

website/screens/components/status-light/specs/StatusLightSpecsPage.tsx
```diff
@@ -0,0 +1,47 @@
...
+    content: (
+      <>
+        <Image src={anatomyImage} alt="Status Light anatomy" />
+        <DxcBulletedList type="number">
+          <DxcBulletedList.Item>Status Light</DxcBulletedList.Item>
+          <DxcBulletedList.Item>Label</DxcBulletedList.Item>
+        </DxcBulletedList>
+      </>
+    ),
+  },
```
GomezIvann, line 30: Should we add a Design Tokens section specifying that it doesn't have tokens? It's a section present on every specifications page.

GomezIvann: Did Lara said anything about a possible accessibility section?

Mil4n0r: I will ask for her opinion on both topics

Mil4n0r: She thinks that we should add the design tokens section but not the accessibility one (at least for now).

website/screens/components/status-light/usage/StatusLightUsagePage.tsx
```diff
@@ -0,0 +1,161 @@
...
+          corresponding to its semantic meaning:
```
GomezIvann, line 76: This sentence is a bit odd. Having the two dots and the line jump to again have two dots. I see two possibilities here:

1. Remove the first two dots and end the sentence with a single dot.
2. Join both sentences: "The Status light component has five different modes, each one corresponding to its semantic meaning: **default**, **info**, **success**, **warning** and **error**."

GomezIvann (changes requested):

website/screens/components/status-light/usage/StatusLightUsagePage.tsx
Mil4n0r: You are right, I added the first option in order to follow the same structure as the `Button` Usage.

GomezIvann (approved):

PR #1850, opened 2024-02-26
New Contextual Menu component

GomezIvann:
**Checklist**

- [x] The Build process is done without errors. All tests pass in the `/lib` directory.

**Description**
Implementation of the new component `DxcContextualMenu`. Based on [Lara's Designs](https://www.figma.com/file/ld2CyC6umeONXpRCK6LWm1/Halstack-Design-Library-(WIP%E2%9A%A0%EF%B8%8F)?type=design&node-id=3744-5731&mode=design&t=lcXwCQJ3IL98exxa-4).

**Component definition**
This component is a collection of actions that modify in some way the current Page where it is been displayed. Don't confuse this component with the Sidenav, which is a navigational component at the Application level.


lib/src/contextual-menu/MenuItem.tsx
```diff
@@ -0,0 +1,82 @@
+import React from "react";
+import styled from "styled-components";
+import CoreTokens from "../common/coreTokens";
+import { MenuItemProps } from "./types";
+
+const MenuItem = ({ label, icon, slot, selected, onSelect }: MenuItemProps) => (
+  <Action role="menuitem" aria-selected={selected} selected={selected} onClick={() => onSelect()}>
+    <Label>
+      {icon && <Icon aria-hidden>{typeof icon === "string" ? <img src={icon} /> : icon}</Icon>}
```
Mil4n0r, line 9: Missing alt text

lib/src/contextual-menu/ContextualMenu.tsx
```diff
@@ -0,0 +1,113 @@
...
+import MenuPropsType, { Item, Section as SectionType, BadgeProps as BadgePropsType } from "./types";
+import DxcBadge from "../badge/Badge";
+import MenuItem from "./MenuItem";
+
+const DxcContextualMenu = ({ items, defaultSelectedItemIndex = -1 }: MenuPropsType) => {
+  const [selectedItemIndex, setSelectedItemIndex] = useState<number>(defaultSelectedItemIndex);
+
+  const renderSingleItem = (item: Item, index: number) => (
+    <Li key={`option-${index}`}>
+      <MenuItem
```
Mil4n0r, line 13: I think that role="menuitem" must be placed at `Li` level rather than in the inner level.

lib/src/contextual-menu/ContextualMenu.tsx
```diff
@@ -0,0 +1,113 @@
...
+    </Li>
+  );
+
+  let accLength = 0;
+  const renderSection = (section: SectionType, currentSectionIndex: number, items: SectionType[]) => {
+    const startingIndex = accLength;
+    accLength += section.items.length;
+    return (
+      <Fragment key={`separator-${currentSectionIndex}`}>
+        <Li key={`option-${currentSectionIndex}`}>
```
Mil4n0r, line 30: I think that role="group" must be placed at `Li` level rather than in the inner level.

lib/src/contextual-menu/ContextualMenu.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+const sections = [
+  {
+    title: "Team repositories",
+    items: [{ label: "Approved locations" }, { label: "Approved locations" }, { label: "Approved locations" }],
+  },
+  {
+    items: [{ label: "Approved locations" }, { label: "Approved locations" }, { label: "Approved locations" }],
+  },
+];
```
Mil4n0r, line 15: Slots could be added to some of the items (badges and SVGs).

lib/src/contextual-menu/ContextualMenu.test.js
```diff
@@ -0,0 +1,44 @@
...
+      {
+        label: "Tested item",
+        onSelect: jest.fn(),
+      },
+    ];
+    const { getByRole } = render(<DxcContextualMenu items={test} />);
+    const item = getByRole("menuitem");
+    fireEvent.click(item);
+    expect(test[0].onSelect).toHaveBeenCalled();
+  });
```
Mil4n0r, line 43: The following tests should be added:
- Keyboard navigation (Arrow keys, focus and selection).
- Edge cases (Empty menu, menu with only 1 item, default selected item out of bounds...).

Mil4n0r (changes requested):

lib/src/contextual-menu/ContextualMenu.test.js
GomezIvann: I follow your structure to answer:

- There's no focus management or functionality from our side. This is the usual browser tab sequence, so there is no need to test in this case.
- Since we document these scenarios and type the props to avoid those edge cases, I consider that there is no need to over-program. If the component breaks the user should know that it is his fault (this is our approach since we added Typescript).

lib/src/contextual-menu/ContextualMenu.test.js
GomezIvann: Do they add anything relevant to the unit tests? They are checked on the storybooks.

lib/src/contextual-menu/MenuItem.tsx
GomezIvann: This is a hidden icon, why should we add an `alt` text?

lib/src/contextual-menu/ContextualMenu.tsx
GomezIvann: Agreed Enrique! I uploaded the changes.

lib/src/contextual-menu/MenuItem.tsx
Mil4n0r: I thought that using an empty string for `alt` (alongiside `aria-hidden`) was a good practice in terms of accessibility even if the images were decorative.

After reading a bit about it, it is supposed to be unnecesary, the only advantage might be future additional compatibility with W3C validation tools, so it is totally up to you.

Sorry for the inconvenience.

Mil4n0r (approved):

raquelarrojo (approved):

PR #1854, opened 2024-02-29
Fixing dependency issues

Jialecl:
**Checklist**

**Purpose**
This PR fixes an issue related to dependencies import in storybook.


Mil4n0r (changes requested):
When I am updating the storybook in local while being inside the 'Table' story, I am getting the following error.
image
These kind of errors are usually caused by circular dependencies, so I detected the following circular dependencies that may be causing inconsistent behaviour inside the storybook:
graph
The problem is that, in multiple components, the internal components used are imported from 'main' instead of their appropiate directory.

If we change this:
`import { DxcDropdown, DxcFlex, HalstackProvider } from "../main";`

For this:
```
import DxcDropdown from "../dropdown/Dropdown";
import DxcFlex from "../flex/Flex";
import { HalstackProvider } from "../HalstackContext";

The issue will be fixed.

We can make a separate PR to address the rest of the circular dependencies or fix them in this PR, it's up to you!
```

Mil4n0r (dismissed):
Now it is working correctly for my end.

Waiting for @raquelarrojo's confirmation

NOTE: 
There are still circular dependencies which should be addressed in a separate PR.
066b438d-d997-4a93-bec2-2407af549562

PR #1858, opened 2024-03-04
Removed circular dependencies

Mil4n0r:
**Checklist**

**Description**
While reviewing #1854, we discovered that there were multiple circular dependencies inside the project which, in some scenarios, could cause issues.

In this PR, we are fixing the remaining ones that are:

```
src/number-input/NumberInput.tsx <--> src/text-input/TextInput.tsx
src/nav-tabs/NavTabs.tsx <--> src/nav-tabs/Tabs.tsx
src/accordion-group/AccordionGroup.tsx <--> src/accordion-group/AccordionGroupAccordion.tsx
```

To address those issues, I have moved the logic related to the respective contexts (which were the ones causing the circular dependencies) outside the component file, following the same pattern that was used for `SideNavContext`.


Mil4n0r:
> The context files should not be located in the layout folder.

I agree, I did it because there was already the SideNav context there, but it does not make that much sense.

We should discuss what the appropiate location would be. In my opinion the lib/src/[component]/ directory is appropiate (maybe adding a context/ additional folder inside).

lib/src/accordion-group/AccordionGroup.tsx
```diff
@@ -4,9 +4,8 @@ import { getMargin } from "../common/utils";
 import { spaces } from "../common/variables";
 import useTheme from "../useTheme";
 import AccordionGroupAccordion from "./AccordionGroupAccordion";
-import AccordionGroupPropsType, { AccordionGroupAccordionContextProps } from "./types";
-
-export const AccordionGroupAccordionContext = createContext<AccordionGroupAccordionContextProps | null>(null);
+import AccordionGroupPropsType from "./types";
+import { AccordionGroupAccordionContext } from "../layout/AccordionGroupContext";
```
Jialecl, line 8: I think this case is still considered as a circular dependency.
Context imported by AccordionGroup and AccordionGroup.Accordion but AccordionGroup is also imported into AccordionGroup.Accordion.

lib/src/nav-tabs/NavTabs.tsx
```diff
@@ -1,10 +1,9 @@
-import React, { useState, ReactElement, createContext, useMemo } from "react";
+import React, { useState, ReactElement, useMemo } from "react";
 import styled, { ThemeProvider } from "styled-components";
 import useTheme from "../useTheme";
-import NavTabsPropsType, { NavTabsContextProps } from "./types";
+import NavTabsPropsType from "./types";
 import DxcTab from "./Tab";
-
-export const NavTabsContext = createContext<NavTabsContextProps | null>(null);
+import { NavTabsContext } from "../layout/NavTabsContext";
```
Jialecl, line 6: This is the same scenario as the AccordionGroup.

Jialecl:
> We should discuss what the appropiate location would be. In my opinion the lib/src/[component]/ directory is fine(maybe adding a context/ additional folder inside).

I think in the  `lib/src/[component]/` is the correct path there is no need to add an additional level to the path.

lib/src/accordion-group/AccordionGroup.tsx
Mil4n0r: image
As far as I know this is not considered a circular dependency, we will have to discuss it with the rest of the team.

Jialecl (approved):

PR #1860, opened 2024-03-05
Icon component

Jialecl:
**Checklist**

**Description**
Created a new icon component to be used internally.


lib/src/icon/Icon.tsx
```diff
@@ -0,0 +1,38 @@
...
+  display: inline-block;
+  white-space: nowrap;
+  word-wrap: normal;
+  direction: ltr;
+  -webkit-font-feature-settings: "liga";
+  -webkit-font-smoothing: antialiased;
+  font-variation-settings: ${(props) => (props.filled ? "'FILL' 1" : "'FILL' 0")};
+  ::before {
+    content: "${(props) => props.icon}";
+  }
```
Mil4n0r, line 35: The following rules could be added for extra browser compatibility
```
/* Support for Safari and Chrome. */
text-rendering: optimizeLegibility;

/* Support for Firefox. */
-moz-osx-font-smoothing: grayscale;

/* Support for IE. */
font-feature-settings: 'liga';
```
[Google fonts](https://developers.google.com/fonts/docs/material_icons?hl=es-419)

lib/src/icon/Icon.tsx
```diff
@@ -0,0 +1,38 @@
...
+import IconPropsType from "./types";
+
+const DxcIcon = ({ icon }: IconPropsType): JSX.Element => {
+  let filled = false;
+  let iconName = icon;
+  if (icon.startsWith("filled_")) {
+    filled = true;
+    iconName = icon.replace(/filled_/g, "");
+  }
+  return <IconContainer role="img" aria-label={icon} filled={filled} icon={iconName} />;
```
Mil4n0r, line 12: In terms of accessibility, this is not accurate, as the role would depend on the purpose of the icon.
I would consider those scenarios:
- If the icon is **purely decorative** and does not convey any meaningful information, it can have a `role="presentation"` and `aria-hidden="true"` to indicate screen readers that it should be ignored.
- If the icon is clickable and **performs an action when activated**, it can have a `role="button"`.
- If the icon acts as a link to **navigate to another page or resource**, it can have a `role="link"`.
- Only if the icon is **purely informative** and provides important context or content to the user, it can have a `role="img"`.

Taking this into account, maybe it would be interesting to add a parameter in a similar way as it was done for the `Divider` component to decide what is the most appropiate role (being default to decorative, which will probably be the most common scenario).

Mil4n0r (changes requested):

lib/src/icon/Icon.tsx
Jialecl: Based on WAI ARIA role should always be "img" in the case of a font representing an icon.
I agree about `aria-hidden=true `, which we can consider being the case for all of our use cases.

See [WAI ARIA doc](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA24).

Mil4n0r: Great info!

Mil4n0r (approved):
It has been decided that jest testing would be redundant as all the possible tests are already covered by chromatic and, soon, accessibility testing with axe. (Role does not have to be checked due to aria-hidden invalidating it).

PR #1868, opened 2024-03-07
Button Icon documentation updated

Jialecl:
**Checklist**

**Purpose**
Button documentation updated to include the new icon.

### DIFF
--- a/website/screens/components/button/code/ButtonCodePage.tsx
+++ b/website/screens/components/button/code/ButtonCodePage.tsx
@@ -85,8 +85,14 @@ const sections = [
               </TableCode>
             </td>
             <td>
-              Element or path used as the icon that will be placed next to the
-              button label.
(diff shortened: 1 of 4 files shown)


website/screens/components/button/code/ButtonCodePage.tsx
```diff
@@ -85,8 +85,14 @@ const sections = [
               </TableCode>
             </td>
             <td>
-              Element or path used as the icon that will be placed next to the
-              button label.
+              <DxcLink newWindow href="https://fonts.google.com/icons">
```
kukoc, line 88: I think we should link to our internal library. https://app.uxpin.com/design-system/670ac8ca647bace04103/iconography

We already explain in the iconography principles documentation that we are using a subset of Material symbols. But in the rest of our documentation, we should stick to our internal library

kukoc (changes requested):

kukoc (approved):

PR #1878, opened 2024-03-12
🚀 Create text-In-between.ts

HanojHanu:
image


lcordobapuerto:
Hi there @HanojHanu !
Could you please show us where exactly you want to use this and why?
Thank you!

HanojHanu:
image

lcordobapuerto:
Thanks for the example @HanojHanu! 
In the Use case you just proposed, the divider is not necessary. Is an acceptable use to have the "Sign in with LinkedIn" following the Sign in button. The divider does not add any value to it.

PR #1887, opened 2024-03-12
Using DxcIcons in NavTabs

kukoc:
**Checklist**

### DIFF
--- a/lib/src/nav-tabs/Tab.tsx
+++ b/lib/src/nav-tabs/Tab.tsx
@@ -52,7 +52,7 @@ const DxcTab = forwardRef(
           {...otherProps}
         >
           {icon && (
-            <TabIconContainer iconPosition={iconPosition}>
+            <TabIconContainer iconPosition={iconPosition} active={active} disabled={disabled}>
               {typeof icon === "string" ? <DxcIcon icon={icon} /> : icon}
             </TabIconContainer>
           )}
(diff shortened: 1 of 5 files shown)


lib/src/nav-tabs/Tab.tsx
```diff
@@ -149,11 +137,31 @@ const Tab = styled.a<{
 
 const TabIconContainer = styled.div<{ iconPosition: NavTabsPropsType["iconPosition"] }>`
   display: flex;
+  font-size: 24px;
+  color {
```
Jialecl, line 141: Color is not being applied since it is not a selector.

lib/src/nav-tabs/Tab.tsx
```diff
@@ -149,11 +137,31 @@ const Tab = styled.a<{
...
+    ${(props) => props.theme.disabledIconColor};
+  }
+  &[aria-selected="true"] {
```
Jialecl, line 144: Instead of using the aria pass the selected prop to be used also in the styled component.

kukoc: Chenged!

Jialecl: color styling can be removed from svg and  &[aria-selected="true"] since we have it already applied in the parent element.

Jialecl: the props added to the declaration of the styled component also need to be added above when the styled component is used.

Jialecl (approved):

PR #1891, opened 2024-03-13
Jsuarezgonz select icon

jsuarezgonz:
**Checklist**

**Purpose**
Icon added to the Select component.

### DIFF
--- a/lib/src/select/Listbox.tsx
+++ b/lib/src/select/Listbox.tsx
@@ -3,7 +3,7 @@ import styled from "styled-components";
 import useTranslatedLabels from "../useTranslatedLabels";
 import { ListboxProps } from "./types";
 import Option from "./Option";
-import selectIcons from "./Icons";
+import DxcIcon from "../icon/Icon";
 
 const groupsHaveOptions = (options) =>
   options?.[0].options ? options.some((groupOption) => groupOption.options?.length > 0) : true;
@@ -103,7 +103,9 @@ const Listbox = ({
     >
       {searchable && (options.length === 0 || !groupsHaveOptions(options)) ? (
         <OptionsSystemMessage>
-          <NoMatchesFoundIcon>{selectIcons.searchOff}</NoMatchesFoundIcon>
+          <NoMatchesFoundIcon>
+            <DxcIcon icon="search_off" />
+          </NoMatchesFoundIcon>
           {translatedLabels.select.noMatchesErrorMessage}
         </OptionsSystemMessage>
       ) : (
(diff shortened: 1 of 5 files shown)


lib/src/select/Listbox.tsx
```diff
@@ -161,6 +163,9 @@ const NoMatchesFoundIcon = styled.span`
   width: 16px;
   padding: 4px;
   margin-right: 0.25rem;
+  & > span {
```
Jialecl, line 166: There is no need for this selector right?

lib/src/select/Option.tsx
```diff
@@ -134,7 +137,7 @@ const OptionSelectedIndicator = styled.span`
   display: flex;
   align-items: center;
   color: ${(props) => props.theme.selectedListOptionIconColor};
-
+  font-size: 16px;
   svg {
```
Jialecl, line 141: The svg selector can be removed it will never be a svg element.

lib/src/select/Select.tsx
```diff
@@ -457,7 +457,11 @@ const DxcSelect = React.forwardRef<RefType, SelectPropsType>(
                     </SelectedOption>
                   )}
                 </SearchableValueContainer>
-                {!disabled && error && <ErrorIcon>{selectIcons.error}</ErrorIcon>}
+                {!disabled && error && (
+                  <ErrorIcon>
```
Jialecl, line 461: The svg selector can be removed and the font-size needs to be set.

lib/src/select/Option.tsx
```diff
@@ -104,7 +107,6 @@ const OptionIcon = styled.span<{ grouped: OptionProps["isGroupedOption"]; multip
   margin-left: ${(props) => (props.grouped && !props.multiple ? "16px" : "8px")};
   color: ${(props) => props.theme.listOptionIconColor};
 
-  svg,
   img {
```
Jialecl, line 110: The OptionIcon in Option.tsx still uses the img element.

lib/src/select/types.ts
```diff
@@ -22,7 +22,7 @@ export type OptionGroup = {
 export type Option = {
   /**
    * Element used as the icon that will be placed before the option label.
-   * It can be a url of an image or an inline SVG. If the url option
+   * It can be a url of an image, an inline SVG or Material Symbol name. If the url option
```
Jialecl, line 25: remove 'url of an image,'

Jialecl (approved):

PR #1906, opened 2024-03-15
Replace the Status tag with our Badge

raquelarrojo:
**Checklist**

**Description**
Replace the Status tag with our Badge component.
We will need to change the status of several components and props after this change and before creating the release.


website/screens/common/sidenav/SidenavLogo.tsx
Mil4n0r, line 1: Why not adding StatusBadge to match the rest of the components? (With the "Experimental" status)

raquelarrojo: Because the sidenav one is not showing status, it is showing the version number/dev (if we are in localhost). I could have changed the status badge for doing this also, but I thinks it is better to have the status badge only for the status and not to difficult its logic. So here, we only have our badge.

Mil4n0r: Sounds logical. Thank you.

Mil4n0r (approved):

Jialecl (approved):

Issue #1666, opened 2023-08-02
Date Input size fillParent does not work in a flex container


Jialecl:
**Describe the bug**
When the Date Input with `size="fillParent"` is placed inside a flex container it does not take the whole width as it should.
It only works correctly when the flex container it is also `direction="column"`.

**Screenshots**
![image](https://github.com/dxc-technology/halstack-react/assets/12238686/419ae3be-e924-465d-a2e2-ea3f3f69b2dd)

**Additional context**
It might be due to the `div` that contains the date input.

PR #1919, opened 2024-03-21
Added accessibility fixes for DateInput component

### LINKED ISSUE #1944: Review accessibility for DateInput component
There are currently some accessibility issues regarding the incorrect use of aria-selected in a button, inappropiate roles inside the Calendar, etc.

Right now, a custom set of rules is being used to prevent errors, but the warnings must be addressed.

Mil4n0r:
**Checklist**

**Description**
There were some accessibility issues regarding the incorrect use of aria-selected in a `button`, inappropiate roles inside the Calendar, etc.

IMPORTANT: This task required modifying the existing structure of the component so I would highly recommend to wait for the next release to merge it. Feel free to review it though.

**Closes #1944**


Mil4n0r:
I think that it would be very helpful to add a `role` prop to our current `DxcFlex` component to make it more comfortable to meet requirements.
For example, in the current state of this PR, using `DxcFlex` limits us to not be able to use role="row" for the header column (the one that has the name of the weekdays).

lib/src/date-input/Calendar.tsx
```diff
@@ -153,38 +162,45 @@ const Calendar = ({
         break;
     }
   };
-
   return (
-    <CalendarContainer>
+    <CalendarContainer role="grid">
       <DxcFlex alignItems="center" justifyContent="space-between">
```
Mil4n0r, line 167: I think that it would be very helpful to add a `role` prop to our current `DxcFlex` component to make it more comfortable to meet requirements.
In the current state of this PR, using `DxcFlex` limits us to not be able to use role="row" for the header column (the one that has the name of the weekdays).

Mil4n0r: It has been decided that DxcFlex should not have a role prop, so another implementation was done instead.

Mil4n0r:
Reopened.

Mil4n0r:
Updated, now also fixes #1666

lib/src/date-input/Calendar.tsx
```diff
@@ -39,22 +39,29 @@ const getDateToFocus = (selectedDate, innerDate, today) => {
...
+      ? today
+      : innerDate.set("date", 1);
 };
 
 const isDaySelected = (date: { day: number; month: number; year: number }, selectedDate) =>
   selectedDate?.get("month") === date.month &&
   selectedDate?.get("year") === date.year &&
   selectedDate?.get("date") === date.day;
 
+const chunk = (data: any[], chunkSize: number) => {
```
Jialecl, line 51: The names feel generic and makes this function difficult to understand.

lib/src/date-input/Calendar.tsx
```diff
@@ -153,38 +161,46 @@ const Calendar = ({
         break;
     }
   };
-
   return (
-    <CalendarContainer>
-      <DxcFlex alignItems="center" justifyContent="space-between">
+    <CalendarContainer role="grid">
+      <CalendarRow role="row">
```
Jialecl, line 166: DxcFlex is no longer used the component can be removed.

lib/src/date-input/Calendar.tsx
```diff
@@ -153,38 +161,46 @@ const Calendar = ({
...
-              today.get("date") === date.day &&
-              today.get("month") === innerDate.get("month") &&
-              today.get("month") === date.month &&
-              today.get("year") === innerDate.get("year")
-            }
-          >
-            {date.day}
-          </DayCell>
+      </CalendarRow>
+      <WeeksContainer onBlur={handleOnBlur} role="rowgroup">
```
Jialecl, line 173: We should review the naming of the containers. I think they don't describe correctly what they represent.

lib/src/date-input/DateInput.stories.tsx
```diff
@@ -202,22 +189,22 @@ const DatePickerButtonStates = () => {
           }}
         >
           <DxcDateInput label="From" defaultValue="01-12-1995" />
-          <DxcDateInput label="To" />
+          <DxcDateInput label="To" defaultValue="01-12-1995" />
```
Jialecl, line 192: Why was a defaultValue added?

lib/src/date-input/Calendar.tsx
Mil4n0r: Yup, you are totally right, I might try something like `divideDaysIntoWeeks`

lib/src/date-input/Calendar.tsx
Mil4n0r: Done! Take a look at it when possible.

lib/src/date-input/DateInput.tsx
```diff
@@ -212,4 +228,8 @@ const StyledPopoverContent = styled(Popover.Content)`
   }
 `;
 
+const DateInputContainer = styled.div<{ margin: DateInputPropsType["margin"]; size: DateInputPropsType["size"] }>`
+  width: ${(props) => calculateWidth(props.margin, props.size)};
```
Jialecl, line 232: There is no need to use calculate width or the size.
` ${(props) => props.size == "fillParent" && "width: 100%;"}`
If 1005 is applied to the width when the size is fillParent fixes the weird behavior.

lib/src/date-input/Calendar.tsx
```diff
@@ -210,15 +228,24 @@ const WeekHeaderCell = styled.span`
   height: 36px;
 `;
 
-const DayCellsContainer = styled.div`
+const DayCellContainer = styled.span``;
```
Jialecl, line 231: DayCellContainer is no longer used.

Mil4n0r: As you proposed, the container with the `gridcell` role should not be needed, the error thrown by axe-core is listed as a false positive (https://github.com/microsoft/fluentui/issues/30305#issuecomment-1889892108) So I added a custom rule for Date Input and added the role directly to the button.

Jialecl (approved):

PR #1924, opened 2024-03-22
Multiselect keyboard interaction

### LINKED ISSUE #861: Review keyboard accessibility for select component clear actions

Mil4n0r:
**Checklist**

**Description**
Previously, when using the multiple select, there was no way for only-keyboard users to clear the selection. According to Combobox Pattern from WAI-ARIA, it is expected that, if the popup is not visible, the Esc key should clear the options.
[Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)

**Closes #861**


Mil4n0r:
I don't think that this should be merged before releasing the version 12.

GomezIvann (commented):
We still have the same problem as at the beginning, the clear action is not reachable by the keyboard. When screen readers read the component, they will inform the user of the existence of a clear button that can clear all selected values but can never be accessed by tabs or arrows. I don't think it is intuitive for them to clear the value with the Esc key, as it is not standard knowledge, besides the fact that we tell them about the existence of a button for the same thing.

The solution here lies in a new design for some sort of internal/special focus for the component. Look at how it is done by [React Select](https://react-select.com/home):

image

GomezIvann:
> > We still have the same problem as at the beginning, the clear action is not reachable by the keyboard. When screen readers read the component, they will inform the user of the existence of a clear button that can clear all selected values but can never be accessed by tabs or arrows. I don't think it is intuitive for them to clear the value with the Esc key, as it is not standard knowledge, besides the fact that we tell them about the existence of a button for the same thing.
> > The solution here lies in a new design for some sort of internal/special focus for the component. Look at how it is done by [React Select](https://react-select.com/home):
> > image
> 
> You are right, however in the approach that you have shared, I have seen that the "clear all" button is not reachable either (it uses aria-hidden and focusable=false, so it also has a problem with the keyboard interaction.
> 
> I see 3 possible options here:
> 
> 1. Maintain the component as it is but add the `aria-hidden=true` and `tabIndex=-1` to the close button.
> 2. Maintain the component and add some internal focus management (could lead to accessibility problems due to nested interaction but they probably can be fixed)
> 3. Same as 2 but removing the special interaction added for the Escape key (I would not do this due to this behavior being included in WCAG 2 as a recommended way to interact).
> 
> Probably the option 2 is the most complete one. Let me know what you think.

The example I shared was to show a possible visual solution. What I meant with this is that we may need some sort of solution from our designers and investigate this a little bit further than your proposal with this PR.

Also, I agree that the second point is probably the best approach. We need to investigate this and talk with a designer.

Mil4n0r:
I will try to discuss with @lcordobapuerto regarding this topic today.

Thank you for the inputs!

Mil4n0r:
Lara does not seem to approve opting for the Escape key alternative to clear due to possible misinputs so we would probably have to discuss it with the rest of the team.

On the other hand, she likes the approach that where there is an internal focus management so I will try to take a look at it.

Meanwhile, I will close the PR to reduce confusion.

Mil4n0r:
Carbon Design, which has a similar structure in their `Dropdown` uses the "Escape" key approach.

https://carbondesignsystem.com/components/dropdown/code

I agree that the best strategy would be to handle a focus that can target every selected item and the "Clear all options" button. However, that would require a redesign which is not possible due to other tasks having a higher priority.

So it is up to you if you want to add this PR with only the "Escape" workaround, maybe adding some Accessibility section to explain the behavior or we simply leave it for now until redesign.

GomezIvann:
> Carbon Design, which has a similar structure in their `Dropdown` uses the "Escape" key approach.
> 
> https://carbondesignsystem.com/components/dropdown/code
> 
> I agree that the best strategy would be to handle a focus that can target every selected item and the "Clear all options" button. However, that would require a redesign which is not possible due to other tasks having a higher priority.
> 
> So it is up to you if you want to add this PR with only the "Escape" workaround, maybe adding some Accessibility section to explain the behavior or we simply leave it for now until redesign.

There's no urgency in providing a solution for this topic Enrique, no one has requested us this feature so we can stick with Lara's plan.

PR #1925, opened 2024-03-22
NavTabs overflow and z-index behaviour improved

### LINKED ISSUE #1923: Nav tabs responsiveness not working
**Describe the bug**
The NavTabs do not work correctly when the container is smaller than the width of the NavTabs.
Also, the NavTabs overlap elements that should be over the NavTabs.

**To Reproduce**
Right now the next version site does not work correctly on smaller devices.

**Screenshots**
image
image

Jialecl:
**Checklist**

**Purpose**
Overflow was added to the NavTabs and fixed an issue related to z-index.

**Description**
Overflow was added and the width of the NavTabs is now based on the available space including the scroll in case it is scrollable.
The z-index for the whole component is set to 0 and the underline is set to -1, instead of setting the component tabs to 1.

**Screenshots**
image
image

**Closes #1923**


Mil4n0r (commented):
There should be a cleaner way to perform this task without having to use `scrollWidth`, but I haven't been able to do it either.

Maybe there is a way to get rid of `Underline` entirely and just handle it in the `Tab` border-bottom. However, that would cause the problem stated here #1525 if no other modifications are done.

Mil4n0r (approved):
No other working solutions have been found and this pattern was already used in regular `Tabs` so we will stick to it.

PR #1933, opened 2024-04-02
New SideNav structure

### LINKED ISSUE #1642: Reimplement the Sidenav as a `<nav>` element
Refactor Sidenav's current implementation as a `<nav>` element with <ul> and <li> inside for Groups and Links. This will provide users with a more semantic HTML code.

Other design system examples:

- https://primer.style/css/components/side-nav
- https://spectrum.adobe.com/page/side-navigation/
- https://www.delldesignsystem.com/components/side-navigation/?tab=Code
- https://baseweb.design/components/side-nav/
- https://carbondesignsystem.com/components/UI-shell-header/usage

Mil4n0r:
**Checklist**

**Description**
The previous HTML structure of the SideNav was very poor in terms of semantics.

**Closes #1642**

### DIFF
--- a/lib/src/nav-tabs/NavTabs.tsx
+++ b/lib/src/nav-tabs/NavTabs.tsx
@@ -1,4 +1,4 @@
-import React, { useState, ReactElement, useMemo } from "react";
+import React, { useState, ReactElement, useMemo, useRef, useEffect } from "react";
 import styled, { ThemeProvider } from "styled-components";
 import useTheme from "../useTheme";
 import NavTabsPropsType from "./types";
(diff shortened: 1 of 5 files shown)


Mil4n0r:
There is still one problem to address. `Section` compound component is now treated as a `ul`, however, there are some scenarios where its child is a paragraph or any other text. In my opinion, we should create a specific compound component regarding that scenario. (For example `TextSection`) in case we don't want to lose that functionality, another more opinionated option would be to simply force the user to use list items as children. (Which is generally the correct case scenario for a SideNav)

Mil4n0r:
Closed for now until the topic is discussed.

PR #1935, opened 2024-04-02
Dialog popups interaction (Escape) fix

### LINKED ISSUE #1515: Pop up components keyboard interactions inside a dialog

Mil4n0r:
**Checklist**

**Description**
There was a PR for this (#1817) but the solution was found to not be appropiate. In the end it seems that the simplest and most appropiate solution for this case scenario is to handle the Escape button interaction for our current and future components in a special way, adding `event.stopPropagation()` to prevent them from closing the Dialog in case they are inside one.

**Closes #1515**


Jialecl (approved):

GomezIvann (changes requested):
Two things:

- The event propagation should only be stopped when the popovers are opened. If not, the propagation should work as usual (closing the dialog).
- We need to update the Dialog test to check this funcionality.

GomezIvann (approved):

PR #1938, opened 2024-04-03
Added new logic for resultsetTable to manage sorting and adding/removing rows properly

### LINKED ISSUE #1936: ResultsetTable returns to the first page when re-rendered.
**Describe the bug**
When the resultsetTable is re-rendered it always returns to the first page.
If we have inside the resultsetTable one component that changes its state, independently of the page in which it is when the state changes the resultsetTable returns to the first page.

**To Reproduce**
1. Create a resultsetTable with several pages and with radio buttons in one of its columns.
2. Move to the last page.
3. Click on one radio button.
4. ResultsetTable returns to the first page

**Expected behaviour**
ResultsetTable should stay on the page it was on.

Mil4n0r:
**Checklist**

**Description**
When interacting with `Checkbox` or any other changeable element, the page would be set to 1 no matter what the situation were. I have implemented a new logic in order to track changes when adding or removing rows.

I still haven't changed the tests because I want the new logic to be validated first, I also have doubts about how to handle both add/remove rows and cases where the data changes entirely (this last one not being taken into account right now).

**Closes #1936**


Mil4n0r:
Closed for now until the topic is discussed.

Mil4n0r:
Reopened, will remove it from Draft when I receive more feedback in the issue.

Mil4n0r:
The new problem shared by Alan is caused by the fact that we are currently using `index` from the map to set the keys. That is a problem that I detected while working in the code quality PR and is pending to be fixed, as it is considered a bad practice in React (Specially for sortable data).

This would lead to 2 possible solutions:
- Leaving the structure as it is but setting the keys with a randomly generated Id (such as useId or uuidv4), this is not the best solution because it is forced to generate new numbers each time that the data changes or is sorted.
- Changing the structure to ensure that each row has a unique id (either adding a field to the data structure and telling users to make sure that it is not repeated or using the current sortValue and force it to be unique).

lib/src/resultset-table/ResultsetTable.stories.tsx
```diff
@@ -409,5 +410,46 @@ export const DropdownAction = ResultsetActionsCellDropdown.bind({});
...
+    { displayValue: "003", sortValue: "003" },
+    {
+      displayValue: <DxcCheckbox size="fillParent" />,
+    },
+    { displayValue: "Lana" },
+    { displayValue: "Amsterdam" },
+  ],
+];
+
+export const ResultsetCheckbox = () => (
```
Jialecl, line 450: This test does not provide anything related to this fix unless we add interactions.
It is also covered in the new functional test.

Mil4n0r: I used it to test and forgot to remove it. Done!

Jialecl (approved):

Issue #1707, opened 2023-10-13
`disabled` in chip should only affect to the icon


raquelarrojo:
**Describe the bug**
At the moment, the chip can be `disabled` even it there is no action icon. It has no sense to be disabled since there is no interaction with it.

**Expected behaviour**
The `disabled` should only affect to the action icon.

**Screenshots**
![image](https://github.com/dxc-technology/halstack-react/assets/10975289/73757402-3db5-410e-8a63-08d94002ff66)

raquelarrojo:
Fix available here:
https://github.com/dxc-technology/halstack-react/tree/rarrojolopez/chip-fixes
https://github.com/dxc-technology/halstack-react/tree/rarrojolopez/chip-specs 

It has been reverted to avoi creating a breaking change.

PR #1943, opened 2024-04-10
Several fixes in the Chip

### LINKED ISSUE #1873: Chip must not have both prefix and suffix actions
The chip can only have one action.
It can have 2 icons if one of them has an action, if not, both icons are not allowed.

This decision has been taken after a design investigation.

raquelarrojo:
**Checklist**

**Description**
Several fixes in the chip component:

- Disabled state only affects to the icon with action, not to the entire component.
- Several types fixed:
       - `disabled` is only allowed when `onClickPrefix` or `onClickSuffix` exist.
       - It is not possible to add both `prefixIcon` and `suffixIcon` with action, only one of them with action can be added.
       - It is not possible to add both `prefixIcon` and `suffixIcon` (without action), only one of them can be added.
       - It is possible to add an icon with action and an icon without. For example: prefix with action but suffix without it.

**Snapshots**
Before the `disabled` affected the whole component:
image

Now `disabled` only affects the icon with action:

image

Documentation will be updated in a separate PR after this one is merged.

**Closes #1873 and #1707**


lib/src/chip/types.ts
```diff
@@ -8,31 +8,60 @@ type Margin = {
...
+       */
+      disabled?: boolean;
+    }
+  | {
+      prefixIcon?: Icon;
+      onClickPrefix?: never;
+      suffixIcon?: never;
+      onClickSuffix?: () => void;
+      disabled?: never;
+    };
```
Mil4n0r, line 42: I think that the typings are not right here. You are allowing the user to have undefined suffixIcon but an onClickSuffix function, which does not make sense.

lib/src/chip/types.ts
```diff
@@ -8,31 +8,60 @@ type Margin = {
...
+      onClickPrefix?: never;
+      disabled?: boolean;
+    }
+  | {
+      suffixIcon?: Icon;
+      onClickSuffix?: never;
+      prefixIcon?: never;
+      onClickPrefix?: () => void;
+      disabled?: never;
+    };
```
Mil4n0r, line 58: Same here for onClickPrefix and prefixIcon

lib/src/chip/types.ts
```diff
@@ -8,31 +8,60 @@ type Margin = {
...
-   * This function will be called when the prefix is clicked.
-   */
-  onClickPrefix?: () => void;
-  /**
-   * If true, the component will be disabled.
-   */
-  disabled?: boolean;
   /**
    * Size of the margin to be applied to the component ('xxsmall' | 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge').
    * You can pass an object with 'top', 'bottom', 'left' and 'right' properties in order to specify different margin sizes.
```
Mil4n0r, line 67: Maybe call this part CommonProps like it is done in some components such as `Badge` so that

```
type Props = (PrefixIconProps | SuffixIconProps) & CommonProps;
```

lib/src/chip/Chip.stories.tsx
```diff
@@ -50,33 +44,56 @@ const opinionatedTheme = {
...
     <ExampleContainer>
-      <Title title="Chip with prefix (SVG) and suffix (URL)" theme="light" level={4} />
-      <DxcChip label="Chip with prefix and suffix" prefixIcon={iconSVG} suffixIcon="filled_check_circle" />
+      <Title title="Action prefix and suffix" theme="light" level={4} />
+      <DxcChip
+        label="Action prefix and suffix"
+        prefixIcon="filled_check_circle"
+        onClickPrefix={() => {}}
+        suffixIcon={iconSVG}
+      />
```
Mil4n0r, line 73: Could you add an example with both prefix and suffix with the suffix being the one with the action?

lib/src/chip/Chip.stories.tsx
```diff
@@ -50,33 +44,56 @@ const opinionatedTheme = {
 export const Chromatic = () => (
   <>
     <ExampleContainer>
-      <Title title="Basic chip" theme="light" level={4} />
-      <DxcChip label="Default Chip" />
```
Mil4n0r, line 54: Maybe keep the default one with neither suffix or prefix?

Mil4n0r (changes requested):

lib/src/chip/types.ts
raquelarrojo: Done!

lib/src/chip/Chip.stories.tsx
raquelarrojo: Done!

lib/src/chip/Chip.stories.tsx
raquelarrojo: Done!

lib/src/chip/types.ts
raquelarrojo: Now, if `onClickPrefix` is put without `prefixIcon`, it throws an error.

lib/src/chip/types.ts
```diff
@@ -42,6 +71,8 @@ type Props = {
    * Value of the tabindex attribute.
    */
   tabIndex?: number;
-};
+} & (PrefixIconProps | SuffixIconProps);
```
Mil4n0r, line 74: I think this is not part of the `CommonProps` but part of the `Props` as you stated in the next line.

Mil4n0r (changes requested):

lib/src/chip/Chip.stories.tsx
```diff
@@ -50,33 +44,69 @@ const opinionatedTheme = {
...
+        suffixIcon={iconSVG}
+        onClickSuffix={() => {}}
+      />
+    </ExampleContainer>
+    <ExampleContainer>
+      <Title title="Disabled action suffix (Material icon)" theme="light" level={4} />
+      <DxcChip label="Disabled action suffix" suffixIcon="filled_check_circle" onClickSuffix={() => {}} disabled />
+    </ExampleContainer>
+    <ExampleContainer>
+      <Title title="Disabled action prefix and suffix" theme="light" level={4} />
```
Mil4n0r, line 93: I would separate this case in this one and another one where the (disabled) action is in the suffix. Something like this:
```
 <ExampleContainer>
      <Title title="TBD" theme="light" level={4} />
      <DxcChip
        label="Disabled"
        disabled
        prefixIcon={iconSVG}
        suffixIcon="filled_check_circle"
        onClickSuffix={() => {}}
      />
    </ExampleContainer>
```

Mil4n0r (changes requested):

lib/src/chip/Chip.stories.tsx
raquelarrojo: Done!

lib/src/chip/types.ts
raquelarrojo: Oh yes, it was a mistake to push this! Done!

Mil4n0r (approved):
Looks good to me.

Good job!

Mil4n0r (approved):

PR #1945, opened 2024-04-15
Code Quality Update

Mil4n0r:
**Checklist**

**Description**
We want to follow the best practices and achieve a higher code quality by using automatic tools (Husky, Prettier, ESLint and Staged-Lint). This PR contains both the setup of the tools and the fix of most existing code quality issues.

The goal to achieve is to remove all the custom rules existing in the `.eslintrc.json` file and pass with no errors and no warnings.


lib/src/BackgroundColorContext.tsx
Mil4n0r, line 1: Is this file going to be used with the new core tokens / themes? Otherwise it should be removed as it is not currently being used.

lib/.eslintrc.json
```diff
@@ -1,68 +1,120 @@
...
+    },
+    // Storybook files
+    {
+      "files": ["**/?(*.)+(stories).[jt]s?(x)"],
+      "rules": {
+        "jsx-a11y/no-noninteractive-tabindex": "off"
+      }
+    },
+    // FocusLock file
+    {
```
Mil4n0r, line 37: Maybe we should take a look at FocusLock implementation. Div (Containers in general) should never be focusable.

lib/src/text-input/TextInput.test.tsx
```diff
@@ -675,93 +853,178 @@ describe("TextInput component synchronous autosuggest tests", () => {
...
+      code: "Enter",
+      keyCode: 13,
+      charCode: 13,
+    });
     expect(input.value).toBe("Barbados");
     expect(queryByRole("listbox")).toBeFalsy();
-    fireEvent.keyDown(input, { key: "Esc", code: "Esp", keyCode: 27, charCode: 27 });
+    fireEvent.keyDown(input, {
+      key: "Esc",
+      code: "Esp",
```
Mil4n0r, line 1012: There was a typo here.

lib/src/text-input/Suggestion.tsx
```diff
@@ -3,13 +3,16 @@ import styled from "styled-components";
 import { SuggestionProps } from "./types";
 
 const transformSpecialChars = (str: string) => {
-  const specialCharsRegex = /[\\*()\[\]{}+?/]/;
+  const specialCharsRegex = /[\\*()[\]{}+?/]/;
```
Mil4n0r, line 6: Accor

According to ESLint, it seems that the backslash to escape the square bracket was unneeded. I have checked it and it seems to work equally this way but I would suggest you to test it just in case.

lib/src/accordion-group/AccordionGroup.tsx
Mil4n0r, line 1: Requires review.

Mil4n0r (commented):
Remember to add the correct configuration for Prettier before merging.

The old one is currently being used to reduce the amount of differences to check in the review.

Wanted configuration:
```
{
  "bracketSpacing": true,
  "semi": true,
  "singleQuote": false,
  "trailingComma": "es5",
  "tabWidth": 2,
  "printWidth": 80,
  "endOfLine": "lf",
  "arrowParens": "always"
}
```

Mil4n0r:
Will keep this PR as a Draft until Monorepo update is done.

Mil4n0r:
Moved to #2033

PR #1965, opened 2024-04-26
Use custom `DxcTooltip` instead of default HTML tooltip in Halstack Components

### LINKED ISSUE #1964: Add the `DxcTooltip` component to the library
Multiple Halstack components currently use the default tooltip. In order to achieve better consistency in terms of appearance and behavior, it has been decided to add this new component that will be used instead of the `title` prop when wanting to add a tooltip.

Mil4n0r:
**Checklist**

Related to PR #1965


lib/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@radix-ui/react-popover": "^1.0.7",
+    "@radix-ui/react-tooltip": "^1.0.7",
```
Mil4n0r, line 22: I tried implementing it with the same react-popover to prevent adding an extra library. However, the interaction with the current modals is buggy that way (dropdown, calendar...)

GomezIvann:
Hello Enrique!

But have you already implemented the component without making the API call after the corresponding research and drafting?

Mil4n0r:
Yes, I wanted to leave this as a draft until the discussion but it definitely made things easier for me to do the implementation first in order to make sure it covered all the possible scenarios.

After the logic is validated I will open it back.

lib/src/action-icon/ActionIcon.tsx
Mil4n0r, line 1: We should decide whether or not the `DxcIcon` should also have a `title` prop so it can have its own `DxcTooltip` as well.

lib/src/date-input/DateInput.tsx
```diff
@@ -209,7 +209,7 @@ const DxcDateInput = React.forwardRef<RefType, DateInputPropsType>(
 );
 
 const StyledPopoverContent = styled(Popover.Content)`
-  z-index: 2147483647;
+  z-index: 2147483646;
```
Mil4n0r, line 212: Reduced z-index by one to make sure `DxcTooltip` always on top.

lib/src/tooltip/Tooltip.stories.tsx
Mil4n0r, line 1: This is not the expected way to use `DxcTooltip` with Halstack components, as they should include a `title` prop that internally handles the tooltip (See `Button` for example). The only exception for that would be to add some customization (for example, an specific orientation for the tooltip).
These codes are equivalent:
```
<DxcTooltip label="Tooltip Test">
  <DxcButton label="Hoverable button" />
</DxcTooltip>
```
```
<DxcButton label="Hoverable button" title="Tooltip Test" />
```

lib/src/tooltip/Tooltip.accessibility.test.tsx
Mil4n0r, line 1: Maybe I should add examples with limited width/height to ensure that collision prevention is working as expected.

PR #1966, opened 2024-04-26
New attribute for Contextual Menu items

### LINKED ISSUE #1962: [DxcContextualMenu] Select an item by default
**Describe the bug**
We need to define a way of selecting one of the items of the DxcContextualMenu by default. For example, if the initial state of a screen is showing the content of one of the items of the contextual menu, we have no way to show the item as selected.

**Expected behaviour**
We should have a way to select by default one of the items of the DxcContextualMenu and show it in the same way as if it had been clicked.

**Screenshots**
image

GomezIvann:
**Checklist**

**Description**
We have decided to include a new attribute in the items of the Contextual Menu called `selectedByDefault`. This attribute will allow our users to mark an item from the menu as selected in the first render of the component.

We have also fixed two issues with the component:

- The gap between the section title and its items has been correctly set to `4px` / `0.25rem`.
- The icon of an item now is properly sized when the text is truncated.

**Closes #1962**


lib/src/contextual-menu/types.ts
Mil4n0r, line 1: For better consistency with the rest of our components, maybe consider wrapping `badge`, `icon` and `label` in a `CommonProps` variable (so we prevent code repetition).

lib/src/contextual-menu/ContextualMenu.stories.tsx
Mil4n0r, line 1: Unused import

`import { userEvent, within } from "@storybook/testing-library";`

lib/src/contextual-menu/GroupItem.tsx
```diff
@@ -1,32 +1,36 @@
...
 import ItemAction from "./ItemAction";
 import { ContextualMenuContext } from "./ContextualMenu";
 import DxcIcon from "../icon/Icon";
 
 const isGroupSelected = (items: GroupItemProps["items"], selectedItemId: number): boolean =>
-  items.some((item) => ("id" in item ? item.id === selectedItemId : isGroupSelected(item.items, selectedItemId)));
+  items.some((item) => {
+    if ("items" in item) return isGroupSelected(item.items, selectedItemId);
+    else if (selectedItemId !== -1) return item.id === selectedItemId;
+    else return (item as ItemWithId).selectedByDefault ? true : false;
```
Mil4n0r, line 14: This ternary operator is redundant (selectedByDefault is already a boolean).

GomezIvann: `selectedByDefault` can be `undefined` (optional prop) and our function always returns something. We need to ensure that even if it is undefined, it returns a boolean value (in this case, `false`).

lib/src/contextual-menu/ContextualMenu.stories.tsx
GomezIvann: Removed!

lib/src/contextual-menu/types.ts
GomezIvann: Done!

lib/src/contextual-menu/GroupItem.tsx
Mil4n0r: `undefined` will be treated as falsy inside `some` though.

Anyway, in these kind of scenarios where you want to cast something to boolean, what is considered the best practice is using double `!`.
```
else return !!(item as ItemWithId).selectedByDefault
```

GomezIvann: I didn't know that some treated undefined as false! I removed it.

Mil4n0r (approved):

PR #1968, opened 2024-04-30
Added `Tooltip` component

### LINKED ISSUE #1964: Add the `DxcTooltip` component to the library

Mil4n0r:
**Checklist**

**Description**
Adding a tooltip component is very important in order to achieve higher consistency, scalability, better interactivity and accessibility.

**Screenshots**
image

**Closes #1964**


lib/.storybook/test-runner.ts
Mil4n0r, line 1: https://github.com/storybookjs/test-runner/issues/395

The existing problem that occurs sometimes in the `preVisit` hook is being investigated.

lib/src/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,129 @@
+import React from "react";
+import styled from "styled-components";
+import TooltipPropsType from "./types";
+import * as Tooltip from "@radix-ui/react-tooltip";
+import CoreTokens from "../common/coreTokens";
+
+const DxcTooltip = ({ position = "bottom", label, children }: TooltipPropsType): JSX.Element => {
+  const SIZE = 8;
```
Jialecl, line 8: Rename it to `size` I would prefer not to use a const for the size of the arrow.

lib/src/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,129 @@
...
+            </TooltipArrow>
+          </StyledTooltipContent>
+        </Tooltip.Portal>
+      </Tooltip.Root>
+    </Tooltip.Provider>
+  ) : (
+    <>{children}</>
+  );
+};
+const Triangle = styled.span<{ size: number }>`
```
Jialecl, line 30: This is only a styled component because it needs the const `size`.

lib/src/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,129 @@
...
+      transform: translateX(-10px);
+    }
+    to {
+      opacity: 1;
+      transform: translateX(0);
+    }
+  }
+`;
+
+const TooltipArrow = styled(Tooltip.Arrow)`
```
Jialecl, line 113: Tooltip.Arrow does not need to be a styled component.

lib/src/tooltip/types.tsx
```diff
@@ -0,0 +1,18 @@
+type Position = "bottom" | "top" | "left" | "right";
+
+type Props = {
+  /**
+   * Preferred position for displaying the tooltip. It may adjust automatically based on available space.
+   */
+  position?: Position;
```
Jialecl, line 7: This is not a very complex type. I think it is better for the users to have the possible strings here instead ot a new type.

lib/src/tooltip/Tooltip.tsx
Jialecl: Instead of using a span have you considered a SVG?

lib/src/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+import DxcInset from "../inset/Inset";
+import Title from "../../.storybook/components/Title";
+import DxcFlex from "../flex/Flex";
+
+export default {
+  title: "Tooltip",
+  component: DxcTooltip,
+};
+
+const Tooltip = () => (
```
Jialecl, line 15: It is useless to have all tooltip visual tests in a single snapshot since it is not possible to show them all at the same time.

lib/src/tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+    );
+    const triggerElement = getByText("Hoverable button");
+    await userEvent.hover(triggerElement);
+    await waitFor(() => {
+      const tooltipContent = getByRole("tooltip", { name: "Tooltip Test" });
+      expect(tooltipContent).toBeTruthy();
+    });
+  });
+
+  test("Tooltip stops being rendered when hover is stopped", async () => {
```
Jialecl, line 44: the first 3 tests can be merged into a single test that mimics better the user behavior.

Jialecl: Should we also test the delay since it is a value we are assigning?

lib/src/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,129 @@
...
+const TooltipTriggerContainer = styled.div`
+  display: inline-flex;
+  position: relative;
+`;
+
+const StyledTooltipContent = styled(Tooltip.Content)`
+  z-index: 2147483647;
+
+  animation-duration: 0.2s;
+  animation-timing-function: ease-out;
```
Jialecl, line 53: Was the animation and delay discussed?

lib/src/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const SIZE = 8;
+
+  return label ? (
+    <Tooltip.Provider delayDuration={300}>
+      <Tooltip.Root>
+        <Tooltip.Trigger asChild>
+          <TooltipTriggerContainer>{children}</TooltipTriggerContainer>
+        </Tooltip.Trigger>
+        <Tooltip.Portal>
+          <StyledTooltipContent side={position} sideOffset={SIZE} data-testid={"popover-content"}>
```
Jialecl, line 17: I don't see the need to add a data-testid.

lib/src/tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const { getByText, getByRole } = render(
+      <DxcTooltip label="Tooltip Test">
+        <DxcButton label="Hoverable button" />
+      </DxcTooltip>
+    );
+    const triggerElement = getByText("Hoverable button");
+    await userEvent.hover(triggerElement);
+    await waitFor(() => {
+      const tooltipContent = getByRole("tooltip", { name: "Tooltip Test" });
+      const position = tooltipContent.closest("div").getAttribute("data-side");
```
Jialecl, line 69: I don't see the point of testing the position here.
We should cover this scenario in visual tests and `data-side` is a radix data attribute.

lib/src/tooltip/Tooltip.tsx
Mil4n0r: The animation was approved by Lara.

For the delay, I read some recommendations from multiple authors, the correct delay for a tooltip is 0.3s-0.5s (while many popular libraries set 0.5s by default) so I decided that a base delay of 0.3s and an animation of 0.2s would be appropiate. We can open a discussion regarding this topic if needed.

lib/src/tooltip/Tooltip.tsx
Mil4n0r: Totally! I forgot to remove it.

lib/src/tooltip/Tooltip.tsx
Mil4n0r: I just added the arrow SVG.

lib/src/tooltip/Tooltip.stories.tsx
Mil4n0r: The problem is that even though I trigger multiple hovers in the play function, only 1 tooltip shows at a time.

On the other hand, creating a story for each scenario seems overkill.

What is your suggestion?

lib/src/tooltip/Tooltip.test.tsx
Mil4n0r: I am currently having trouble setting it up because the code from the event `unhover` seems unreachable after a `waitFor`, will probably need some help to fix it.

lib/src/tooltip/Tooltip.stories.tsx
Mil4n0r: In the end it has been decided that, due to not being able to isolate the tooltip, it is needed to have a separate story for each position

lib/src/tooltip/Tooltip.test.tsx
Mil4n0r: The current code that I am testing is:

```
test("Tooltip displays when intended", async () => {
    const { getByText, queryByRole, getByRole } = render(
      <DxcTooltip label="Tooltip Test">
        <DxcButton label="Hoverable button" />
      </DxcTooltip>
    );
    const triggerElement = getByText("Hoverable button");
    await waitFor(() => {
      const tooltipElement = queryByRole("tooltip");
      expect(tooltipElement).toBeFalsy();
    });
    await userEvent.hover(triggerElement);
    await waitFor(() => {
      const tooltipContent = getByRole("tooltip", { name: "Tooltip Test" });
      expect(tooltipContent).toBeTruthy();
    });
    await userEvent.unhover(triggerElement);
    await waitFor(() => {
      const tooltipElement = queryByRole("tooltip");
      expect(tooltipElement).toBeFalsy();
    });
  });
```

image

lib/src/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+          <DxcButton label="Hoverable button" />
+        </DxcTooltip>
+      </DxcInset>
+    </ExampleContainer>
+  </>
+);
+
+const LargeTextWithinTooltip = () => (
+  <>
+    <Title title="Default tooltip" theme="light" level={4} />
```
Jialecl, line 30: Change the title and the snapshot gets cut.
image

Jialecl (approved):

PR #1990, opened 2024-05-10
Added custom tooltip to `Table` and `ResultsetTable`

Mil4n0r:
**Checklist**

**Description**
After adding the new `Tooltip` component, it is required to use it instead of title in the `Table` and `ResultsetTable` components.

**Related to #1968**


Mil4n0r:
Closed until `Tooltip` component lifecycle status is `New`.

Mil4n0r:
It may be inconvenient to add this feature, as it is overlapping the Dropdown content. I will investigate further

Mil4n0r:
Reopened with new solution to handle the overlapping of the tooltip with the Dropdown

packages/lib/src/dropdown/types.ts
```diff
@@ -76,6 +76,10 @@ type Props = {
    * Value of the tabindex attribute.
    */
   tabIndex?: number;
+  /**
+   * The title of the dropdown.
```
GomezIvann, line 80: I suggest to comment it as in other components. From the Button documentation:

image

GomezIvann (commented):
There's an example in the Usage page of the Dropdown with only icon as content. I recommend to add the title prop to this example and also a small explanation suggesting its use in this kind of scenarios.

Mil4n0r:
Hey @GomezIvann , I am facing a quite specific problem here.

As you already know, our `DxcTooltip` currently looks like this
image

You may be tempted to remove `<TooltipTriggerContainer>` and just leave it as `<>{children}</>` as that would make much more sense. 
image

However, this wouldn't work due to a limitation which is explained here: https://www.radix-ui.com/primitives/docs/guides/composition#your-component-must-forward-ref

The thing is that, in order to use a React Component as a direct child, it is needed that it provides a `forwardRef`. This is not a good approach as `forwardRef` will be removed from React soon, so our current workaround was to use this styled component as a parent, which was covering all the cases until now.

The problem comes in this scenario:
image

If we use `DxcTooltip` inside Dropdown like this, it won't work, because `Popover.Trigger` is expecting a `button` as a children (with the `aria-haspopup`, `disabled`, `onClick`, etc.
As `DxcTooltip` is just a div/span, it won't bind properly as a trigger and won't be working.

On the other hand, we could do this (moving down one level the tooltip):
image

In this case, the functionality itself works as expected, the problem is that it messes up our styles for the Dropdown, due to `DropdownTrigger` expecting 2 children to justify the content with space-between, adding the gaps, etc.

Do you have an idea on how to maintain the exact same layout as the current version but adding the tooltip?

Mil4n0r:
As the only approach that seems to be viable in this scenario is to go for another library to handle tooltips, I am keeping this as closed for now.

PR #1992, opened 2024-05-10
Added custom tooltip to `Select`

Mil4n0r:
**Checklist**

**Description**
After adding the new `Tooltip` component, it is required to use it instead of title in the `Select` component.

**Related to #1968, #2036**


Mil4n0r:
Closed until `Tooltip` component lifecycle status is `New`.

packages/lib/src/select/Select.tsx
```diff
@@ -624,6 +629,8 @@ const ClearSearchAction = styled.button`
   color: ${(props) => props.theme.actionIconColor};
   font-size: 1rem;
   cursor: pointer;
+  width: 100%;
```
GomezIvann, line 632: I think these extra values are not needed. I removed them and it looks correct.

packages/lib/src/select/Select.tsx
```diff
@@ -520,6 +523,8 @@ const ClearOptionsAction = styled.button`
   color: ${(props) =>
     props.disabled ? props.theme.disabledColor : props.theme.enabledSelectionIndicatorActionIconColor};
   font-size: 16px;
+  width: 100%;
+  height: 100%;
```
GomezIvann, line 527: `height: 100%;` is not needed, by default, a grid container always occupies the full height.

Mil4n0r: Correct.

packages/lib/src/select/Select.tsx
Mil4n0r: From my tests, this width is required, otherwise the icons won't be centered in the wrapper.
image

GomezIvann: The styles I'm suggesting removing are from the searchable input's clear action. What you show me is from the clear selected options!

Mil4n0r: You are absolutely right! Sorry for the confusion.

GomezIvann (approved):

PR #2005, opened 2024-05-23
New `DxcBarChart` component

GomezIvann:
**Checklist**

**Description**
New `DxcBarChart` component added to the Halstack Design System. Based on the Figma designs below, it uses Cloudscape's component under the hood.

image

**Props table**
https://nettle-opera-f35.notion.site/Bar-Chart-d8092551790b467f8e611f0f8033ee91?pvs=4

**Additional changes**
Some accessibility issues were reported at the Spinner component when the component was undetermined and the `mode` was small. Both `aria-label` and `aria-labelledby` attributes have been included to solve this issues.

image

**Additional context**
[Cloudscape Bar Chart](https://cloudscape.design/components/bar-chart)
[Halstack Bar Chart Designs](https://www.figma.com/design/ld2CyC6umeONXpRCK6LWm1/Halstack-Design-Library-(WIP%E2%9A%A0%EF%B8%8F)?node-id=4553-35653&t=tgmQwY5oHt3gEiRY-4)


lib/src/bar-chart/types.ts
```diff
@@ -0,0 +1,44 @@
+type DataTypes = string | number | Date;
```
Jialecl, line 1: There's no use for Date if the user is unable to format it.
image
I think we should remove Date from our valid types and force the user to format it into a string.

lib/src/bar-chart/BarChart.tsx
```diff
@@ -0,0 +1,179 @@
...
+    <ErrorStatus>
+      <Label>
+        <DxcIcon icon="cancel" />
+        {error}
+      </Label>
+      <DxcButton
+        label={translatedLabels.barChart.retryButtonLabel}
+        mode="secondary"
+        onClick={() => {
+          onRetry();
```
Jialecl, line 76: While most functional tests are unnecessary. I do think we should add some test for how the bar chart interacts with our components.

lib/src/bar-chart/types.ts
Jialecl: Another  issue with not being able to format the values is that you are not able to add a prefix or suffix to the Y domain, such as "100K" instead of "100000".

lib/src/bar-chart/BarChart.tsx
GomezIvann: Yes, I agree on this but there's a problem with Cloudscape's Bar Chart and our Jest configurations I couldn't solve...

lib/src/bar-chart/types.ts
GomezIvann: I know this, but providing our users with a formatting prop as Cloudscape does can be a bit overwhelming. I don't like either the idea of formatting us by default the Date values. Do you have any suggestions?

Jialecl: To format Date values I think the best approach is not to accept Date as a valid type. This will force the user to format it into a string, for example:

image

GomezIvann: I was considering the same idea! I'm going to restrict the prop type.

lib/src/bar-chart/BarChart.tsx
GomezIvann: I'm giving it a second try.

Jialecl: Something that I did not consider is that a condition can be added to check if onRetry is available before rendering the button since it is not a mandatory prop.

GomezIvann: Updated. I also think it should be optional.

lib/src/bar-chart/BarChart.test.tsx
```diff
@@ -0,0 +1,121 @@
+import React from "react";
+import { render, waitFor } from "@testing-library/react";
```
Jialecl, line 2: waitFor is not being used

lib/src/bar-chart/BarChart.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+          {
+            title: "Average",
+            type: "threshold",
+            y: 300000,
+          },
+        ]}
+      />
+    );
+    expect(getByText("Error")).toBeTruthy();
+    await userEvent.click(getByText("Retry"));
```
Jialecl, line 118: getByRole("button") is higher in priority than getByText

GomezIvann: done!

Jialecl (dismissed):

Jialecl (approved):

PR #2039, opened 2024-07-12
Fix `DxcLink` border radius

### LINKED ISSUE #1719: Update Link component `border-radius`
For consistency purposes, the  `border-radius` of the Link component is `2px` and we want to update it to `4px`. This is its value in most of our components. So, the tasks here are:

- [ ] Update the component implementation.
- [ ] Update the specifications, both images and possible token values.

Mil4n0r:
**Checklist**

**Description**
For consistency purposes, the border-radius of the Link component is 2px and we want to update it to 4px. This is its value in most of our components. 

**Additional context**
Even though it was listed as 2px, the images used were using `border-radius: 4px` internally. That is why it is only required to change the first image in the Specifications tab.

**Closes #1719**


packages/lib/src/link/Link.tsx
```diff
@@ -107,7 +107,7 @@ const StyledLink = styled.div<{
        cursor: pointer;`}
   }
   &:focus {
-    border-radius: 2px;
+    border-radius: 4px;
```
GomezIvann, line 110: If you hover the link while being focused, the `border-bottom` bends over the edges, which was noticeable until now, since the `border-radius` has been increased, we need to figure out another way to implement this.

GomezIvann (changes requested):

packages/lib/src/link/Link.tsx
Mil4n0r: I'll contact design to determine what the best solution for this is 
image

Mil4n0r: After talking to Irene, adding a 2px padding and slightly changing how the styles are applied seems to be appropiate.

Mil4n0r:
All images have been updated.

packages/lib/src/link/Link.tsx
```diff
@@ -107,8 +107,19 @@ const StyledLink = styled.div<{
...
+      margin: -2px;
+      position: absolute;
+      top: 0;
+      right: 0;
+      bottom: 0;
+      left: 0;
+      border-radius: 4px;
+      outline: 2px solid ${(props) => props.theme.focusColor};
+      pointer-events: none;
+    }
```
Mil4n0r, line 122: This is not the cleanest workaround but I haven't found anything else that worked as expected. Feel free to suggest a different solution.

packages/lib/src/link/Link.tsx
```diff
@@ -78,44 +78,37 @@ const StyledLink = styled.div<{
...
+      ${(props) => `color: ${props.theme.visitedFontColor};
+                    border-bottom-color: ${props.theme.visitedUnderlineColor};`}
     }
   }
-  &:hover {
-    ${(props) =>
-      `color: ${props.theme.hoverFontColor};
-       border-bottom-color: ${props.theme.hoverUnderlineColor};
-       cursor: pointer;`}
+  &:active {
```
GomezIvann, line 100: We can move the `:active` selector to the `LinkContainer` styled component and remove the `& > span` selector.

GomezIvann (commented):
One comment, comparing the component with the designs and the specs images I still think the component is too tall. Maybe we need to adjust this by updating the `line-height`, what do you think? Maybe we need Lara or Irene's help.

image

packages/lib/src/link/Link.tsx
Mil4n0r: AFAIK, :active selector only works for <a> or <button> (the same as :visited), so I don't think it is suitable for this case.

GomezIvann: Did you try? I tested it and it worked fine.

Mil4n0r:
> One comment, comparing the component with the designs and the specs images I still think the component is too tall. Maybe we need to adjust this by updating the `line-height`, what do you think? Maybe we need Lara or Irene's help.
> 
> image

I missed that, but I agree with you that it is too tall. Using 100% line-height for the anchor seems to look just like in the images.

apps/website/screens/components/link/specs/images/link_states.png
GomezIvann, line 1: The underline of the hover should fulfil the whole width of the component.

GomezIvann (changes requested):

GomezIvann (approved):

PR #2061, opened 2024-07-26
New `Tabs` component implementation

Mil4n0r:
**Checklist**

**Description**
Current `DxcTabs` API is confusing, as it is expected to work in a similar way as DxcNavTabs, but it does not provide Compound Components, so we have no access to onClick/onHover events separately for each Tab.

For consistency purposes and easier implementation of the wrappers required in Halstack Studio and UXPin, it is convenient to redo it.

For now, we are maintaining the old API available to prevent adding breaking changes, however, those props will be marked as LEGACY/DEPRECATED to not encourage using them in the future (and eventually, removing them).


Mil4n0r:
Missing changes in Doc.

Mil4n0r:
Good job @Jialecl. Changes look legit. The only thing to improve I can think of would be to improve the typings so that only legacy behavior or new behavior are allowed, not both at once. However, I am not sure if that is worth it taking into account that the legacy behavior will eventually be removed.

Let me know what you guys think @Jialecl @GomezIvann.

packages/lib/src/tabs/Tab.tsx
```diff
@@ -1,85 +1,139 @@
...
+      activeLabel,
+      hasLabelAndIcon,
+      setActiveLabel,
+      setActiveIndicatorWidth,
+      setActiveIndicatorLeft,
+    } = useContext(TabsContext);
+
+    useEffect(() => {
+      focusedLabel === label && tabRef?.current?.focus();
+    }, [focusedLabel]);
```
GomezIvann, line 41: The `label` prop is missing in the three hook's dependency arrays.

packages/lib/src/tabs/Tabs.tsx
```diff
@@ -34,57 +52,70 @@ const DxcTabs = ({
...
   useEffect(() => {
-    let sumWidth = refTabs?.current?.reduce(function (count, obj) {
-      return count + obj.offsetWidth;
-    }, 0);
-    setTotalTabsWidth(sumWidth);
-    setActiveIndicatorWidth(refTabs?.current[activeTabIndex ?? innerActiveTabIndex]?.offsetWidth);
-    setActiveIndicatorLeft(refTabs?.current[activeTabIndex ?? innerActiveTabIndex]?.offsetLeft);
-  }, [refTabs]);
+    setTotalTabsWidth(refTabList?.current?.firstElementChild?.offsetWidth);
+  }, [refTabList]);
```
GomezIvann, line 99: refs should not go inside the dependency arrays.

packages/lib/src/tabs/Tabs.tsx
```diff
@@ -34,57 +52,70 @@ const DxcTabs = ({
...
-  };
+  const contextValue = useMemo(() => {
+    const focusedChild = Children.toArray(children)[innerFocusIndex];
+    return {
+      iconPosition,
+      tabIndex,
+      focusedLabel: isValidElement(focusedChild) && focusedChild.props.label,
+      isControlled: childrenArray.some((child) => isValidElement(child) && typeof child.props.active !== "undefined"),
+      activeLabel: activeTabLabel,
+      hasLabelAndIcon: hasLabelAndIcon,
```
GomezIvann, line 113: When the value is the same as the property name you can remove it:

```
{
   {...}
   hasLabelAndIcon,
   {...}
}
```

packages/lib/src/tabs/Tabs.tsx
```diff
@@ -34,57 +52,70 @@ const DxcTabs = ({
...
+      setActiveLabel: setActiveTabLabel,
+      setActiveIndicatorWidth,
+      setActiveIndicatorLeft,
+    };
+  }, [iconPosition, tabIndex, innerFocusIndex, activeTabLabel]);
```
GomezIvann, line 118: Theoretically, `hasLabelAndIcon` and `childrenArray` should go inside the dependency array, but, I think that may trigger an endless loop because their value comes from `Children`... I'm not sure about this implementation 🤔.

GomezIvann: I would encourage you to look over the usage of Children and try to avoid it as much as possible. I link you some alternatives:

https://react.dev/reference/react/Children#alternatives

packages/lib/src/tabs/types.ts
```diff
@@ -22,7 +22,20 @@ type TabCommonProps = {
   notificationNumber?: boolean | number;
 };
 
-export type TabLabelProps = TabCommonProps & {
+export type TabsContextProps = {
+  iconPosition: "top" | "left";
+  tabIndex: number;
+  focusedLabel: string;
+  isControlled: boolean;
+  // setFocusedLabel: (_tab: string) => void;
```
GomezIvann, line 30: Remove code comments.

GomezIvann (changes requested):

packages/lib/src/tabs/Tabs.tsx
Mil4n0r: I think I found a workaround for the endless loop using memoizing. I have just added it, can you check it?

GomezIvann (commented):
Just a general thought. There are too many variables depending on each other (because of `useMemo`) inside `Tabs.tsx` which are only use once or twice. For example:

1. `hasActiveChild`, `initialActiveTabLabel` can be transformed into an initialization function inside the `useState` of `activeTabLabel`:

```
  const childrenArray: ReactElement<TabProps>[] = useMemo(
    () => Children.toArray(children) as ReactElement<TabProps>[],
    [children]
  );
  const hasLabelAndIcon = useMemo(
    () => childrenArray.some((child) => isValidElement(child) && child.props.icon && child.props.label),
    [childrenArray]
  );

  const [activeTabLabel, setActiveTabLabel] = useState(() => {
    const hasActiveChild = childrenArray.some(
      (child) => isValidElement(child) && (child.props.active || child.props.defaultActive) && !child.props.disabled
    );
    const initialActiveTab = hasActiveChild
      ? childrenArray.find(
          (child) => isValidElement(child) && (child.props.active || child.props.defaultActive) && !child.props.disabled
        )
      : childrenArray.find((child) => isValidElement(child) && !child.props.disabled);

    return isValidElement(initialActiveTab) ? initialActiveTab.props.label : "";
  });
```
2. You can directly pass a function to `aria-disabled` from `ActiveIndicator` in order to remove another const (`isActiveIndicatorDisabled`).
```
                  aria-disabled={childrenArray.some((child) =>
                    isValidElement(child) ? activeTabLabel === child.props.label && child.props.disabled : false
                  )}
```

Since you are already memorizing the value of `childrenArray`, all this changes make sense and don't trigger rerenders.

GomezIvann (approved):

PR #2082, opened 2024-08-22
Fixed problem in ResultsetTable when sortValue is missing

Mil4n0r:
**Checklist**

**Description**
SortValue must take displayValue instead if it is undefined. After changing the  structure adding the "cells" item, this is not working.


Mil4n0r:
I will be adding interaction tests for this scenario to prevent it from happening again.

jsuarezgonz (changes requested):
Please add a functional test to check that the columns are correctly sorted according to 'displayValue' when 'sortValue' is undefined

jsuarezgonz (approved):

PR #2099, opened 2024-10-04
DataGrid improvements

Mil4n0r:
**Checklist**

**Purpose**
A clear description of the purpose of the change. Applicable in case this change is not linked to an issue.

**Description**
The following problems have been addressed:
- Adding the possibility to sort any kind of component based on any criteria.
- Add the possibility to paginate the DataGrid component.
- Fixed the unwanted behavior for draggable columns along with expandable content (When expanding or collapsing the content, the column order would roll back to the initial status)
- Fixed unwanted behavior when toggling a grandchild shown in this quick video:
[grandchild-toggle-demo.webm](https://github.com/user-attachments/assets/6934c93d-9130-440e-af88-c987331e0158)


Mil4n0r:
I would like to add many edge cases to both functional and visual tests, I also have to tidy up the current tests that I added.

I also have to update the API with all the documentation required for pagination and the sorting strategy.

apps/website/screens/components/data-grid/code/DataGridCodePage.tsx
```diff
@@ -167,7 +171,7 @@ const sections = [
           <tr>
             <td>uniqueRowId</td>
             <td>
-              <TableCode>string</TableCode>
+              <TableCode>string | number</TableCode>
```
Jialecl, line 174: The type should be string only, but the type which the key has can be string or number.

Jialecl (approved):

PR #2108, opened 2024-10-25
Actions cell added to the dataGrid

Jialecl:
**Checklist**

**Purpose**
Actions cell compound component added to the data grid.

### DIFF
--- a/apps/website/screens/components/data-grid/code/DataGridCodePage.tsx
+++ b/apps/website/screens/components/data-grid/code/DataGridCodePage.tsx
@@ -1,4 +1,4 @@
-import { DxcFlex, DxcTable } from "@dxc-technology/halstack-react";
+import { DxcFlex, DxcLink, DxcParagraph, DxcTable } from "@dxc-technology/halstack-react";
 import QuickNavContainer from "@/common/QuickNavContainer";
 import QuickNavContainerLayout from "@/common/QuickNavContainerLayout";
 import DocFooter from "@/common/DocFooter";
@@ -13,6 +13,8 @@ import hierarchical from "./examples/hierarchical";
 import hierarchicalSelectable from "./examples/hierarchicalSelectable";
 import paginator from "./examples/paginator";
 import customSorting from "./examples/customSort";
+import Link from "next/link";
+import actionsCell from "./examples/actionsCell";
 
 const columnsTypeString = `{
   key: string;
@@ -38,6 +40,20 @@ const ExpandableGridRowTypeString = `GridRow & {
   expandedContent?: React.ReactNode;
   expandedContentHeight?: number;
 }`;
+
+const actionsType = `{
+  icon: string | SVG;
+  title: string;
+  onClick: () => void;
+  disabled?: boolean;
+  tabIndex?: number;
+}[] | {
+  title: string;
+  onClick: (value?: string) => void;
+  disabled?: boolean;
+  tabIndex?: number;
+  options: Option[];
+}[]`;
 
 const sections = [
   {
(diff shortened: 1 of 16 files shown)


apps/website/screens/theme-generator/components/previews/DataGrid.tsx
```diff
@@ -60,6 +80,7 @@ const expandableRows = [
     priority: "High",
     issueType: "Epic",
     expandedContent: <DxcContainer>Custom content 2</DxcContainer>,
+    actions: <DxcDataGrid.ActionsCell actions={actions as any} />,
```
Mil4n0r, line 83: What do you think about exporting the type `Action` in `packages/lib/src/table/types.ts` 

```
export type ActionCellsPropsType = {
  actions: Actions;
};

export type Actions = Array<
  | {
      icon: string | SVG;
      title: string;
      onClick: () => void;
      disabled?: boolean;
      tabIndex?: number;
      options?: never;
    }
  | {
      icon?: never;
      title: string;
      onClick: (value?: string) => void;
      disabled?: boolean;
      tabIndex?: number;
      options: Option[];
    }
>;
```

This way, we could prevent using `actions as any` everywhere, including `Table.stories` and `ResultsetTable.stories`

packages/lib/src/common/variables.ts
```diff
@@ -366,6 +366,17 @@ export const componentTokens = {
     focusColor: CoreTokens.color_blue_600,
     scrollBarThumbColor: CoreTokens.color_grey_700,
     scrollBarTrackColor: CoreTokens.color_grey_300,
+    // action icon tokens
```
Mil4n0r, line 369: Shouldn't all the tokens contain the `actionIcon` keyword? (Instead of just action for the last 5)

packages/lib/src/data-grid/DataGrid.stories.tsx
```diff
@@ -8,6 +8,7 @@ import { disabledRules } from "../../test/accessibility/rules/specific/data-grid
 import preview from "../../.storybook/preview";
 import { userEvent, within } from "@storybook/test";
 import DxcBadge from "../badge/Badge";
+import DxcTable from "../table/Table";
```
Mil4n0r, line 11: Unused import.

apps/website/screens/components/data-grid/code/DataGridCodePage.tsx
```diff
@@ -262,14 +278,74 @@ const sections = [
...
+                    <StatusBadge status="required" />
+                    actions
+                  </DxcFlex>
+                </td>
+                <td>
+                  <ExtendedTableCode>{actionsType}</ExtendedTableCode>
+                </td>
+                <td>
+                  <p>
+                    It represents a list of interactuable elements that will work as buttons or as a dropdown. Those
```
Mil4n0r, line 315: `interactuable` is not correct, please place `interactable` or `interactive` instead (I personally prefer the latter).

I've realised that this typo exists in other doc and also in the `Header` and `Chip` implementation. We should fix that.

Mil4n0r (changes requested):

Mil4n0r (changes requested):
Also check the changes to the alignment having affected chevrons in the headers

apps/website/screens/theme-generator/components/previews/DataGrid.tsx
Jialecl: I added the export, but I don't think there will be any change in this file.

Mil4n0r (dismissed):
Good job, as always 😺

Mil4n0r (approved):

PR #2109, opened 2024-10-25
New Installation page

GomezIvann:
**Checklist**

**Description**
We wanted to improve our documentation site as part of this release's Halstack enhancements. The first idea is to add Installation guidelines to help users install the `halstack-react` npm package.

**Extra changes:**
- Link padding removed to avoid extra spacing between words when there's a link in between.
- Adjusted responsive breakpoints in the website. Now the side nav disappears at a small laptop size and the quick nav at a table size.
- New CodeBlock component in the site.
- Example site component has been updated using Halstack color palette (it used to have its own colors).

**Screenshots**
image

### DIFF
--- a/apps/website/screens/common/CodeBlock.tsx
+++ b/apps/website/screens/common/CodeBlock.tsx
@@ -1,0 +1,69 @@
+import { ReactNode, useState } from "react";
+import { DxcButton, DxcContainer, DxcFlex, HalstackProvider } from "@dxc-technology/halstack-react";
+import styled from "styled-components";
+
+const PreformattedText = styled.pre`
+  position: relative;
+  box-sizing: border-box;
+  margin: 0;
+  border-radius: 0.25rem;
+  padding: 1rem 1rem 1rem 1.5rem;
+  background-color: #333;
+`;
+
+const Code = styled.code`
+  font-size: 0.875rem;
+  white-space: pre-wrap;
+  word-wrap: break-word;
+  color: #fff;
+`;
+
+const buttonTokens = {
+  button: {
+    tertiaryDefaultFontColor: "#fff",
+    tertiaryHoverDefaultBackgroundColor: "#4d4d4d",
+    tertiaryActiveDefaultBackgroundColor: "#666666",
+  },
+};
+
+export default function CodeBlock({ children }: { children: ReactNode; language?: string }) {
+  const [copyActionIsVisible, setCopyActionIsVisible] = useState(false);
+  const [copied, setCopied] = useState(false);
+
+  const onMouseEnter = () => {
+    setCopyActionIsVisible(true);
+  };
+  const onMouseLeave = () => {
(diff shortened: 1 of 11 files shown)


apps/website/screens/overview/introduction/IntroductionPage.tsx
```diff
@@ -109,30 +109,26 @@ const sections = [
           </DxcParagraph>
         ),
       },
+      {
+        title: "Figma",
+        content: (
+          <DxcParagraph>
+            Halstack's Design System is built on top of Figma, a web-based design tool that allows designers to create
```
Mil4n0r, line 116: Halstack's Design System -> Halstack Design System

apps/website/screens/overview/installation/InstallationPage.tsx
```diff
@@ -0,0 +1,114 @@
...
+          </>
+        ),
+      },
+    ],
+  },
+  {
+    title: "What's next?",
+    content: (
+      <DxcParagraph>
+        You can now start building your project using the Halstack Design System. But, before you do so, we recommend
```
Mil4n0r, line 92: First sentence is a bit redundant with the 3rd and last step from the subsections.

packages/lib/src/link/Link.tsx
```diff
@@ -79,7 +79,6 @@ const StyledLink = styled.div<{
   border: none;
   border-radius: 4px;
   width: fit-content;
-  padding: 0 2px;
```
Mil4n0r, line 82: Even though padding is not adequate because of the extra spacing between words when using links, I think `line-height` and `outline-offset` should be used to maintain the margin on focus as it is shown in the docs.

apps/website/screens/common/CodeBlock.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const [copied, setCopied] = useState(false);
+
+  const onMouseEnter = () => {
+    setCopyActionIsVisible(true);
+  };
+  const onMouseLeave = () => {
+    setCopyActionIsVisible(false);
+    setCopied(false);
+  };
+  const copyCode = () => {
```
Mil4n0r, line 40: This is cool. However I think we could add a few minor features:
- When hovering and pressing Ctrl + C, it would be interesting to trigger the `copyCode`.
- It would be much more visually appealing to have it display the tooltip while the "Copied" message is active.

GomezIvann: - About the first point, I don't think It makes much sense. I find it very unintuitive since you do not have any visual clue of what is currently selected.
- How can you do that?

packages/lib/src/link/Link.tsx
GomezIvann: I'm gonna review this with the designers!

GomezIvann: Updated!

PR #2111, opened 2024-10-31
Added controlled behavior to `Tooltip`

Mil4n0r:
**Checklist**

**Purpose**
We didn't have a way to have control over the tooltip display, We don't have to reduce it to be opened on hover and this will help us cover a lot of possible use cases in the future.

### DIFF
--- a/apps/website/screens/components/tooltip/code/TooltipCodePage.tsx
+++ b/apps/website/screens/components/tooltip/code/TooltipCodePage.tsx
@@ -41,6 +41,38 @@ const sections = [
             <td>-</td>
           </tr>
           <tr>
+            <td>open</td>
+            <td>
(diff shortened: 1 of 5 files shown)


Mil4n0r:
For some reason, radix's `onOpenChange` callback is not functioning as expected when using controlled `open`. So I had to go for a different approach in order to implement it.

I am going to create an issue in their repo showing the problem, as the function only triggers when hovering the element.

Mil4n0r:
This PR is being closed as it has been decided that this use case does not really fit the expected behavior of a tooltip and it could lead to overloading it. 

For our internal cases such as `Slider`, we are going to use an internal `Bubble` component (name has to be confirmed) which covers this case. In the future we could consider exporting it.

PR #2112, opened 2024-10-31
Fixed underline issue within `Tabs`

### LINKED ISSUE #2110: The bottom border of the active tab disappears under certain conditions
**Describe the bug**
The bottom border of the active tab disappears under certain conditions while using the new "Controlled" Tabs.

**To reproduce**
Follow these steps:

1. Open the following link: https://developer.dxc.com/halstack/14/components/tabs/.
2. Scroll down to the "Controlled" Tabs example.
3. Select different tabs and then click outside the tabs component.
4. Notice that the bottom border of the active tab disappears.

**Expected behaviour**
The bottom border should remain visible on the active tab while using the "Controlled" Tabs.

Mil4n0r:
**Checklist**

**Description**
Under certain interactions, the underline for the active item would move down until not being shown at all.

**Additional context**
It has been happening since `DxcTabs` API was refactorized to use compound component. A change in the dependency array to calculate the width and height was causing this problem.

**Closes #2110**

### DIFF
--- a/packages/lib/src/tabs/Tabs.tsx
+++ b/packages/lib/src/tabs/Tabs.tsx
@@ -93,7 +93,7 @@ const DxcTabs = ({
       setTotalTabsWidth(refTabList.current.firstElementChild?.offsetWidth);
       setMinHeightTabs(refTabList?.current?.offsetHeight + 1);
     }
-  }, [refTabList]);
+  }, []);
 
   const contextValue = useMemo(() => {
     const focusedChild = childrenArray[innerFocusIndex];


packages/lib/src/tabs/Tabs.tsx
```diff
@@ -93,7 +93,7 @@ const DxcTabs = ({
       setTotalTabsWidth(refTabList.current.firstElementChild?.offsetWidth);
       setMinHeightTabs(refTabList?.current?.offsetHeight + 1);
     }
-  }, [childrenArray]);
+  }, [refTabList]);
```
GomezIvann, line 96: Refs do not go inside the dependency array.

GomezIvann (changes requested):

packages/lib/src/tabs/Tabs.tsx
Mil4n0r: That's why it was initially changed, the problem here is that in case the tab list is updated, the total width needs to be updated.

Mil4n0r: Totally right. I'm leaving it as a Draft until I fix this.

GomezIvann (approved):

PR #2124, opened 2024-12-03
Fix: Internal `BaseTypography` removed

Jialecl:
**Checklist**

- [x] The Build process is done without errors. All tests pass in the `/lib` directory.

**Description**

Removing the internal `BaseTypography` component to prevent build issues with some apps using a different bundler. The properties set using the Typography context where no applied correctly.

We can revisit this in the future since it is not our fault or an implementation issue but, for now, we will solve it like this.


packages/lib/src/nav-tabs/Tab.tsx
```diff
@@ -60,7 +59,7 @@ const DxcTab = forwardRef(
             </TabIconContainer>
           )}
           <DxcFlex alignItems="center" gap="0.5rem">
-            <BaseTypography
+            <LabelContainer
```
GomezIvann, line 62: I don't understand this implementation. Why aren't you consuming the theme from the provider as in the other styled-components?

I know it is an internal component without a `ThemeProvider` but it should only be used outside the NavTabs, which do have such a provider, so this component would have access to it.

packages/lib/src/nav-tabs/Tab.tsx
```diff
@@ -137,6 +133,28 @@ const Tab = styled.a<{
...
 
+const LabelContainer = styled.span<{
+  color: string;
+  fontFamily: string;
+  fontSize: string;
+  fontStyle: string;
+  fontWeight: string;
+}>`
+  display: inline;
+  color: ${(props) => props.color};
```
GomezIvann, line 144: Use `props.theme.xxx` instead and remove the props.

packages/lib/src/tabs/Tab.tsx
```diff
@@ -111,12 +110,9 @@ const DxcTab = forwardRef(
               fontSize={colorsTheme.tabs.fontSize}
```
GomezIvann, line 110: Same as in the Nav Tabs, don't use props and instead, consume the theme.

packages/lib/src/tabs/TabLegacy.tsx
```diff
@@ -170,6 +166,28 @@ const MainLabelContainer = styled.div<{
       : "unset"};
 `;
 
+const LabelContainer = styled.span<{
+  color: string;
```
GomezIvann, line 170: Same as in the Nav Tabs, don't use props and instead, consume the theme.

packages/lib/src/paragraph/Paragraph.tsx
```diff
@@ -1,20 +1,35 @@
...
 import useTheme from "../useTheme";
-import BaseTypography from "../utils/BaseTypography";
 
 const DxcParagraph = ({ children }: { children: React.ReactNode }): JSX.Element => {
   const colorsTheme = useTheme();
 
   return (
-    <BaseTypography
-      as="p"
+    <ParagraphContainer
```
GomezIvann, line 8: Implement the component as we usually do Jiale, use the ThemeProvider!

`<ThemeProvider theme={colorsTheme.paragraph}>`

GomezIvann (changes requested):

packages/lib/src/typography/Typography.tsx
```diff
@@ -1,14 +1,120 @@
...
-    textOverflow={textOverflow}
-    whiteSpace={whiteSpace == null && textOverflow != null && textOverflow !== "unset" ? "nowrap" : whiteSpace}
-    {...props}
-  >
-    {children}
-  </BaseTypography>
-);
+const TypographyContext = React.createContext<TypographyContextPropTypes | null>(null);
 
+const ValidTypographyTags = [
```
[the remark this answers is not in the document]
GomezIvann: I did the change, we need this fix. We can discuss it later.

packages/lib/src/typography/Typography.tsx
```diff
@@ -1,14 +1,120 @@
...
+  "h5",
+  "h6",
+  "p",
+  "pre",
+  "small",
+  "span",
+  "strong",
+];
+
+const isValidTypography = (tag: keyof HTMLElementTagNameMap) => {
```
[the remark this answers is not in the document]
GomezIvann: I did the change, we need this fix. We can discuss it later.

Mil4n0r (approved):

GomezIvann (approved):

PR #2125, opened 2024-12-03
Updates on the FocusLock and Dialog

GomezIvann:
**Checklist**

**Purpose**
The focus was not following the tabbing sequence in Dialog and Alert components when there was a component with the new Tooltip. It was an issue with the Focus Lock internal component.

**Description**
Summary of changes:
- Fixed the issue. The `MutationObserver` from the `FocusLock` component was executing itself each time the focus changed, due to a flag of the observer (`{ attributes: true }`). Now it is only executing when the tree inside the component changes, which is the expected behavior.
  - Fixed some usual issues with the Focus Lock like the Date Input calendar focus return on close, added focus guards etc.
  - New tests.
- Change Dialog's prop `isCloseVisible` to `closable`, to match the new Alert approach for this same property.

### DIFF
--- a/packages/lib/src/utils/FocusLock.tsx
+++ b/packages/lib/src/utils/FocusLock.tsx
@@ -69,7 +69,7 @@ const useFocusableElements = (ref: React.MutableRefObject<HTMLDivElement>): HTML
       const observer = new MutationObserver(() => {
         setFocusableElements(getFocusableElements(ref.current));
       });
-      observer.observe(ref.current, { childList: true, subtree: true, attributes: true });
+      observer.observe(ref.current, { childList: true, subtree: true });
       return () => {
         observer.disconnect();
       };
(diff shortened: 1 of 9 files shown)


packages/lib/src/utils/FocusLock.tsx
```diff
@@ -90,23 +90,43 @@ const useFocusableElements = (ref: React.MutableRefObject<HTMLDivElement>): HTML
...
+        !radixPortalContains(event.relatedTarget as Node)
+      )
+        focusFirst();
+    };
+
+    document.addEventListener("focusout", focusGuardHandler);
+    return () => {
+      document.removeEventListener("focusout", focusGuardHandler);
+    };
   }, [focusFirst]);
```
Mil4n0r, line 130: I think this should be refactored for easier readability and reduced code repetition:

```
useEffect(() => {
  const focusGuardHandler = (event: FocusEvent) => {
    const target = event.relatedTarget as Node | null;
    const container = childrenContainerRef.current;

    if (
      target &&
      !(
        container?.contains(target) ||
        container?.nextElementSibling?.contains(target) ||
        container?.previousElementSibling?.contains(target) ||
        radixPortalContains(target)
        // This can of course also be expressed as a conjunction 
        // of negations instead of negation of a disjunction, 
        // I find the latter a bit easier to comprehend.
      )
    ) {
      focusFirst();
    }
  };
  ```

GomezIvann: It's great Enrique, I will apply those changes.

Mil4n0r (approved):

PR #2142, opened 2025-01-14
ariaLabel prop added to select component

Jialecl:
**Checklist**

**Purpose**
`ariaLabel` prop added to select component.


packages/lib/src/select/Select.tsx
```diff
@@ -52,6 +52,7 @@ const DxcSelect = forwardRef<RefType, SelectPropsType>(
       margin,
       size = "medium",
       tabIndex = 0,
+      ariaLabel = "Select input",
```
GomezIvann, line 55: As with the slider, I suggest removing the input word and having the same component name.

GomezIvann (changes requested):

PR #2145, opened 2025-01-15
ariaLabel prop added to spinner

Jialecl:
**Checklist**

`ariaLabel` prop added to the spinner component to improve accessibility when label is not available.

### DIFF
--- a/packages/lib/src/spinner/Spinner.test.tsx
+++ b/packages/lib/src/spinner/Spinner.test.tsx
@@ -13,9 +13,12 @@ describe("Spinner component tests", () => {
   });
 
   test("Small spinner hides value and label correctly", () => {
-    const { queryByText } = render(<DxcSpinner mode="small" label="test-loading" value={75} showValue></DxcSpinner>);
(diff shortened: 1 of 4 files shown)


packages/lib/src/spinner/Spinner.test.tsx
Mil4n0r, line 1: On the 3rd test you could put the following:
```diff
test("Small spinner hides value and label correctly", () => {
    const { queryByText } = render(<DxcSpinner mode="small" label="test-loading" value={75} showValue></DxcSpinner>);
    expect(queryByText("test-loading")).toBeFalsy();
    expect(queryByText("75%")).toBeFalsy();-
+  const spinner = getByRole("progressbar");
+    expect(spinner.getAttribute("aria-label")).toBe("Spinner");
  });
```

So that we test that the default value is being used instead of the label when no ariaLabel is provided. Feel free to put it in a separate test if you prefer so.

Mil4n0r (changes requested):

PR #2148, opened 2025-01-27
Button redesign

GomezIvann:
**Checklist**

**Description**
Redesign of the button component following the new design kit specifications, already using the new token architecture (CSS variables).

**Additional context**


Mil4n0r (changes requested):
A few minor changes.

Won't accept the PR until we get to update the docs though.

packages/lib/src/button/Button.tsx
```diff
@@ -1,502 +1,98 @@
...
   display: flex;
-  font-size: ${(props) => (props.size?.height === "small" ? "16" : props.size?.height === "medium" ? "16" : "24")}px;
+  font-size: var(${({ size }) => (size?.height === "large" ? "--height-s" : "--height-xxs")});
   svg {
-    height: ${(props) => (props.size?.height === "small" ? "16" : props.size?.height === "medium" ? "16" : "24")}px;
-    width: ${(props) => (props.size?.height === "small" ? "16" : props.size?.height === "medium" ? "16" : "24")}px;
+    height: var(${({ size }) => (size?.height === "large" ? "--height-s" : "--height-xxs")});
+    width: ${({ size }) => (size?.height === "large" ? "24" : "16")}px;
   }
 `;
```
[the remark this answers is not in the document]
GomezIvann: Updated!

packages/lib/src/button/types.ts
```diff
@@ -19,11 +21,11 @@ type Props = {
...
    */
-  mode?: "primary" | "secondary" | "tertiary";
+  mode?: Mode;
   /**
    * Specifies the semantic meaning of the buttons, which determines its color.
    */
-  semantic?: "default" | "error" | "warning" | "success" | "info";
+  semantic?: Semantic;
   /**
    * If true, the component will be disabled.
```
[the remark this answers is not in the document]
GomezIvann: The disabled prop comment is the same in every component (docs + types). For consistency purposes I suggest to keep it like that.

packages/lib/src/button/utils.ts
```diff
@@ -0,0 +1,220 @@
...
+  const commonStyles = `
+    font-family: var(--typography-font-family);
+    font-size: var(${size?.height === "large" ? "--typography-label-l" : "--typography-label-m"});
+    font-weight: var(--typography-label-semibold);
+    border: var(--border-width-none) var(--border-style-default) transparent;
+    border-radius: var(--border-radius-s);
+
+    &:focus:enabled {
+      outline: var(--border-width-m) var(--border-style-default) var(--border-color-secondary-medium);
+      outline-offset: -2px;
```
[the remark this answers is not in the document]
GomezIvann: It doesn't make sense to set the `outline` in an element's active state since an active button cannot be present without gaining the focus; it would be redundant.

It is like that on Figma because the code is autogenerated and based on the design. You cannot make a focused active component in Figma without setting the focus for both states (an active state in Figma doesn't apply the designs of the focus state, that's a functional implication.

packages/lib/src/button/Button.tsx
```diff
@@ -17,7 +17,8 @@ const Button = styled.button<{
   align-items: center;
   justify-content: center;
   flex-direction: ${({ iconPosition }) => (iconPosition === "after" ? "row" : "row-reverse")};
-  gap: ${({ size }) => (size.height === "large" ? "var(--spacing-gap-s)" : "var(--spacing-gap-xs)")};
+  gap: ${({ size }) =>
+    ["medium", "small"].includes(size?.height ?? "") ? "var(--spacing-gap-xs)" : "var(--spacing-gap-s)"};
```
Mil4n0r, line 21: Don't you think using includes for such a simple validation is unnecessary? I would honestly rather check the value of size.height with an `OR` operator as it would be much more readable and a bit more efficient. (Same applies for the rest in the same file (`font-size`, `height`, `width`)

```
gap: ${({ size }) =>
  size?.height === "medium" || size?.height === "small"
    ? "var(--spacing-gap-xs)"
    : "var(--spacing-gap-s)"};
```

packages/lib/src/button/utils.ts
```diff
@@ -13,7 +13,7 @@ export const getButtonStyles = (
 
   const commonStyles = `
     font-family: var(--typography-font-family);
-    font-size: var(${size?.height === "large" ? "--typography-label-l" : "--typography-label-m"});
+    font-size: var(${["medium", "small"].includes(size?.height ?? "") ? "--typography-label-m" : "--typography-label-l"});
```
Mil4n0r, line 16: Same as the feedback in `Button.tsx`

Mil4n0r (changes requested):

packages/lib/src/button/Button.tsx
GomezIvann: There's no difference in terms of performance since the length of the array is just two, it was just a matter of code preference from my side.

GomezIvann: In any case, I don't have a strong preference.

Mil4n0r (approved):

Mil4n0r (approved):

PR #2150, opened 2025-02-04
Alert redesign

Mil4n0r:
**Checklist**

**Description**
Redesign of the alert component following the new design kit specifications, already using the new token architecture (CSS variables).

**Additional context**


packages/lib/src/action-icon/ActionIcon.tsx
```diff
@@ -11,43 +10,33 @@ const ActionIcon = styled.button`
...
-        background-color: ${props.theme.hoverActionBackgroundColor ?? CoreTokens.color_grey_100};
-        color: ${props.theme.hoverActionIconColor ?? CoreTokens.color_grey_900};
+        background-color: var(--color-bg-alpha-light);
+        color: var(--color-fg-neutral-dark);
       }
       &:active {
-        background-color: ${props.theme.activeActionBackgroundColor ?? CoreTokens.color_grey_300};
-        color: ${props.theme.activeActionIconColor ?? CoreTokens.color_grey_900};
+        outline: var(--border-width-m) var(--border-style-default) var(--border-color-secondary-medium);
+        color: var(--color-fg-neutral-dark);
```
GomezIvann, line 33: Why are you setting color in every state of the action icon since it never changes (for example, you don't do that with the hover and the active, which share the same background color)? Same with the outline, I would just have it on the focus selector.

packages/lib/src/alert/Alert.tsx
```diff
@@ -17,66 +16,67 @@ const AlertContainer = styled.div<{
...
 `;
 
 const typographyStyles = css`
-  font-family: ${(props) => props.theme.fontFamily};
-  font-size: ${(props) => props.theme.fontSize};
-  font-style: ${(props) => props.theme.fontStyle};
-  font-weight: ${(props) => props.theme.fontWeight};
-  color: ${(props) => props.theme.fontColor};
+  font-family: var(--typography-font-family);
+  font-style: normal;
```
GomezIvann, line 53: `font-style: normal;` is the default value, we do not need to set it!

Mil4n0r:
Thank you Iván! Fixed both comments. Now I am going to start adding the new documentation 🚀

packages/lib/src/alert/types.ts
```diff
@@ -1,4 +1,4 @@
-import { ReactNode } from "react";
+import { KeyboardEvent, ReactNode } from "react";
```
GomezIvann, line 1: `KeyboardEvent` type is not been used.

packages/lib/src/alert/Alert.tsx
```diff
@@ -157,91 +156,94 @@ export default function DxcAlert({
     if (currentIndex === messages.length) handlePrevOnClick();
   }, [currentIndex, messages, handlePrevOnClick]);
 
+  useEffect(() => {
```
GomezIvann, line 159: To ease the component load a bit, does it make sense to move this `useEffect` to the `ModalAlertWrapper` component?

GomezIvann: In any case, to avoid unnecessary re-renders, we need to wrap `handleOnClose` in a `useCallback`.

GomezIvann (approved):

PR #2167, opened 2025-02-19
Checkbox redesign

Mil4n0r:
**Checklist**

**Description**
Redesign of the checkbox component following the new design kit specifications, already using the new token architecture (CSS variables).

**Additional context**


apps/website/pages/components/checkbox/code.tsx
```diff
@@ -0,0 +1,19 @@
+import Head from "next/head";
+import type { ReactElement } from "react";
+import CheckboxPageLayout from "screens/components/checkbox/CheckboxPageLayout";
+import CheckboxCodePage from "screens/components/checkbox/code/CheckboxCodePage";
+
+const Code = () => {
+  return (
```
GomezIvann, line 7: Remove the return statement!

apps/website/screens/components/checkbox/code/CheckboxCodePage.tsx
```diff
@@ -23,14 +23,14 @@ const sections = [
...
             <td>
-              <TableCode>boolean</TableCode>
+              <TableCode>string</TableCode>
             </td>
-            <td>Initial state of the checkbox, only when it is uncontrolled.</td>
             <td>
-              <TableCode>false</TableCode>
+              Specifies a string to be used as the name for the checkbox element when no <Code>label</Code> is provided.
             </td>
+            <td>'Checkbox'</td>
```
GomezIvann, line 33: This string should be surrounded by a `TableCode` component.

apps/website/screens/components/checkbox/overview/CheckboxOverviewPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+import anatomy from "./images/checkbox_anatomy.png";
+
+const sections = [
+  {
+    title: "Introduction",
+    content: (
+      <>
+        <DxcParagraph>
+          Checkboxes support different states, including checked, unchecked, and indeterminate, providing clear visual
+          feedback. Checkboxes should be used when multiple selections are needed, unlike radio buttons, which are for
```
GomezIvann, line 17: **"should be used when multiple selections are needed"** should be bold.

GomezIvann: image

apps/website/screens/components/checkbox/overview/CheckboxOverviewPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+            are keyboard-navigable. Labels should be linked correctly for screen readers to interpret them properly.
+          </DxcBulletedList.Item>
+        </DxcBulletedList>
+      </>
+    ),
+  },
+];
+
+const CheckboxInputOverviewPage = () => {
+  return (
```
GomezIvann, line 118: We can remove this return statement.

packages/lib/src/checkbox/Checkbox.tsx
```diff
@@ -1,115 +1,9 @@
...
-              {(checked ?? innerChecked) && checkedIcon}
-            </Checkbox>
-          </CheckboxContainer>
-        </MainContainer>
-      </ThemeProvider>
-    );
-  }
-);
-
 const sizes = {
```
GomezIvann, line 7: `sizes`, `spaces`, `calculateWidth`, `getDisabledColor`, and `getEnabledColor` can be moved to a `utils.ts` file for cleaner code!

packages/lib/src/checkbox/Checkbox.tsx
```diff
@@ -268,20 +179,134 @@ const MainContainer = styled.div<{
...
+              aria-readonly={readOnly}
+              aria-required={!disabled && !optional}
+              aria-labelledby={label ? labelId : undefined}
+              aria-label={label ? undefined : ariaLabel}
+              checked={checked ?? innerChecked}
+              disabled={disabled}
+              readOnly={readOnly}
+              ref={checkboxRef}
+            >
+              {(checked ?? innerChecked) && checkedIcon}
```
GomezIvann, line 303: After looking at the designs, I saw a few issues in the styling, like:

image

(The positioning of the check is not well placed.)

Maybe we can simplify this code and do it as in the designs, with two icons and a switch between them depending on the status of the `Checkbox`:

image
image

This would make our code much easier and solve the issues mentioned!

packages/lib/src/checkbox/Checkbox.tsx
```diff
@@ -268,20 +179,134 @@ const MainContainer = styled.div<{
...
+      optional = false,
+      readOnly = false,
+      onChange,
+      margin,
+      size = "fitContent",
+      tabIndex = 0,
+      ariaLabel = "Checkbox",
+    },
+    ref
+  ): JSX.Element => {
```
GomezIvann, line 232: Remove the return type.

packages/lib/src/checkbox/Checkbox.tsx
```diff
@@ -268,20 +179,134 @@ const MainContainer = styled.div<{
...
       if (!props.disabled)
-        return props.readOnly
-          ? getReadOnlyColor(props.theme, "hoverBackground")
-          : getEnabledColor(props.theme, "hoverBackground");
+        return props.readOnly ? getReadOnlyColor("hoverBackground") : getEnabledColor("hoverBackground");
+    }};
+  }
+
+  &:active ${Checkbox} {
+    border: var(--border-width-m) solid
```
GomezIvann, line 193: `solid` --> `var(--border-style-default)`

packages/lib/src/checkbox/Checkbox.tsx
```diff
@@ -268,20 +179,134 @@ const MainContainer = styled.div<{
   cursor: ${(props) => (props.disabled ? "not-allowed" : props.readOnly ? "default" : "pointer")};
 
   &:hover ${Checkbox} {
-    border: 2px solid
+    border: var(--border-width-m) solid
```
GomezIvann, line 182: `solid` --> `var(--border-style-default)`

packages/lib/src/checkbox/Checkbox.tsx
```diff
@@ -210,37 +121,37 @@ const Checkbox = styled.span<{
...
     ${(props) =>
       props.disabled
-        ? getDisabledColor(props.theme, "border")
+        ? getDisabledColor("border")
         : props.readOnly
-          ? getReadOnlyColor(props.theme, "border")
-          : getEnabledColor(props.theme, "border")};
+          ? getReadOnlyColor("border")
+          : getEnabledColor("border")};
   border-radius: 2px;
```
GomezIvann, line 131: `2px` --> `var(--border-radius-xs)`

packages/lib/src/checkbox/Checkbox.tsx
```diff
@@ -195,7 +106,7 @@ const CheckboxContainer = styled.span`
   align-items: center;
   justify-content: center;
   height: 24px;
-  width: 24px;
+  width: var(--height-s);
```
GomezIvann, line 109: The token should be on `height` instead of on the `width`.

GomezIvann (changes requested):

packages/lib/src/checkbox/Checkbox.tsx
Mil4n0r: As far as I know, this happens for all the components that are using outline and is not related to the checkbox itself, as they are rendered outside the box model so it has some inconsistencies when zooming at certain levels. It would require some serious work to replace them in all the focusable components of the library so we should discuss it.

GomezIvann: I know the error you refer to, but this is not the case. Even without zooming, the outline still overlaps, and I think it is a matter of positioning and can be fixed!

Mil4n0r: I just did the changes, however, in my local, for some zoom levels it still happens, we can discuss it in a call.

GomezIvann: Fixed!

GomezIvann (dismissed):

packages/lib/src/checkbox/utils.tsx
Mil4n0r, line 1: Why rename to tsx if we are not returning any JSX?

packages/lib/src/checkbox/Checkbox.tsx
Mil4n0r, line 1: Is it not a bit weird to have the same color for `active` and `focus`. In other components such as `Button` the color is darker in the `active` state (and this makes sense as a visual indicator that it is being pressed. I know this is design related but I would like to confirm it.

packages/lib/src/checkbox/utils.tsx
GomezIvann: The only way to assign an SVG to a variable is to have it as a tsx file (considering it as a component).

packages/lib/src/checkbox/Checkbox.tsx
Mil4n0r: Sorry I meant `hover`, not `focus`. It turns out that both `active` and `hover` are purposedly equals.

GomezIvann (approved):

PR #2185, opened 2025-03-11
Wizard redesign

GomezIvann:
**Checklist**

**Description**
Redesign of the wizard component following the new design kit specifications, already using the new token architecture (CSS variables).

**Additional context**
[Figma designs](https://www.figma.com/design/68NdlTuOcN4kpYxUguJqdf/Halstack-DS-%C2%B7-UI-Library?node-id=8674-6515&m=dev)


packages/lib/src/wizard/Wizard.tsx
```diff
@@ -1,303 +1,231 @@
...
+  align-items: center;
+  gap: var(--spacing-gap-s);
+  background-color: transparent;
+  border: none;
+  border-radius: var(--border-radius-s);
+  margin: ${({ mode }) =>
+    mode === "horizontal"
+      ? "var(--spacing-padding-none) var(--spacing-padding-l)"
+      : "var(--spacing-padding-l) var(--spacing-padding-none)"};
+  padding: var(--spacing-padding-none);
```
Mil4n0r, line 92: Any reason why you are using margin instead of padding? I don't see any focus examples in design so I am not sure what is the expected property for this.

packages/lib/src/wizard/Wizard.tsx
```diff
@@ -1,303 +1,231 @@
...
-  justify-content: center;
-  align-items: center;
-  overflow: hidden;
-  font-size: ${(props) => props.theme.stepIconSize};
+  box-sizing: border-box;
+  display: grid;
+  place-items: center;
+  border-radius: 50%;
+  border: var(--border-width-m) var(--border-style-default) var(--border-color-neutral-dark);
+  height: 32px;
```
Mil4n0r, line 46: `--height-m`

packages/lib/src/wizard/Wizard.tsx
```diff
@@ -1,303 +1,231 @@
...
-  margin-top: ${(props) =>
-    props.margin && typeof props.margin === "object" && props.margin.top ? spaces[props.margin.top] : ""};
-  margin-right: ${(props) =>
-    props.margin && typeof props.margin === "object" && props.margin.right ? spaces[props.margin.right] : ""};
-  margin-bottom: ${(props) =>
-    props.margin && typeof props.margin === "object" && props.margin.bottom ? spaces[props.margin.bottom] : ""};
-  margin-left: ${(props) =>
-    props.margin && typeof props.margin === "object" && props.margin.left ? spaces[props.margin.left] : ""};
+  ${({ mode }) => mode === "vertical" && "height: 100%; width: fit-content;"};
+  margin: ${({ margin }) => (margin && typeof margin !== "object" ? spaces[margin] : "0px")};
```
Mil4n0r, line 19: `--spacing-padding-none` ?

packages/lib/src/wizard/Wizard.tsx
```diff
@@ -1,303 +1,231 @@
...
 import DxcIcon from "../icon/Icon";
-import HalstackContext from "../HalstackContext";
 import WizardPropsType, { StepProps } from "./types";
+import DxcFlex from "../flex/Flex";
 import icons from "./Icons";
+import { DxcInset } from "..";
 
-const StepsContainer = styled.div<{
-  mode: WizardPropsType["mode"];
+const Wizard = styled.div<{
```
Mil4n0r, line 11: Is this container supposed to have fixed height?

Mil4n0r (changes requested):

packages/lib/src/wizard/Wizard.tsx
GomezIvann: If there's no value, I prefer that the the margin CSS property not appear in the style of the element; I will write `""`.

packages/lib/src/wizard/Wizard.tsx
GomezIvann: I don't think so, since the wizard can be both vertical and horizontal, I don't think we should limit the height.

Mil4n0r (approved):

PR #2203, opened 2025-03-31
Grid redesign

GomezIvann:
**Checklist**

**Description**
Redesign of the grid component following the new design kit specifications, already using the new token architecture (CSS variables).


apps/website/screens/components/grid/overview/GridOverviewPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </>
+    ),
+  },
+  {
+    title: "Best practices",
+    content: (
+      <DxcBulletedList>
+        <DxcBulletedList.Item>
+          <strong>Use grid for complex layouts:</strong> the grid component is ideal for creating complex layouts with
+          multiple rows and columns. It allows you to create responsive designs that adapt to different screen sizes.
```
Mil4n0r, line 56: This is a bit redundant with the `grid layout` section's first paragraph

apps/website/screens/components/grid/overview/GridOverviewPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </DxcBulletedList.Item>
+        <DxcBulletedList.Item>
+          <strong>Keep it simple:</strong> while the grid component is powerful, it's essential to keep your layouts as
+          simple as possible. Avoid overcomplicating your designs with too many nested grids or complex structures,
+          which can reduce the readability and maintainability of your code.
+        </DxcBulletedList.Item>
+        <DxcBulletedList.Item>
+          <strong>Use grid items correctly:</strong> they are represented by the <Code>DxcGrid.Item</Code> tag and
+          should be used when the decision to follow the Grid layout pattern is assumed only by the parent container.
```
Mil4n0r, line 65: This sounds a bit confusing. I would put something like:
"Use grid items correctly: The `DxcGrid.Item` tag should be used when the parent container dictates the Grid layout pattern."

apps/website/screens/components/grid/overview/GridOverviewPage.tsx
GomezIvann: I know, but I wanted to be consistent with the flex, which also has a first introductory point.

apps/website/screens/components/grid/overview/GridOverviewPage.tsx
GomezIvann: I know it may be unclear, but it was done on purpose. What I meant with this point was that you should only use the `DxcGrid.Item` when the grid pattern is only assumed for the parent since the item is a regular container with the properties of a grid item, but the display is not set to the grid. If the child also follows a grid pattern, you should again use the `DxcGrid`, which also has every prop the `DxcGrid.Item` allows.

PR #2204, opened 2025-03-31
Inset redesign

GomezIvann:
**Checklist**

- [ ] The build process is done without errors. All tests pass in the `/lib` directory.

**Description**
Redesign of the inset component following the new design kit specifications, already using the new token architecture (CSS variables). This PR also includes several documentation updates in the `DxcContainer` that weren't addressed in its PR.

- Now the `children` prop from the `**<DxcContainer>**` is not mandatory anymore. This will allow users to build blocks without children in Halstack layouts.


packages/lib/src/button/Button.stories.tsx
```diff
@@ -4934,7 +4934,7 @@ const NestedTooltip = () => (
   <>
     <Title title="Nested tooltip" theme="light" level={2} />
     <ExampleContainer>
-      <DxcInset top="3rem">
+      <DxcInset top="var(--spacing-padding-xl)">
```
Mil4n0r, line 4937: I think you meant xxl here

packages/lib/src/inset/Inset.tsx
```diff
@@ -1,28 +1,19 @@
...
-    bottom={bottom}
-    left={left}
-  >
-    {children}
-  </StyledInset>
-);
-
-export default Inset;
+export default function DxcInset({ bottom, children, horizontal, left, right, space, top, vertical }: InsetPropsType) {
+  return (
```
Mil4n0r, line 5: We could remove the return statement and put it directly.

Mil4n0r (changes requested):

packages/lib/src/button/Button.stories.tsx
GomezIvann: Yes, my fault!

packages/lib/src/inset/Inset.tsx
GomezIvann: We cannot do an export default in one line with the component definition without using `function`, which cannot have one line arrow function bodies.

Mil4n0r (approved):

PR #2211, opened 2025-04-11
Nav tab fixes and doc updates

Jialecl:
**Checklist**

**Description**
- Documentation updates: 
  - New best practice for the Tabs and Nav tabs regarding icon usage.
  - Tabs best practices restructured to match the format from other components.
  - Small typos fixed.
- Nav tabs `iconPosition` prop default changed from `'top'` to `'left'`, as in the Tabs.
- Fixed bug with the outline and with the active tab underline.


GomezIvann (dismissed):

packages/lib/src/nav-tabs/NavTabs.tsx
```diff
@@ -1,79 +1,32 @@
...
-const Underline = styled.div<{ underlineWidth: number }>`
+const Underline = styled.div`
   position: absolute;
   bottom: 0;
   left: 0;
   height: var(--border-width-m);
   background-color: var(--border-color-neutral-medium);
-  z-index: -1;
-  width: ${(props) => props.underlineWidth}px;
+  width: 100%;
```
Jialecl, line 20: this change causes issues when the nav tabs don't have enough space
image

packages/lib/src/nav-tabs/Tab.tsx
```diff
@@ -5,57 +5,55 @@ import DxcFlex from "../flex/Flex";
...
 }>`
   box-sizing: border-box;
   display: flex;
-  flex-direction: ${(props) => (props.hasIcon && props.iconPosition === "top" ? "column" : "row")};
+  flex-direction: ${({ iconPosition }) => (iconPosition === "top" ? "column" : "row")};
   justify-content: center;
   align-items: center;
   gap: var(--spacing-gap-xs);
-  height: ${(props) => (props.hasIcon && props.iconPosition === "top" ? "78px" : "100%")};
+  height: ${({ iconPosition }) => (iconPosition === "top" ? "78px" : "100%")};
```
Jialecl, line 26: This has the unintended effect that the height can change even if none of the tabs has an icon.

GomezIvann: I know, but it makes zero sense if you have a tab without an icon and `iconPosition` set on purpose to `"top"` right?

packages/lib/src/nav-tabs/NavTabs.tsx
GomezIvann: Solved!

packages/lib/src/nav-tabs/Tab.tsx
GomezIvann: Also, I understand that if you set `iconPosition` to top for the whole component is that you anticipate that it can go with an icon, so if the height it's already set it won't do a size "jump".

Jialecl:
With the new approach to the underline we no longer need the styled component `Underline`

PR #2217, opened 2025-04-22
Header redesign

Mil4n0r:
**Checklist**

**Description**
Redesign of the header component following the new design kit specifications, already using the new token architecture (CSS variables).

**Additional context**


Mil4n0r:
Still have some TODOs which require designer feedback. Also missing doc and API updates for logo.

packages/lib/src/header/Header.tsx
```diff
@@ -166,34 +23,34 @@ const HeaderContainer = styled.header<{
   margin: HeaderPropsType["margin"];
   underlined: HeaderPropsType["underlined"];
 }>`
+  background-color: var(--color-bg-neutral-lightest);
+  /* TODO: ASK ABOUT UNDERLINE */
+  border-bottom: ${(props) =>
+    props.underlined && `var(--border-width-m) var(--border-style-default) ${props.theme.underlinedColor}`};
```
Jialecl, line 29: the correct token to be used for the color of the underline is `var(--border-color-neutral-strongest)` as per the design.

packages/lib/src/header/Header.tsx
```diff
@@ -166,34 +23,34 @@ const HeaderContainer = styled.header<{
...
   justify-content: space-between;
-  min-height: ${(props) => props.theme.minHeight};
   margin-bottom: ${(props) => (props.margin ? spaces[props.margin] : "0px")};
-  padding: ${(props) =>
-    `${props.theme.paddingTop} ${props.theme.paddingRight} ${props.theme.paddingBottom} ${props.theme.paddingLeft}`};
-  background-color: ${(props) => props.theme.backgroundColor};
-  border-bottom: ${(props) =>
-    props.underlined &&
-    `${props.theme.underlinedThickness} ${props.theme.underlinedStyle} ${props.theme.underlinedColor}`};
+  min-height: 64px;
```
Jialecl, line 36: should we consider a new alias for this min-height?

packages/lib/src/header/Header.tsx
```diff
@@ -210,89 +67,93 @@ const ContentContainer = styled.div`
...
   flex-direction: column;
+  font-family: var(--typography-font-family);
+  font-size: var(--typography-label-s);
+  font-weight: var(--typography-label-semibold);
+  height: var(--height-xl);
   justify-content: center;
-  align-items: center;
+  padding: var(--spacing-padding-none) var(--spacing-padding-m);
+  text-transform: uppercase;
   width: 54px;
```
Jialecl, line 89: This width value is no longer correct

packages/lib/src/header/Header.tsx
```diff
@@ -210,89 +67,93 @@ const ContentContainer = styled.div`
...
   width: calc(100% - 186px);
-  color: ${(props) => props.theme.contentColor};
+  color: var(--color-fg-neutral-dark);
 `;
 
 const HamburgerTrigger = styled.button`
+  align-items: center;
+  background-color: transparent;
+  border-radius: var(--border-radius-xs);
+  border: var(--border-width-s) var(--border-style-default) transparent;
```
Jialecl, line 77: I don´t think a border is needed

packages/lib/src/header/Header.tsx
```diff
@@ -210,89 +67,93 @@ const ContentContainer = styled.div`
...
-  border: 1px solid transparent;
-  border-radius: 2px;
-  background-color: transparent;
   :hover {
-    background-color: ${(props) => props.theme.hamburgerHoverColor};
+    background-color: var(--color-bg-neutral-medium);
   }
   &:focus {
-    outline: ${(props) => props.theme.hamburgerFocusColor} auto 1px;
+    outline: var(--border-color-secondary-medium) solid var(--border-width-m);
```
Jialecl, line 94: `solid` should be a token

packages/lib/src/header/Header.tsx
Mil4n0r: yes it is needed, html button does have a border by default.

packages/lib/src/header/Header.tsx
```diff
@@ -321,7 +183,153 @@ const Overlay = styled.div<{ hasVisibility: boolean }>`
...
+              <DxcIcon icon="menu" />
+              {translatedLabels.header.hamburgerTitle}
+            </HamburgerTrigger>
+          </ChildContainer>
+          <ResponsiveMenu hasVisibility={isMenuVisible}>
+            <DxcFlex justifyContent="space-between" alignItems="center">
+              <ResponsiveLogoContainer>{headerResponsiveLogo}</ResponsiveLogoContainer>
+              <Tooltip label={translatedLabels.header.closeIcon}>
+                <CloseAction tabIndex={tabIndex} onClick={handleMenu} aria-label={translatedLabels.header.closeIcon}>
+                  <DxcIcon icon="close" />
```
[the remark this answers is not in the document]
Mil4n0r: I just added a comment in Figma asking if we could do that.

packages/lib/src/header/Header.tsx
Mil4n0r: I would discuss this in a future iteration. For now it should be fine, at the end of tokens implementation we can search and replace all the pending "em",  "px", etc. units. It is a good point though.

PR #2227, opened 2025-05-15
TabId prop added to identify each tab

Jialecl:
**Checklist**

**Purpose**
With this change tabs can be used without a label as it is documented.

**Description*
This is a major change due to a required prop `tabId` being added to the `DxcTabs.Tab` component. This prop allows the tabs component to identify each tab even if there is not a label present.

### DIFF
--- a/apps/website/screens/components/tabs/code/TabsCodePage.tsx
+++ b/apps/website/screens/components/tabs/code/TabsCodePage.tsx
@@ -304,6 +304,7 @@ const sections = [
               <tr>
                 <td>
                   <DxcFlex direction="column" gap="var(--spacing-gap-xs)" alignItems="baseline">
+                    <StatusBadge status="required" />
                     tabId
                   </DxcFlex>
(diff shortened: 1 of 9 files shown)


apps/website/screens/components/tabs/code/TabsCodePage.tsx
```diff
@@ -301,6 +301,19 @@ const sections = [
                 <td>This function will be called when the user hovers this tab.</td>
                 <td>-</td>
               </tr>
+              <tr>
+                <td>
+                  <DxcFlex direction="column" gap="var(--spacing-gap-xs)" alignItems="baseline">
+                    <StatusBadge status="required" />
+                    tabId
```
Mil4n0r, line 308: Why don't we use just `id` instead? Seems more intuitive.

Mil4n0r (changes requested):

Mil4n0r (approved):

PR #2229, opened 2025-05-21
Added accordion stories

raquelarrojo:
**Checklist**

**Description**
Added a special story in the Accordion when the sublabel is larger than the label so that  we can confirm that the label position is fine.

### DIFF
--- a/packages/lib/src/accordion/Accordion.stories.tsx
+++ b/packages/lib/src/accordion/Accordion.stories.tsx
@@ -335,20 +335,13 @@ const Accordion = () => (
       <Title title="Sublabel longer than label" theme="light" level={4} />
       <DxcAccordion>
         <DxcAccordion.AccordionItem
-          label="Avg, actions per session"
-          subLabel="Tue, May 20, 10:11 AM"
+          label="Bounce Rate"
+          subLabel="Mon, May 19, 3:17 PM"
           badge={{ position: "after", element: <DxcBadge label="Resolved" icon="check_circle" color="green" /> }}
         >
           <DxcInset space="var(--spacing-padding-l)">
             To edit your profile you need to go to Settings and click on Profile.
           </DxcInset>
-        </DxcAccordion.AccordionItem>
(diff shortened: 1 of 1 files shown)


packages/lib/src/accordion/Accordion.stories.tsx
```diff
@@ -331,6 +331,27 @@ const Accordion = () => (
...
+      <DxcAccordion>
+        <DxcAccordion.AccordionItem
+          label="Avg, actions per session"
+          subLabel="Tue, May 20, 10:11 AM"
+          badge={{ position: "after", element: <DxcBadge label="Resolved" icon="check_circle" color="green" /> }}
+        >
+          <DxcInset space="var(--spacing-padding-l)">
+            To edit your profile you need to go to Settings and click on Profile.
+          </DxcInset>
+        </DxcAccordion.AccordionItem>
```
Mil4n0r, line 345: Is there any reason to add this additional accordion item?

Mil4n0r (changes requested):

packages/lib/src/accordion/Accordion.stories.tsx
raquelarrojo: I agree! I only kept the one in which the sublabel is longer than the label. Thanks!

packages/lib/src/accordion/Accordion.stories.tsx
```diff
@@ -331,6 +331,27 @@ const Accordion = () => (
...
+        <DxcAccordion.AccordionItem
+          label="Bounce Rate"
+          subLabel="Mon, May 19, 3:17 PM"
+          badge={{ position: "after", element: <DxcBadge label="Resolved" icon="check_circle" color="green" /> }}
+        >
+          <DxcInset space="1rem">To edit your profile you need to go to Settings and click on Profile.</DxcInset>
```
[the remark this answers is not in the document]
raquelarrojo: Done!

Mil4n0r (approved):

PR #2232, opened 2025-05-23
`tabId` prop is no longer required

Jialecl:
**Checklist**

**Purpose**
To avoid creating a breaking change for now we will default the `tabId` value to the label if this one is set.

### DIFF
--- a/packages/lib/src/tabs/Tab.tsx
+++ b/packages/lib/src/tabs/Tab.tsx
@@ -155,6 +155,7 @@ const DxcTab = forwardRef(
           role="tab"
           tabIndex={activeTabId === label && !disabled ? tabIndex : -1}
           type="button"
+          aria-label={label ?? tabId ?? "tab"}
         >
           <LabelIconContainer iconPosition={iconPosition}>
             {icon && <IconContainer>{typeof icon === "string" ? <DxcIcon icon={icon} /> : icon}</IconContainer>}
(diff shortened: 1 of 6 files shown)


packages/lib/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
-  "extends": "@dxc-technology/typescript-config/react-library.json",
+  "extends": "../typescript-config/react-library.json",
```
Mil4n0r, line 2: For consistency with the tsconfig.json from website, I would leave it as "@dxc-technology/typescript-config/react-library.json", the idea is to have centralized config files which can be used across all the packages, so it's easier to use it that way instead of having to figure out the relative route.

packages/lib/src/tabs/Tab.tsx
Mil4n0r, line 1: You have to add an aria-label to `<Tab />` because it is a button and requires an accessible name (which is no longer going to be met due to tabs possibly having no label).

That should fix the accessibility issues.

Mil4n0r (changes requested):

Mil4n0r (changes requested):
Looks good to me. However, can you do a change in the stories so that we have both an example of top/left icon position with text and another one with no text? (instead of replacing it like you have done). That way we canmake sure that we are covering all scenarios.

Mil4n0r (approved):

PR #2238, opened 2025-06-02
Fix for `DxcDatagrid` state maintenance using hierarchy + selectable rows

Mil4n0r:
**Checklist**


apps/website/screens/components/data-grid/code/examples/hierarchicalSelectable.ts
```diff
@@ -1,8 +1,8 @@
 import { DxcDataGrid, DxcInset } from "@dxc-technology/halstack-react";
-import { useState } from "react";
+import { useMemo, useState } from "react";
 
 const code = `() => {
-  const columns = [
+  const columns = useMemo(() => [
```
Jialecl, line 5: Instead of useMemo I think useState without the setter is more correct for this case.

PR #2241, opened 2025-06-11
Implementing a trigger event for hierarchical datagrid expand

Mil4n0r:
**Checklist**

**Purpose**
The idea of the change is being able to handle highly nested data in a more appropiate way. With the previous API we were forcing the users to load the whole dataset inside the rows item from the `DxcDatagrid`. Which is not feasible most of the time due to big performance impact. 

**Description**
In order to achieve that purpose, a new prop has been added to the component, `loadChildren?: (expandChildren: HierarchyGridRow) => HierarchyGridRow[] | Promise<HierarchyGridRow[]>;`. As its name says, it is an additional way of setting the children for a hierarchical data, this data can come from any type of source and is specially useful for lazy loading purposes (until you click on the expand button, the children data for that node is not processed).

**Additional context**
There are still some edge cases which have to be reviewed as well as including additional documentation, testing and stories for the component.


Mil4n0r:
Should we add a mechanism to display a spinner while the data is being loaded when setting the loadChildren function?

Jialecl:
> Should we add a mechanism to display a spinner while the data is being loaded when setting the loadChildren function?

They are able to use the `Spinner` component easily without any mechanism from the datagrid

Jialecl (commented):
We need to discuss about this approach because I find issues with delays in case the children are not returned instantly.

packages/lib/src/data-grid/DataGrid.stories.tsx
```diff
@@ -997,6 +1060,15 @@ export const Chromatic: Story = {
 
 export const Controlled: Story = {
   render: DataGridControlled,
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    await userEvent.click(canvas.getByText("Root Node 1 Lazy"));
+    await userEvent.click(canvas.getByText("Root Node 2 Lazy"));
+    // await userEvent.click(canvas.getByText("Root Node 1 Lazy Child 1"));
```
Jialecl, line 1067: Remove comments

Jialecl:
Updates missing in the documentation.

packages/lib/src/data-grid/types.ts
```diff
@@ -48,7 +48,30 @@ export type GridRow = {
 };
 
 export type HierarchyGridRow = GridRow & {
+  /**
+   * Array of child rows nested under this row, enabling hierarchical (tree-like) structures.
+   * These child rows will be displayed when the parent row is expanded.
+   */
   childRows?: HierarchyGridRow[] | GridRow[];
+  /**
+   * Function called whenever a cell with children is expanded or collapsed. Returns the children array
```
Jialecl, line 57: This comment is too simple. We should document more the return.

Jialecl:
There are still problems with the selection.
You can select other rows or the select all row and it won't be taken into account once the rows are rendered.
Maybe we should consider a global state for all the checkboxes instead of a row state.

Jialecl:
There is just one small bug I was able to reproduce:
image

It seems to be related to collapsed grandchildren while selected are stored in the selected row set and stayed on the list after unselected.

PR #2297, opened 2025-10-02
Added virtualization to Paginator Select

Mil4n0r:
**Checklist**

**Description**
We already support virtualization for `Select` but `Paginator` was not making use of it. I have defined a threshold in which arrays with size higher than 100, where the performance starts to become worse, it is appropiate to incorporate it, using a `virtualHeight` of 304px which is the `max-height` that we previously supported on the Select.

**Closes**
[OC-19102](https://jira.dxc.com/browse/OC-19102)


Mil4n0r:
We can discuss the threshold in which we start virtualizing.

PR #2313, opened 2025-10-15
Fix scrolling when navigating between pages on the documentation site.

PelayoFelgueroso:
**Checklist**

**Purpose**
Scroll to Top when we switch from when page to another


Mil4n0r (changes requested):
No changes are required in `_app` or `MainContent`.

`MainContent` is causing this problem because it is part of the `ApplicationLayout` itself, and it is the one handling the scroll, not `ApplicationLayout.Main`. the thing is that we would need to have a ref to it (instead of an id). 

By having the current `useRef` from `ApplicationLayout` pointing the `MainContent` instead of the whole content, we would be able to add inside the same file something like this:

```
  useEffect(() => {
    if (ref.current) {
      ref.current.scrollTop = 0;
    }
  }, [children]);
```

We should discuss it with @Jialecl to make sure it does not break anything though.

apps/website/screens/common/MainContent.tsx
```diff
@@ -65,7 +65,7 @@ const MainContent = ({ children }: { children: ReactNode }) => {
     }
   }, [latestRelease, toast]);
 
-  return <MainContainer>{children}</MainContainer>;
+  return <MainContainer id="hello">{children}</MainContainer>;
```
Mil4n0r: This id is no longer needed.

Mil4n0r (changes requested):

Jialecl:
This approach messes with the navigation to different sections. For example: if you navigate to a link that looks like https://(,,,).com/(...)/#intro it won't scroll to #intro

PR #2318, opened 2025-10-22
Fix Quick Nav bug with HashRouter

PelayoFelgueroso:
**Checklist**

**[Closes](https://jira.dxc.com/browse/OC-19355)


Jialecl:
A problem with this implementation is that the `#` does not longer appear in the url when any link or sublink is clicked.
I don't think we want that behaviour.

PR #2330, opened 2025-11-11
Fix preventDefult on Breadcrumbs item 

PelayoFelgueroso:
…owsing to an unexisting page

**Checklist**

**Purpose**
Fix preventDefult on Breadcrumbs item and modify examples to avoid browsing to an unexisting page.


packages/lib/src/breadcrumbs/Item.tsx
```diff
@@ -48,8 +48,10 @@ const Item = ({ isCurrentPage = false, href, label, onClick }: ItemPropsType) =>
   const currentItemRef = useRef<HTMLSpanElement | null>(null);
 
   const handleOnClick = (event: MouseEvent<HTMLAnchorElement>) => {
-    event.preventDefault();
-    if (href) onClick?.(href);
+    if (onClick) {
+      event.preventDefault();
+      if (href) onClick?.(href);
```
Jialecl: There is no need to check if onClick does exist with a `?`.
And I think it is more correct to check the type of onClick. If (typeof onClick === "function")

PR #2356, opened 2025-12-01
Logo prop moved to the application layout

Jialecl:
**Checklist**

**Purpose**
Moved logo prop to the app layout. Logo will be present in 1 component: the header (with priority) or the sidenav.


packages/lib/src/layout/types.ts
```diff
@@ -18,7 +38,22 @@ export type AppLayoutSidenavPropsType = {
...
+   * Indicates if the header exists.
+   */
+  headerExists: boolean;
+};
+
 type ApplicationLayoutPropsType = {
+  /**
+   * Logo properties.
+   */
+  logo?: LogoPropsType;
```
PelayoFelgueroso: I think logo should be mandatory in the application layout

PR #2504, opened 2026-07-20
[minor] New Message Input component implementation

PelayoFelgueroso:
**Checklist**

**Purpose**
Implemente the new message input component based on the design.


packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+  const languageContext = useContext(HalstackLanguageContext);
+  const locale = languageContext.locale ? languageContext.locale : undefined;
+  const inputId = `input-${useId()}`;
+  const inputRef = useRef<HTMLTextAreaElement | null>(null);
+  const overlayRef = useRef<HTMLDivElement | null>(null);
+  const [innerValue, setInnerValue] = useState(defaultValue);
+  const [isFocused, setIsFocused] = useState(false);
+  const fileInputRef = useRef<HTMLInputElement | null>(null);
+  const { transcript, isRecording, startRecording, stopRecording, resetTranscript } = useVoiceTranscription({
+    lang: locale ?? "en-US",
```
Jialecl: Change it so that the default value of locale is "en-US" instead of defaulting to en-US if locale is undefined.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+  const handleInputContainerOnClick = () => inputRef.current?.focus();
+
+  const handleInputContainerOnMouseDown = (event: MouseEvent<HTMLDivElement>) => {
+    if (document.activeElement === inputRef.current) event.preventDefault();
+  };
+
+  const handleInputOnChange = (event: ChangeEvent<HTMLTextAreaElement>) => {
+    const el = event.target;
+    el.style.height = "auto";
```
Jialecl: There is no need to change the scrollHeight to achieve a growing textarea.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+  overflow-y: hidden;
+  padding-bottom: 4px;
+
+  ${scrollbarStyles};
+  &::-webkit-scrollbar {
+    height: 4px;
+  }
+`;
+
+const Input = styled.textarea<{ isRecording?: boolean }>`
```
Jialecl: This variable should be more descriptive. Input is just too generic.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+  ${({ disabled }) => disabled && "cursor: not-allowed;"}
+
+  ${scrollbarStyles};
+  &::-webkit-scrollbar {
+    width: 4px;
+  }
+`;
+
+const InputWrapper = styled.div`
+  max-height: 150px;
```
Jialecl: Where does this max-height comes from? Based on the design the max height is for the whole component, so the maximun height can change depending how many elements are inside the message input.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+  position: relative;
+  flex-grow: 1;
```
Jialecl: I do not see the point of this flex grow and direction.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+    if (!textarea || !overlay || !isRecording) return;
+
+    overlay.scrollTop = textarea.scrollTop;
+    const handleScroll = () => (overlay.scrollTop = textarea.scrollTop);
+
+    textarea.addEventListener("scroll", handleScroll);
+    return () => textarea.removeEventListener("scroll", handleScroll);
+  }, [isRecording]);
+
+  const handleSubmit = () => !disabled && !isGenerating && onButtonClick?.("submit");
```
Jialecl: handleSubmit and handleStop should not return any value.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+} as const;
+
+const MessageInputContainer = styled.div<{ size: PromptInputPropsType["size"] }>`
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-gap-xs);
+  max-height: 320px;
+  background-color: var(--color-bg-neutral-lightest);
+  box-shadow: 0 -24px 10px 4px rgba(255, 255, 255, 0.6);
```
Jialecl: This shadow does not make much sense. It does not match any of our shadow tokens, it is unnoticeable and it does not ake into account the border radius.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+const Input = styled.textarea<{ isRecording?: boolean }>`
+  min-height: 20px;
+  max-width: 100%;
+  background: none;
+  border: none;
+  outline: none;
+  padding: var(--spacing-padding-none) var(--spacing-padding-xs);
+  resize: none;
+  flex-grow: 1;
```
Jialecl: this does nothing at all.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+  }
+`;
+
+const Input = styled.textarea<{ isRecording?: boolean }>`
+  min-height: 20px;
+  max-width: 100%;
+  background: none;
+  border: none;
+  outline: none;
+  padding: var(--spacing-padding-none) var(--spacing-padding-xs);
```
Jialecl, line 76: Why is the padding in the textarea instead of the container?

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+  flex-direction: column;
+`;
+
+const TranscriptOverlay = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  padding: var(--spacing-padding-none) var(--spacing-padding-xs);
```
Jialecl: The same applies as the textarea. If the padding is the same and they share the same container, apply it to the container.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+            maxLength={maxLength}
+            minLength={minLength}
+            value={value ?? innerValue}
+          />
+          {isRecording && (
+            <TranscriptOverlay ref={overlayRef}>
+              {baseTextRef.current && <span>{baseTextRef.current}</span>}
+              {transcript && (
+                <>
+                  {baseTextRef.current && <span> </span>}
```
Jialecl: There are some ocasions that 2 spaces are added and I suspect is due to this empty space.
And too many containers are added I am not sure they are neccessary.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,404 @@
...
+                  )}
+                </>
+              )}
+            </TranscriptOverlay>
+          )}
+        </InputWrapper>
+        <DxcFlex justifyContent={modelList ? "space-between" : "flex-end"} alignItems="center">
+          {modelList && (
+            <DxcContainer width="35%" maxWidth="240px">
+              <DxcSelect
```
Jialecl, line 370: Based on the design I did see that the select seems to be able to show an error message.

packages/lib/src/message-input/types.ts
```diff
@@ -0,0 +1,114 @@
...
+   * comply the minimum length, the onBlur and onChange functions will be called
+   * with the current value and an internal error informing that the value
+   * length does not comply the specified range. If a valid length is
+   * reached, the error parameter of both events will not be defined.
+   */
+  minLength?: number;
+  /**
+   * Options to be shown on the dropdown under the input.
+   */
+  modelList?: {
```
Jialecl: Didn't we decide that this prop should not contain "model" since the usage of this component is not limited to promp inputs?

packages/lib/src/message-input/useVoiceTranscription.tsx
```diff
@@ -0,0 +1,128 @@
+import { useCallback, useRef, useState } from "react";
+
+interface SpeechRecognitionResult {
```
Jialecl: Did we decide to use interfaces?

packages/lib/src/message-input/useVoiceTranscription.tsx
```diff
@@ -0,0 +1,128 @@
...
+  readonly transcript: string;
+  readonly confidence: number;
```
Jialecl, line 5: I do not see confidence ever used. Review also the other types.

packages/lib/src/message-input/useVoiceTranscription.tsx
Jialecl, line 1: The whole trascription functionality does feel clunky.
It stops on its own and when that happens the component breaks.

packages/lib/src/message-input/utils.ts
Jialecl, line 1: There are a lot of functions that can be added to utils file to reduce visual clutter in the component.

packages/lib/src/select/utils.ts
```diff
@@ -6,7 +6,7 @@ const sizes = {
   small: "240px",
   medium: "360px",
   large: "480px",
-  fillParent: "100%",
+  fillParent: "fit-content",
```
Jialecl: Fill parent should always be 100% size.

packages/lib/src/message-input/MessageInput.stories.tsx
Jialecl, line 1: Review all the visual test cases, there are multiple that does not make sense as visual tests.

packages/lib/src/message-input/MessageInput.test.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+    Object.defineProperty(fileInput, "files", {
+      value: [file],
+      writable: false,
+    });
+
+    fileInput.dispatchEvent(new Event("change", { bubbles: true }));
+
+    expect(callbackFile).toHaveBeenCalled();
+  });
```
Jialecl, line 346: We should add a transcript test.

packages/lib/src/message-input/types.ts
```diff
@@ -0,0 +1,114 @@
...
+   * the error (if the value entered is not valid) will be passed to this
+   * function. If there is no error, error will not be defined.
+   */
+  onChange?: (val: { value: string; error?: string }) => void;
+  /**
+   * This function will be called when the user clicks on the button (submit or stop) or presses enter.
+   * The type parameter indicates whether it's a "submit" or "stop" event.
+   * For submit events, an AbortSignal is provided to allow cancellation.
+   */
+  onButtonClick?: (type: "submit" | "stop") => void;
```
Jialecl: Have we considered about passing all the relevant information on submit to the user in case the component is not controlled? That makes proccessing the input far easier.

packages/lib/src/message-input/useVoiceTranscription.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+      const fullTranscript = results
+        .map((result) => result[0]?.transcript ?? "")
+        .join(" ")
+        .trim();
+      setTranscript(fullTranscript);
+    };
+
+    recognition.onerror = (event: SpeechRecognitionErrorEvent) => {
+      console.error("Speech recognition error:", event.error);
```
Jialecl: This error should be sent to the message input to display any possible error related to the voice transcript.

Jialecl (changes requested):

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: because we need it for the scrollbar position

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: we also need it for the scrollbar

packages/lib/src/message-input/types.ts
PelayoFelgueroso: changed to selectOptions

packages/lib/src/message-input/useVoiceTranscription.tsx
PelayoFelgueroso: fixed using type

packages/lib/src/select/utils.ts
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.stories.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/useVoiceTranscription.tsx
PelayoFelgueroso: removed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/useVoiceTranscription.tsx
PelayoFelgueroso: fixed that

packages/lib/src/message-input/MessageInput.test.tsx
PelayoFelgueroso: added

packages/lib/src/message-input/useVoiceTranscription.tsx
PelayoFelgueroso: done

packages/lib/src/message-input/types.ts
PelayoFelgueroso: done

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,439 @@
...
+  }
+`;
+
+const InputWrapper = styled.div`
+  position: relative;
+  display: flex;
+`;
+
+const TranscriptOverlay = styled.div`
+  position: absolute;
```
Jialecl, line 106: If you keep talking and allowing the transcript to grow it covers the stop button, which does not allow the user to stop the transcript. It also affects the scrollbar behavior.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,439 @@
...
+  box-sizing: border-box;
+  position: relative;
+  display: grid;
+  grid-template-rows: ${({ hasFiles }) =>
+    hasFiles ? "minmax(36px, 40px) minmax(0, 1fr) 34px" : "minmax(0, 1fr) 34px"};
+  gap: var(--spacing-gap-s);
+  padding: var(--spacing-padding-m) var(--spacing-padding-xs);
+  background-color: var(--color-bg-neutral-lightest);
+  box-shadow: 0 -24px 10px 4px rgba(255, 255, 255, 0.6);
+  ${({ disabled, error, focus }) => inputStylesByStatePromptInput(disabled, error, focus)}
```
Jialecl, line 51: This function injects a different background color than the one defined a few lines above.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,439 @@
...
+  overflow: hidden;
+`;
+
+const FilesContainer = styled.div`
+  min-height: 32px;
```
Jialecl: This should not be needed since the trigger is already 32 px

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,439 @@
...
+  medium: "360px",
+  large: "480px",
+  fillParent: "100%",
+} as const;
+
+const MessageInputContainer = styled.div<{ size: PromptInputPropsType["size"] }>`
+  width: ${({ size = "medium" }) => sizes[size]};
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-gap-xs);
```
Jialecl: This gap is unnecessary, the error message does already have 4px of separation.

packages/lib/src/message-input/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+  else if (file.type.includes("image")) return "image";
+  else return "description";
+};
+
+export const inputStylesByStatePromptInput = (disabled: boolean, error: boolean, focus?: boolean) => css`
+  background-color: ${disabled ? `var(--color-bg-neutral-lightest)` : `transparent`};
+
+  border-radius: var(--border-radius-l);
+
+  border: ${!disabled && error ? "var(--border-width-m)" : "var(--border-width-s)"} var(--border-style-default)
```
Jialecl: in disabled state it does have a different bordero color

Jialecl (changes requested):

packages/lib/src/message-input/MessageInput.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+    <ExampleContainer>
+      <Title title="Error" level={4} />
+      <DxcMessageInput placeholder="Ask me anything..." error="Please enter a valid message" />
+    </ExampleContainer>
+    <ExampleContainer pseudoState="pseudo-hover">
+      <Title title="Hover error" level={4} />
+      <DxcMessageInput placeholder="Ask me anything..." error="Error Message" />
+    </ExampleContainer>
+
+    <Title title="Validation" theme="light" level={2} />
```
Jialecl: These two do not really make sense as visual tests

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: removed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: removed

packages/lib/src/message-input/utils.ts
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.stories.tsx
PelayoFelgueroso: removed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,439 @@
...
+    <MessageInputContainer size={size}>
+      <MessageInput
+        disabled={disabled}
+        error={!!error}
+        focus={isFocused}
+        hasFiles={files && typeof callbackFile === "function"}
+        onClick={handleInputContainerOnClick}
+        onMouseDown={handleInputContainerOnMouseDown}
+      >
+        {files && typeof callbackFile === "function" && (
```
Jialecl: Thinking about it even if it is controlled, users can define files as undefined first.

apps/website/screens/components/message-input/overview/MessageInputOverviewPage.tsx
Jialecl, line 1: I would add an explanation on how to enable files in the component.

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

apps/website/screens/components/message-input/overview/MessageInputOverviewPage.tsx
PelayoFelgueroso: done

packages/lib/src/message-input/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+export const inputStylesByStatePromptInput = (disabled: boolean, error: boolean, focus?: boolean) => css`
+  background-color: ${disabled
+    ? `var(--color-bg-neutral-lighter)`
+    : `  background-color: var(--color-bg-neutral-lightest);
+`};
+
+  border-radius: var(--border-radius-l);
+
+  border: ${!disabled && error ? "var(--border-width-m)" : "var(--border-width-s)"} var(--border-style-default)
+    ${(() => {
```
Jialecl: instead of creating a function this could be a ternary applied only to border color and leaving one as the default here.

packages/lib/src/message-input/types.ts
```diff
@@ -0,0 +1,112 @@
...
+type Props = {
+  /**
+   * If true, the voice recording button will be shown.
+   */
+  allowRecording?: boolean;
+  /**
+   * This function will be called when the selection of top items changes.
+   * If this function is provided the message input will allow file selection.
+   */
+  callbackFile: (files: File[]) => void;
```
Jialecl: callbackFile is not mandatory.

apps/website/screens/components/message-input/code/MessageInputCodePage.tsx
```diff
@@ -0,0 +1,252 @@
...
+            <th>Default</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td>allowRecording</td>
+            <td>
+              <TableCode>boolean</TableCode>
+            </td>
+            <td>If true, the voice recording button will be shown.</td>
```
Jialecl: Add a little explanation about the locale and how to change it linking to the HalstackProvider.

packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -0,0 +1,437 @@
...
+import { getLengthErrorMessage } from "../common/utils";
+
+const sizes = {
+  small: "240px",
+  medium: "360px",
+  large: "480px",
+  fillParent: "100%",
+} as const;
+
+const MessageInputContainer = styled.div<{ size: PromptInputPropsType["size"] }>`
```
Jialecl: Since the gap was no needed, the flex is also not needed so this styled component can be swapped for a DxcContainer

Jialecl (changes requested):

packages/lib/src/message-input/utils.ts
PelayoFelgueroso: fixed

packages/lib/src/message-input/types.ts
PelayoFelgueroso: fixed

packages/lib/src/message-input/MessageInput.tsx
PelayoFelgueroso: fixed

apps/website/screens/components/message-input/overview/MessageInputOverviewPage.tsx
```diff
@@ -0,0 +1,319 @@
...
+            </DxcBulletedList>
+          </>
+        ),
+      },
+      {
+        title: "Generating state",
+        content: (
+          <>
+            <DxcParagraph>
+              The generating state is specific to AI-assisted interfaces. It locks the send button to prevent duplicate
```
Jialecl: locks the send button should be bolded.

apps/website/screens/components/message-input/overview/MessageInputOverviewPage.tsx
```diff
@@ -0,0 +1,319 @@
...
+        </Figure>
+      </>
+    ),
+    subSections: [
+      {
+        title: "Top actions zone",
+        content: (
+          <>
+            <DxcParagraph>
+              The top actions zone provides space for contextual chips and a dropdown. When enabled, it appears as a
```
Jialecl: contextual chips should be bolded

Jialecl (approved):

PR #2517, opened 2026-08-05
[minor] Exposing ref in Avatar, badge, button, card and chip to improve popover compatibility

Jialecl:
…er compatibility

**Checklist**

**Purpose**
Popover needs access to the ref to know the exact position it needs to placed itself in reference to the trigger.

### DIFF
--- a/apps/website/screens/components/card/code/CardCodePage.tsx
+++ b/apps/website/screens/components/card/code/CardCodePage.tsx
@@ -268,6 +268,19 @@ const sections = [
             <td>
               <DxcFlex direction="column" gap="var(--spacing-gap-xs)" alignItems="baseline">
                 <StatusBadge status="new" />
+                ref
+              </DxcFlex>
+            </td>
+            <td>
+              <TableCode>{"React.Ref<HTMLDivElement | HTMLAnchorElement>"}</TableCode>
+            </td>
+            <td>Ref to the card element.</td>
+            <td>-</td>
+          </tr>
+          <tr>
+            <td>
+              <DxcFlex direction="column" gap="var(--spacing-gap-xs)" alignItems="baseline">
+                <StatusBadge status="new" />
                 selected
               </DxcFlex>
             </td>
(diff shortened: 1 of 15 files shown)


packages/lib/src/avatar/Avatar.tsx
```diff
@@ -63,107 +63,145 @@ const StatusContainer = styled.div<{
   background-color: ${({ status }) => getModeColor(status!.mode)};
 `;
```
PelayoFelgueroso, line 65: It would be better to keep memo before forwardRef in order to avoid unnecesary re-renders

packages/lib/src/button/Button.tsx
```diff
@@ -66,38 +67,49 @@ const IconContainer = styled.div<{
...
-      size={size}
-      tabIndex={disabled ? -1 : tabIndex}
-      type={type}
-      $mode={mode}
-    >
-      {label && <LabelContainer>{label}</LabelContainer>}
-      {icon && <IconContainer size={size}>{typeof icon === "string" ? <DxcIcon icon={icon} /> : icon}</IconContainer>}
-    </Button>
-  </Tooltip>
+// add ref and useRef to forward the ref to the button element
```
PelayoFelgueroso: do we need this comment?

apps/website/screens/components/card/code/CardCodePage.tsx
```diff
@@ -264,6 +264,19 @@ const sections = [
...
+            <td>
+              <DxcFlex direction="column" gap="var(--spacing-gap-xs)" alignItems="baseline">
+                <StatusBadge status="new" />
+                ref
+              </DxcFlex>
+            </td>
+            <td>
+              <TableCode>{"React.Ref<HTMLDivElement | HTMLAnchorElement>"}</TableCode>
+            </td>
+            <td>Ref to the card element.</td>
```
PelayoFelgueroso: Should we explain when it has HTMLDivElement and when HTMLAnchorElement?

packages/lib/src/button/Button.tsx
Jialecl: removed

apps/website/screens/components/card/code/CardCodePage.tsx
Jialecl: Added an explanation

PelayoFelgueroso (approved):

PelayoFelgueroso (approved):

PR #2539, opened 2026-09-11
[patch] Remove localization errors in v17.1.0

PelayoFelgueroso:
**Checklist**

**Purpose**
Fix errors related lo changes that are not included in the v17


packages/lib/src/message-input/MessageInput.tsx
```diff
@@ -171,8 +170,8 @@ const DxcMessageInput = ({
       value: newValue,
       minLength,
       maxLength,
-      minLengthErrorMessage: translatedLabels.formFields.minLengthErrorMessage,
-      maxLengthErrorMessage: translatedLabels.formFields.maxLengthErrorMessage,
+      minLengthErrorMessage: (min: number) => translatedLabels.formFields.lengthErrorMessage(min, maxLength),
```
Jialecl: This does not make sense at all. Both `minLengthErrorMessage` and `maxLengthErrorMessage` will be the same message.
There is no need to create the `getLengthErrorMessage` function.

PelayoFelgueroso: done

